PR #9, opened 2023-06-13
Feature: Brn Avatar

mihajm:
## Issue
[RFC: Avatar](https://github.com/goetzrobin/spartan/issues/6)

## What was changed/added

Added a brn-avatar component, brnAvatarImage & brnAvatarFallback directives, an initials pipe & some utility methods to help with resolving tailwinds styles for autoColoring the fallback in its hlm variation.

## Expected behavior;

Expected behavior is that the avatar renders the fallback if the image is not provided, doesn't have an src or errors on load.
The avatar component's projected content is restricted to the above directives, no other content should appear.

## Potential for refactor & upcoming

A potential refactor to attempt for the fallback component would be to use cdkObserveContent, I'll attempt that before I submit a complete PR.

Currently the directives (especially the fallback) don't contain much testable behavior, most of these - ex. autoColor will be tested in the hlm directives, once those are done.

### DIFF
--- a/libs/ui/avatar/brain/src/lib/brn-avatar.component.spec.ts
+++ b/libs/ui/avatar/brain/src/lib/brn-avatar.component.spec.ts
@@ -1,5 +1,6 @@
 import { Component } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
 import { BrnAvatarComponent } from './brn-avatar.component';
 import { BrnAvatarFallbackDirective } from './fallback/brn-avatar-fallback.directive';
 import { BrnAvatarImageDirective } from './image/brn-avatar-image.directive';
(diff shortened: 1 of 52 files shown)


mihajm:
Potentially the auto-color logic could be refactored into it's own hlm directive, making it fully re-usable anywhere & cleaning up the fallback component's logic. This could then either be composed into the fallback directive or just used as it's own thing (which is probably best). It would however require the user providing the content for the autoGeneration, since it would be overly-complex to resolve the textContent of any component

ex:
```html
<span *brnAvatarFallback hlmColorsFor="JD">JD</span>
```

goetzrobin:
I like the idea of making this it's own directive, because that way it can probably also be tree shaken for people not using the functionality

libs/ui/avatar/brain/src/lib/brn-avatar.component.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+    expect(empty).toBeFalsy();
+  });
+
+  it('should render the fallback when provided and image with no src', () => {
+    const fallback = fixture.nativeElement.querySelector('#noSrc span');
+    expect(fallback.textContent).toEqual('fallback');
+  });
+
+  it('should not render the fallback, but rather the image when provided with a valid src', () => {
+    const img = fixture.nativeElement.querySelector('#good img');
```
goetzrobin, line 59: This is the test that fails for me :(

PR #56, opened 2023-11-17
fix: abstract injection token creation

nartc:
This PR adds `createInjectionToken` function to abstract the way we create `InjectionToken`

- `InjectFn` is strongly-typed and has overloads to properly support `{optional: true}`
- `ProvideFn` is strongly-typed version of `{provide, useValue}`
- `ProvideFnExisting` is strongly-typed version of `{provide, useExisting}`
- `TOKEN` is exposed for more manual use-cases

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

Currently, the `InjectionToken` inside of `ui-core` is created manually, and the exposed `inject***()` does not support `{ optional: true }` properly, leading to various places using Optional Chaining unnecessarily. 

Closes #

## What is the new behavior?

`InjectionToken` is created with `createInjectionToken`

- [ ] Yes
- [x] No

### DIFF
--- a/libs/nx/src/generators/ui/libs/ui-dialog-helm/files/lib/hlm-dialog-content.directive.ts.template
+++ b/libs/nx/src/generators/ui/libs/ui-dialog-helm/files/lib/hlm-dialog-content.directive.ts.template
@@ -9,7 +9,7 @@ import { ClassValue } from 'clsx';
 export class HlmDialogContentDirective {
   private _inputs: ClassValue = '';
   private _statusProvider = injectExposesStateProvider({ host: true });
-  public state = this._statusProvider?.state ?? signal('closed').asReadonly();
+  public state = this._statusProvider.state ?? signal('closed').asReadonly();
   private _renderer = inject(Renderer2);
   private _element = inject(ElementRef);
 
--- a/libs/nx/src/generators/ui/libs/ui-command-helm/files/lib/hlm-command-dialog.directive.ts.template
+++ b/libs/nx/src/generators/ui/libs/ui-command-helm/files/lib/hlm-command-dialog.directive.ts.template
@@ -10,7 +10,7 @@ import { HlmCommandDirective } from './hlm-command.directive';
 })
 export class HlmCommandDialogDirective {
   private _stateProvider = injectExposesStateProvider({ host: true });
-  public state = this._stateProvider?.state ?? signal('closed').asReadonly();
+  public state = this._stateProvider.state ?? signal('closed').asReadonly();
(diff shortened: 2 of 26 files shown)


libs/nx/src/generators/ui/libs/ui-dialog-helm/files/lib/hlm-dialog-content.directive.ts.template
```diff
@@ -9,7 +9,7 @@ import { ClassValue } from 'clsx';
 export class HlmDialogContentDirective {
   private _inputs: ClassValue = '';
   private _statusProvider = injectExposesStateProvider({ host: true });
-  public state = this._statusProvider?.state ?? signal('closed').asReadonly();
```
goetzrobin, line 12: I remember there being a reason these were optional for some of them... I think it is because technically you can use the hlm components and directives without their counterparts. Let's say you just want to attach the same initial styles to something, but do not care about the status changes. I remember something along those lines

goetzrobin (changes requested):
Let me know what you think about my comment and if we should explicitly require the brn part (e.g. throw an error) or if this is actually already covered by your changes and I am just missing it

libs/nx/src/generators/ui/libs/ui-dialog-helm/files/lib/hlm-dialog-content.directive.ts.template
nartc: The change here is reflected by the inject call right above it. If it is indeed `optional` then it should have `{ optional: true }`

PR #64, opened 2023-11-23
fix: accordion trigger is now a directive

### LINKED ISSUE #62: User Class should not be set to Host too e.g. hlm-accordion-trigger

spartan 0.0.1-alpha.313
"@angular/core": "17.0.2",

accordion
sheet
dialog 
alert-dialog

### Description

Wenn setting a class on a   hlmAccordionTrigger the class is set to the button of the brn-accordion-trigger and to the brn-accordion-trigger.
image

```true
In this case the padding is added twice wich is not expected.
```

### Other information

My idea would be, that the class should not be set to the host (button) in this case. 
The userCls should not be set via setClassToCustomElement to the host. 

@goetzrobin if you also think would be better i would take a look on this. 

- [X] Yes
- [ ] No

elite-benni:
I just edited the trigger for now. 
There are some accesibility problems with the focus index when tabbing through, that i will check now. 
Happy to hear your feedback

Closes #62


goetzrobin:
@elite-benni I hate to ask, but I know my changes broke the preventDefault you had added. I believe I ran unit & e2e tests after and they passed. Is there a test that makes sure this a11y feature won't be broken again? If not, are you able to add a test for it as part of this PR? Also happy, to merge this PR and create an issue for the testing part!

elite-benni:
You are right, test should have been there in the first place. I will add the test in this PR.

elite-benni:
I just added tests for the brn-accordion by now. 
Asserting the preventedDefault was actually harder that i thought. ;)
There are several tests missing since the beginning? :)

PR #98, opened 2024-01-03
feat: brn-label

thatsamsonkid:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] input
- [x] label
- [ ] menubar
- [ ] navigation-menu
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

Label component currently does not have any logic attached currently no auto generated id. But we've found there is some light logic usually required of labels for use in other components in library.

Closes #

## What is the new behavior?

Label component will auto generate a unique id unless an id is provided by user. This is especially helpful in other components where we need some sort of Id for ADA purposes for example.

- [ ] Yes
- [x] No


libs/ui/label/brain/src/lib/brn-label.directive.ts
```diff
@@ -0,0 +1,43 @@
...
+
+@Directive({
+	selector: '[brnLabel]',
+	standalone: true,
+	host: {
+		'[id]': '_id()',
+	},
+})
+export class BrnLabelDirective implements OnInit {
+	protected readonly _id = signal(`brn-label-${nextId++}`);
```
ajitzero, line 14: Question: Is this ID stable?
- On the first page we visit, we get `brn-label-1`, then `brn-label-2`, and so on, but if I visit a new page, I would get `brn-label-3` instead of 1 again.
- If I refresh the second page, we would get `brn-label-1` instead of `brn-label-3`, since I imagine the `id` is not maintained on a per page/component basis.

This may be a non-issue in the SSG case, where pages will be pre-rendered in a particular order, so the second page above will always have `brn-label-3`, not 1. I am thinking about the SSR scenario, which I am not too familiar with.

I'm not sure what the headless-ui team does differently with React or if they have the same problem:

https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/field/field.tsx#L31-L51

https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/hooks/use-id.ts

libs/ui/label/label.stories.ts
```diff
@@ -50,10 +53,8 @@ export const InputRequired: Story = {
...
 		props: { ...args, value: '' },
 		template: `
-    <label hlmLabel>E-Mail *
-        <input [(ngModel)]="value" class='w-80' hlmInput ${argsToTemplate(
-					args,
-				)} type='email' placeholder='Email *' required/>
+    <label hlmLabel ${argsToTemplate(args)}>E-Mail *
+        <input [(ngModel)]="value" class='w-80' hlmInput  type='email' placeholder='Email *' required/>
     </label>
     `,
```
ajitzero, line 59: Minor: I think you may not have the latest Prettier version in your local node modules. Or maybe the Prettier doesn't recognize stories as angular files?

(Arrow for tab, dot for space)
image

libs/ui/label/brain/src/lib/brn-label.directive.ts
thatsamsonkid: Yes in this case the Id will not be stable between page refreshes necessarily. I think the goal with auto generated id is to ensure that the id is absolutely unique on the page. This is good internally for us to use for ada attributes usually. If you need a persisted Id then you would need to supply it and that would be maintained always in that case. Material does id generation in the same way, check out https://github.com/angular/components/blob/main/src/material/input/input.ts but I agree with you that it does seem a little quirky that between refreshes the id continues to increment but I believe it should be a none issue

libs/ui/label/label.stories.ts
thatsamsonkid: I believe it's as you said about Prettier not recognizing angular in stories.ts. But I could be wrong, someone else can correct me and I can check.

libs/ui/label/brain/src/lib/brn-label.directive.ts
ajitzero: That makes sense. Thanks for explaining!

libs/ui/label/label.stories.ts
goetzrobin: @thatsamsonkid seems like it... not sure if its a config issue or limitation...

goetzrobin (approved):
LGTM! @thatsamsonkid feel free to fix the formatting issues in the .stories files if you already figured out how to do it. If not, I'd say we can create a follow up issue to not hold up this PR :)

thatsamsonkid:
Let's merge! I'll check that formatting later. We have bigger fish to fry haha

PR #180, opened 2024-03-01
feat(dialog): open dynamic component in dialog

### LINKED ISSUE #35: Can't open a dialog from a different component.

@angular/core: "16.2.6"
@angular/cdk: "16.2.8"
@spartan-ng/ui-core: "0.0.1-alpha.309"
@spartan-ng/ui-dialog-brain: "0.0.1-alpha.309"

dialog

### Description

Currently the Angular's cdk dialog login is wrapped inside brnDialog (and brnDialogService), so it's impossible to trigger open a dialog from another component.

Say I want to open a dialog from a random button. With Angular's cdk I would do:

```ts
import {Component, inject} from "@angular/core";
import {Dialog} from "@angular/cdk/dialog";

@Component({
    selector: 'app-button',
    standalone: true,
    template: `
    <button (click)="openDialog()">Open</button>`
})
export class ButtonComponent {
    dialog = inject(Dialog);
    
    openDialog() {
        this.dialog.open(MyDialogComponent);
    }
}
```

BrnDialogService is made to be use only inside BrnDialog, so it removes this feature completely.
The only way to show a dialog is to have both dialog and trigger in the same template,

_No response_

### Other information

_No response_

- [ ] Yes
- [ ] No

ty-ler:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [x] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [x] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [x] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [x] sheet
- [ ] skeleton
- [ ] slider
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

Closes #35 

Dialogs can currently only be created and opened from a component's template. Opening a dialog from a service is possible (using a subject in the service), but requires the dialog to exist in the template of an already existing component.

## What is the new behavior?

Dialogs can now be opened from anywhere where the `HlmDialogService` can be injected.

Most dialog-related components relied on the `BrnDialogComponent` existing for the dialog's open/closed state. Since `BrnDialogComponent` did not exist in the context of dynamically created dialogs, those reliant components had to be slightly reworked. These were just internal changes - the public APIs are the same.

- `BrnDialogService` is now provided in root.
- `HlmDialogService` is a new service which has a method for opening a component in a styled dialog, from a component or service.
  - Can pass through a context object and providers to the dialog.

- [x] Yes
- [ ] No

- The `injectDialogCtx` has been renamed to `injectDialogContext`.
- `provideBrnDialog()` is no longer necessary (`BrnDialogService` is provided in root)

## Other information

An example section in the dialog documentation has been added to demonstrate this new feature. A matching story in the storybook has also been added.


libs/ui/dialog/brain/src/lib/brn-dialog-close.directive.ts
```diff
@@ -1,22 +1,23 @@
...
 	private _delay: number | undefined;
+
 	@Input()
 	set delay(value: NumberInput) {
 		this._delay = coerceNumberProperty(value);
 	}
-	close() {
-		this._brnDialog.close(this._delay);
+
+	@HostListener('click')
```
goetzrobin, line 19: Let's stick with the `host` inside of `@Directive`

libs/ui/dialog/brain/src/lib/brn-dialog-content.directive.ts
```diff
@@ -1,27 +1,36 @@
...
-	public state = this._brnDialog.state;
+export class BrnDialogContentDirective<T> {
+	private readonly _brnDialog = inject(BrnDialogComponent, { optional: true });
+	private readonly _brnDialogRef = inject(BrnDialogRef, { optional: true });
+	private readonly _template = inject(TemplateRef);
+	public readonly state = computed(() => this._brnDialog?.state() ?? this._brnDialogRef?.state() ?? 'closed');
 
 	constructor() {
-		this._brnDialog.registerTemplate(this._template);
+		if (this._brnDialog) {
```
goetzrobin, line 18: Can we change this to early returns please!

libs/ui/dialog/brain/src/lib/brn-dialog-trigger.directive.ts
```diff
@@ -30,6 +35,8 @@ export class BrnDialogTriggerDirective {
 	set brnDialogTriggerFor(brnDialog: BrnDialogComponent) {
 		this._brnDialog = brnDialog;
 	}
+
+	@HostListener('click')
```
goetzrobin, line 39: Let's keep this inside `host`

libs/ui/popover/brain/src/lib/brn-popover-trigger.directive.ts
```diff
@@ -7,7 +7,6 @@ import { BrnPopoverComponent } from './brn-popover.component';
 	standalone: true,
 	host: {
 		'[id]': '_id()',
-		'(click)': 'open()',
```
goetzrobin, line 10: Why is this getting removed?

goetzrobin:
I think this is an awesome improvement for the current Dialog! I do want to make sure we are still able to have multiple dialogs open at the same time. Can you add a test for that?

Also, let's continue to support the `injectDialogCtx` method and add the `injectDialogContext`. We can mark `injectDialogCtx` as deprecated and remove before we go stable. I do want to be considerate as I believe that the Dialog Component is one of the most used!

ty-ler:
> I think this is an awesome improvement for the current Dialog! I do want to make sure we are still able to have multiple dialogs open at the same time. Can you add a test for that?
> 
> Also, let's continue to support the `injectDialogCtx` method and add the `injectDialogContext`. We can mark `injectDialogCtx` as deprecated and remove before we go stable. I do want to be considerate as I believe that the Dialog Component is one of the most used!

Thanks for your review! Yes, continuing support for `injectBrnDialogCtx` is a good idea. I will add that function back into the PR.

On a similar note - should the `injectDialogContext` really be named `injectBrnDialogContext`? I realize now that I dropped the `Brn` part without thinking.

libs/ui/popover/brain/src/lib/brn-popover-trigger.directive.ts
ty-ler: I removed this from the host metadata (along with the other directives) because I was seeing an error where open() was being called twice somehow, specifically in the popover trigger. I assumed it was because the inherited directive (`BrnDialogTriggerDirective`) also had the `(click)` event in the host metadata.  

After reverting those changes, it seems like the double call is not happening anymore, so there must have been another issue that I was seeing at the time which has since been cleared up.

ty-ler: Ah actually, I am seeing it again now. I had forgot to add `(click)` back to the popover trigger directive. Now when I click the popover trigger, I can see the `open()` method is being called twice.

I recorded a GIF to demonstrate the issue quickly:

host-click

goetzrobin:
> > I think this is an awesome improvement for the current Dialog! I do want to make sure we are still able to have multiple dialogs open at the same time. Can you add a test for that?
> > Also, let's continue to support the `injectDialogCtx` method and add the `injectDialogContext`. We can mark `injectDialogCtx` as deprecated and remove before we go stable. I do want to be considerate as I believe that the Dialog Component is one of the most used!
> 
> Thanks for your review! Yes, continuing support for `injectBrnDialogCtx` is a good idea. I will add that function back into the PR. Will also add in the test for multiple dialogs.
> 
> On a similar note - should the `injectDialogContext` really be named `injectBrnDialogContext`? I realize now that I dropped the `Brn` part without thinking.

Let's do it! Thanks for working on this!

thatsamsonkid (commented):
This looks really good, I think we just need to update the popover e2e tests along with some of the other feedback and I would say this is probably good.

ty-ler:
> This looks really good, I think we just need to update the popover e2e tests along with some of the other feedback and I would say this is probably good.

Changed the popover component's `ariaDescribedBy` and `ariaLabelledBy` values to be empty strings by default so that the popover tests now work as intended. This way the values are not assigned in the dialog service, and the aria values do not appear in the DOM.

goetzrobin (approved):

Muneersahel:
Does this feature included in the documentation? I can't see it inside the current deployed documentation

elite-benni:
> Does this feature included in the documentation? I can't see it inside the current deployed documentation

The dynamic component example covers this PR.

PR #206, opened 2024-03-14
feat(tabs): paginated tab list

### LINKED ISSUE #203: Paginated tabs

tabs

### Information

Providing a way to scroll through multiple tabs with paginated header. This would be helpful when displaying multiple tabs or for mobile view.

Example from Angular Material Tabs: https://material.angular.io/components/tabs/examples#tab-group-paginated

The paginated header should show before/after buttons when the header needs to scroll, otherwise hides the buttons.

_No response_

- [X] Yes
- [ ] No

marcjulian:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] spinner
- [ ] switch
- [ ] table
- [x] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

Closes #203

## What is the new behavior?

Add `hlm-paginated-tab-list` component which handles pagination built on top of [MatPaginatedTabHeader](https://github.com/angular/components/blob/main/src/material/tabs/paginated-tab-header.ts)

CleanShot 2024-03-14 at 11 23 18

Current issues I am facing:

1. Doesn't scroll on initial load to another tab
Should find the position of the "Tab 20" and scroll it into view.
CleanShot 2024-03-14 at 11 29 54

3. Scroll position is calculated from tabs trigger, but the extra padding (`p-1`) on the list interferes with the correct scroll position
When scrolling left with the keyboard it scrolls to "Tab 29" but does not consider the extra padding. The next paginator button is also still enabled, but should be disabled when it is at the end.
CleanShot 2024-03-14 at 11 31 00
This is when it is scrolled all the way to the end. The next paginator is disabled and the padding around the tab is visible.
CleanShot 2024-03-14 at 11 32 50

- [ ] Yes
- [x] No


libs/ui/tabs/brain/src/lib/brn-tabs-paginated-list.directive.ts
```diff
@@ -0,0 +1,662 @@
...
+		if (currentKey && this._items) {
+			// TODO currentTab is undefined for initial load
+			const currentTab = this._tabs[currentKey];
+			if (currentTab) {
+				activeIndex = this._items.toArray().indexOf(currentTab.trigger);
+			}
+		}
+
+		return activeIndex;
+	});
```
marcjulian, line 170: I am calculating the `selectedIndex` based on the roots tab value. Then I need to get the current tab for the tab value to calculate the position of the tab. However, the `currentTab` is always `undefined` on initial load. Probably the tab value is available before the `tabs` have been registered.

goetzrobin: I think tabs was created pre signals. What if we make `tabs` a signal?

marcjulian: Yes that did the trick! The initial tab is now correctly calculated and scrolled into view.

marcjulian:
@goetzrobin do you have an idea how to handle the extra padding around the tabs? I noticed when I add padding to the material tabs it behaves the same as with the current solution. 

Not sure if we need to specify the padding amount to consider while scrolling.

goetzrobin:
> @goetzrobin do you have an idea how to handle the extra padding around the tabs? I noticed when I add padding to the material tabs it behaves the same as with the current solution.
> 
> Not sure if we need to specify the padding amount to consider while scrolling.

I think this is a tough one to figure out. I'd say as good as Angular Material is good enough for me. What do you think?

marcjulian:
Yes think so too, will be tough to figure out the value. I am fine with the current limitation. I can add a note about it to the docs. I will add paginated tabs as a section to the tabs page.

libs/ui/tabs/helm/src/lib/hlm-tabs-paginated-list.component.ts
```diff
@@ -0,0 +1,92 @@
...
+	template: `
+		<button
+			#previousPaginator
+			type="button"
+			aria-hidden="true"
+			hlmBtn
+			variant="ghost"
+			size="icon"
+			tabindex="-1"
+			[ngClass]="{
```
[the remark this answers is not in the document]
marcjulian: That didn't work correctly. I changed it to 

```html
[class.flex]="_showPaginationControls"
[class.hidden]="!_showPaginationControls"
```

goetzrobin:
Cool that's awesome! Maybe we can make the docs more fun at a later date, but definitely not a blocker! This is great!

marcjulian:
Good idea to improve the docs. Adding all inputs/outputs of a component could also be helpful.

goetzrobin:
@marcjulian is this ready to go?

marcjulian:
@goetzrobin yes is ready to go. I will solve the merge conflicts.

PR #232, opened 2024-03-27
feat(tooltip): support passing `string` to tooltip trigger

### LINKED ISSUE #222: Easier usage of tooltip

tooltip

### Information

Currently to show a tooltip on a button I:

1. need to wrap it with a `hlm-tooltip` component
2. define a `span` element and apply the `brnTooltip` structural directive on it
3. pass a template reference to the `hlmTooltipTrigger` that is applied to the button

Very complex, deeply nested and you need to know exactly which type of directive (hlm vs brn) to apply where.

---

## Proposal

I know I'm gonna sound like a stuck CD on repeat, but Material's API is much more comfortable in this regard.
Simply apply the `[matTooltip]` directive to the button and it's done.

Obviously the HlmTooltip allows for defining any template, also icons, for the tooltip, which is something I like and would definitely want to keep. But for the majority of cases where I just want to apply a quick tooltip to a icon button (hence no icon in the tooltip needed) I think Material's approach is superior.

_No response_

- [X] Yes
- [ ] No

alexciesielski:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [x] tooltip
- [ ] typography

Closes #222

- [ ] Yes
- [x] No


libs/ui/core/src/lib/brain/typeof.pipe.ts
```diff
@@ -0,0 +1,8 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({ name: 'typeofString', standalone: true })
+export class TypeOfStringPipe implements PipeTransform {
+	transform(value: unknown): value is string {
+		return typeof value === 'string';
```
goetzrobin, line 6: We could probably just use a function call in the  template of the component. I am cool with keeping this though if you think we should

libs/ui/tooltip/brain/src/lib/brn-tooltip-content.component.ts
```diff
@@ -60,8 +65,7 @@ export class BrnTooltipContentComponent implements OnDestroy {
 	public readonly _tooltipClasses = signal('');
 	public readonly side = signal('above');
 	/** Message to display in the tooltip */
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any
-	public template: TemplateRef<any> | null = null;
+	public message: string | TemplateRef<unknown> | null = null;
```
goetzrobin, line 68: Maybe we should name this content instead of message?

libs/ui/tooltip/tooltip.stories.ts
```diff
@@ -53,6 +53,44 @@ export const Default: Story = {
...
+		BrnTooltipContentDirective,
+		HlmTooltipTriggerDirective,
+		HlmIconComponent,
+	],
+	providers: [provideIcons({ lucidePlus })],
+	template: `
+		<div class="p-40">
+			<button
+				(click)="disabled.set(!disabled())"
+				[hlmTooltipTrigger]="'Add to library'"
```
goetzrobin, line 71: This does look super clean! I like it a lot!

README.md
```diff
@@ -59,6 +59,7 @@ Our initial 300 contributors and sponsors are featured here and on the front pag
 40. [badsgahhl](https://github.com/badsgahhl)
 41. [monacodelisa](https://github.com/monacodelisa)
 42. [tomdev9](https://github.com/tomdev9)
+43. [alexciesielski](https://github.com/alexciesielski)
```
goetzrobin, line 62: You're already added at 37 :)

goetzrobin (commented):
Some minor comments! Looks great! Thank you!!

libs/ui/core/src/lib/brain/typeof.pipe.ts
alexciesielski: I'm not a huge fan of function calls in templates, but I guess now that signals are coming and are simply functions I guess it's not a big deal to replace the pipe with one.

libs/ui/tooltip/brain/src/lib/brn-tooltip-content.component.ts
alexciesielski: Renamed

README.md
alexciesielski: Whoops

goetzrobin:
Looks great @alexciesielski! One idea. If we pass in a string should we set the aria described by to this string if there's no explicit one set? I'd also like to show a warning in dev mode if there's no aria label set, to let developers know they should set one to help users that use accessibility tech. Thoughts?

alexciesielski:
Hey,

I added a `brnDevMode` flag because I didn't find one yet, it's based on `ngDevMode` which gets set by the Angular compiler to (obviously) false when it is run with production mode enabled. 

This then allows the compiler to trees-hake any code that it encounters, where the flag was used, which now basically looks like:

```ts
if(false) {
 // dev code that should not go into prod
}
```

and I used it to show a warning if the describedby input wasn't set. But I guess it would be more elegant to just set it for the user. How would you suggest I go about this? 

Maybe make an `effect(() => )` in the `constructor` and set it to the `tooltipTrigger` if it is a `string` and the `aria-describedby` isn't set?

elite-benni:
I Really like this addition!
Good Job! ;)

goetzrobin (commented):
One more 🙏

libs/ui/tooltip/brain/src/lib/brn-tooltip-trigger.directive.ts
```diff
@@ -55,6 +56,8 @@ import { BrnTooltipDirective } from './brn-tooltip.directive';
 export type TooltipPosition = 'left' | 'right' | 'above' | 'below' | 'before' | 'after';
 export type TooltipTouchGestures = 'auto' | 'on' | 'off';
 
+declare const ngDevMode: boolean;
```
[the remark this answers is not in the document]
alexciesielski: Forgot to remove, thanks

PR #253, opened 2024-04-06
refactor(select): decouple hlm from brn

### LINKED ISSUE #252: Renaming the generated `hlm-` selector to anything else breaks the `select` component

Any environment with `select`

select

### Description

As already [mentioned in Discord](https://discord.com/channels/1145362148621557921/1145362149133271082/1213189682247106580) I like to make use of the fact that I have full control over the generated components in my repository and like to change the `hlm` prefix to my customer's prefixes.

Unfortunately this breaks the `select` component because the `brn` part specifically selects for both a `label[hlmLabel]` and `hlm-select-trigger` [here](https://github.com/alexciesielski/spartan/blob/b3bbcf8962046d5823b3ffe0803de4496bdfed7a/libs/ui/select/brain/src/lib/brn-select.component.ts#L64).

I argue that this approach is architecturally and conceptually incorrect, because it's not only a (conceptually) circular dependency, but also tightly couples the `hlm` implementation to the `brn` part. 

While, if I understand correctly, `brn` is supposed to be a headless standalone collection of directives and components and should be unaware of where/how it's used.

#### Proposed solution

Again, as already mentioned in Discord, my proposed solution would be to decouple `hlm` from `brn` and instead of using `ng-content` with a `select` I would instead opt for the (while more verbose) more explicit way of passing a template reference either through an `@Input` or (as proposed in Discord) through a `@ContentChild`.

Since we're still in alpha I would even suggest taking the dive and making this a breaking change and completely removing any mention of `hlm` inside `brn`, although I think it should be possible to keep the current solution while adding support for selecting the label and trigger through `ContentChild`.

I would even go as far as suggesting adding a eslint rule for every `brn` library that would consist of a regex looking for `hlm` throughout its code and erroring on that.

_No response_

### Other information

Happy to open a PR with the suggested solution of tagging the (hlm) `label` and `trigger` components with a directive and querying it in the `BrnSelect` component using `@ContentChild`

- [X] Yes
- [ ] No

alexciesielski:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [x] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

Closes #252

## What is the new behavior?

Decouple `hlm` from `brn`. There should be no mention of `hlm` inside `brn`. Otherwise it would be a logical circularity.

I think that `brn` in general should contain no components, otherwise it is taking away control over the final component library from developers.

IMHO, `brn` should be a set of directives and helpers to build up a component library - something along the lines of a cdk. I think that less logic in `brn` and more logic in `hlm` is a good thing as it gives developers the most control. But is also more error-prone in case of future updates. Not sure what the best course of action here is, alhtough I tend to think that more code in developer's codebases is a good thing (because it gives more control).

Anyway back to the PR:

What I'm unhappy about with the current solution is the fact that I had to make `updateArrowDisplay` an async function because it needs to access the `viewport` ElementRef, which is now passed to the directive through a signal. 

Happy to hear suggestions for a cleaner solution for this.

In general I tried to leave as much code as possible inside the brn directives, although again I'm wondering how much UI-related logic should live inside brn (again, imho, none or as little as necessary - although I'm having trouble defining what necessary in this case means).

Also, I didn't touch the tests and generators yet. Once we're happy with the changes I will get to it, too.

- [ ] Yes
- [ ] No
- [x] `hlm` components need to be regenerated, but otherwise the API stays the same (as long as `brn` isn't mentioned in the final code)


thatsamsonkid:
I will be honest this is a lot more changes than I was expecting. So overall this kind of transition is a very slippery slope for us from a library maintenance perspective. To be clear I'm not against your objective here which is, from the helm perspective be able to make all of the customizations you could ever want to it and it still continue to work properly, especially in regards to selector renaming for example. However I don't believe this is the way for the reasons below. (Apologize for the long read and also this is only my view, I don't wish to speak for @goetzrobin  and @elite-benni )

### The UI Library Paradigm (as I understand it at least)

**Brn**: is meant to still be out of the box a functioning component, unstyled but functioning and not just a set of pieces left to the developer to figure out how to use to make a working component, even if documented. 

**Hlm**: is meant to be low logic and primarily focused on applying the shadcn skin and not much else.

### Why Hlm has low logic and should continue this way:

Hlm is in the full control of the end user, moving all the logic there while sure gives great customizability to the end user also has a lot of downsides in terms of maintainability and upgradability of this library. 

- People will feel empowered to make customizations and if something breaks we will see them in our issues log asking us to tell them why something is not working which will make it very difficult to maintaining this library. Blackboxing the majority of the logic in Brn helps with maintenance and longevity of this project

- Getting users to upgrade versions will be very difficult if we were to do this and in some extremes they simply will never upgrade versions for a component and maybe just continuously apply upgrades manually. It will also inhibit us from ever making automatic upgrades/migrations most likely an impossibility at that point, if lets say anything related to the shadcn style or behavior of a component were to change.

- We should be consistent in terms of where the divide is in terms of where the logic lives. This is hard because its not clear cut in some cases but I think we are striving to limit the amount of logic we do keep in hlm with the goal of having none if possible.

### Why helm selectors were placed in select brn:

The decision for having the helm selectors in the brn component was not a design decision that was taken lightly. I ran into very much the same issues you are running into yourself. Most alternative solutions ultimately resulted in a much more verbose API. Ultimately we are limited in what Angular allows us to do but our ultimate goal was to provide the best API possible and having the helm selectors in brn, **while not ideal**,  is ultimately not a hard dependency and seemed reasonable compared to the other options. There is currently no hard requirement for hlm to be present at all for brn component to function.

### Closing words 
Now I'm not saying there is no solution for customizing the helm selector, I feel like there should be a way for us to provide a slightly different alternative if there really is a desire to customize the helm selector, but ultimately I believe we should support both because that alternative will definitely be a much more verbose API and IMHO I feel greater amount of people would prefer a simpler API over the ability to customize the selector, which is the only downside with current implementation I see. 

Again I really want to support the helm selector customization, I really hate that the select component right now is the only component in the library that seems to have this shortcoming at the moment.

PR #270, opened 2024-04-24
feat(repo): oxlint lightning-fast linter with more code quality, correctness and performance checks

benpsnyder:
See https://oxc-project.github.io/docs/guide/usage/linter.html

This PR adds an additional layer of awesome quality control at near zero cost to DX. We can use it as our baby steppingstone to lint on CI and incrementally adopt biomejs and its full set of rules.

Embrace the latest and greatest ;)


benpsnyder:
Failures look like this.
image

PR #278, opened 2024-04-25
fix: select form control status dirtied on initial value and patchValue

### LINKED ISSUE #224: The select control become dirty on multiple mode with initlal value or patch value
If the select control is in multiple mode and I have an initial value or patch value (without interacting with the control), it incorrectly gets marked as dirty (`true`). This is incorrect because only when I interact with it should it become dirty, similar to `input` and other controls.

select

- [X] Yes
- [ ] No

thatsamsonkid:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe:

Currently just adds more test related to select's form control

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [x] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

Closes #224 

- [ ] Yes
- [ ] No

### DIFF
--- a/libs/ui/select/brain/src/lib/brn-select.component.ts
+++ b/libs/ui/select/brain/src/lib/brn-select.component.ts
@@ -5,7 +5,6 @@ import {
 	type ConnectedPosition,
 	OverlayModule,
 } from '@angular/cdk/overlay';
-import { JsonPipe } from '@angular/common';
 import {
 	type AfterContentInit,
 	ChangeDetectionStrategy,
@@ -32,7 +31,7 @@ import {
 	provideExposesStateProviderExisting,
 } from '@spartan-ng/ui-core';
 import { BrnLabelDirective } from '@spartan-ng/ui-label-brain';
-import { Subject, delay, map, of, skip, switchMap } from 'rxjs';
+import { Subject, delay, filter, map, of, switchMap } from 'rxjs';
 import { BrnSelectContentComponent } from './brn-select-content.component';
 import { BrnSelectOptionDirective } from './brn-select-option.directive';
 import { BrnSelectTriggerDirective } from './brn-select-trigger.directive';
@@ -45,7 +44,7 @@ let nextId = 0;
 @Component({
 	selector: 'brn-select, hlm-select',
 	standalone: true,
-	imports: [OverlayModule, BrnSelectTriggerDirective, CdkListboxModule, JsonPipe],
+	imports: [OverlayModule, BrnSelectTriggerDirective, CdkListboxModule],
 	changeDetection: ChangeDetectionStrategy.OnPush,
 	providers: [
 		BrnSelectService,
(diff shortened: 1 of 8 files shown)


thatsamsonkid:
Apart from the fix, keep me honest here i'm pretty sure this make sense. I've been spending some time in the material repo and playing around with mat-select component. I have below stackblitz for the material one if in case you guys find it helpful

https://stackblitz.com/edit/iwefen?file=src%2Fexample%2Fselect-error-state-matcher-example.ts

Apart from the fix also open to suggestions for how we can better organize the tests for select if anyone has any ideas especially since I added a bunch here for each scenario (single mode, multi-mode and required and initial value variants). But I also question if we need them all because we apply the statuses directly from ng control so its not really our logic per say. The root issue here was related more to when we triggered the form change. But yea let me know what you guys think 

@goetzrobin @elite-benni

thatsamsonkid:
Need to rebase once #293 goes in

libs/ui/select/brain/src/lib/brn-select.component.ts
```diff
@@ -155,6 +154,8 @@ export class BrnSelectComponent implements ControlValueAccessor, AfterContentIni
 	// eslint-disable-next-line @typescript-eslint/no-empty-function
 	private _onTouched = () => {};
 
+	private _shouldEmitValueChange = signal(false);
```
elite-benni, line 157: Is a signal really needed here?

elite-benni (approved):
looks good and well organized. ;) GJ

libs/ui/select/brain/src/lib/brn-select.component.ts
thatsamsonkid: I guess technically it does not need to be a signal since this value only ever changes if writeValue is triggered or we change value internally. Does it matter you think or would you prefer to just have it has a regular variable?

elite-benni: I just stopped using signals everywhere. 
I think it is unnecessary overhead, even if the overhead of a signal is extremely small. 
If i is a signal, I will assume that there is something reacting on it.
These are just very small arguments and I am fine with keeping it a signal.

PR #283, opened 2024-05-03
fix(select): initialvalue is recognized when provided with formcontrol

### LINKED ISSUE #282: Select: Initial Value is not set when options are provided vie @for

0.0.1-alpha.347

select

### Description

When options are provided via an array in a @for loop. The initial value is not set when provided with an formcontrol.

_No response_

### Other information

_No response_

- [X] Yes
- [ ] No

elite-benni:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [x] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

When options are provided via an array in a @for loop. The initial value is not set when provided with an formcontrol.

Closes #282 

## What is the new behavior?

The initial value is displayed correctly.

- [ ] Yes
- [x] No

## Other information

The value was set via writevalue before the options were registered, therefore the value was ignored.

### DIFF
--- a/libs/ui/select/brain/src/lib/brn-select.service.ts
+++ b/libs/ui/select/brain/src/lib/brn-select.service.ts
@@ -119,8 +119,13 @@ export class BrnSelectService {
 		this.selectOptionByValue(value);
 		this.state.update((state) => ({
 			...state,
+			value: value as string | string[],
 			initialSelectedOptions: this.selectedOptions(),
 		}));
+	}
+
+	public possibleOptionsValueChanged() {
+		this.selectOptionByValue(this.value());
 	}
 
 	private selectOptionByValue(value: unknown) {
--- a/libs/ui/select/brain/src/lib/brn-select.service.ts
+++ b/libs/ui/select/brain/src/lib/brn-select.service.ts
@@ -124,7 +124,15 @@ export class BrnSelectService {
 		}));
 	}
 
-	public possibleOptionsValueChanged() {
+	/*
+	 * We cannot react directly when the option is added, because at this time the value is not always set.
+	 * The option is registered on constructor of brn-select-option.directive.ts, at this time the value is not set, because it is an @Input.
+	 * When the value is set on Input we trigger this function to tell the service, that the values of the possibleOptions changed,
+	 * which causes the service to check if the new value of the possibleOptions matches the value of the select.
+	 * It is a tricky timing problem, which happens because setting the value with writevalue occurs before the options are registered,
+	 * and then it does not select the possibleOption because it cannot find it.
+	 */
+	public possibleOptionsChanged() {
 		this.selectOptionByValue(this.value());
 	}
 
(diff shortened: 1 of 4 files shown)


libs/ui/select/brain/src/lib/brn-select.service.ts
```diff
@@ -119,10 +119,15 @@ export class BrnSelectService {
 		this.selectOptionByValue(value);
 		this.state.update((state) => ({
 			...state,
+			value: value as string | string[],
 			initialSelectedOptions: this.selectedOptions(),
 		}));
 	}
 
+	public possibleOptionsValueChanged() {
+		this.selectOptionByValue(this.value());
```
goetzrobin, line 128: Do we need an extra function for this? I'm also not sure why it's called possible, but there's no if statement. Maybe a comment explaining why is better ?

elite-benni: ok I can add a comment but first i want to explain here. ;) 

there are _possibleOptions in this service. 
```
private readonly _possibleOptions = signal<Array<CdkOption | null>>([]);
```
	
The problem is, that i cannot react directly when the option is added, because at this time the value is not set. 
The option is registered on  constructor of brn-select-option.directive.ts, at this time the value is not set, because it is an @Input. 

When the value is set on input i call the function to tell the service, that the possibleOptionsValue changed, which causes the service to check if the new value of the possibleOptions matches the value of the select.


This whole issue is a tricky timing problem, which is caused, that setting the value with writevalue occurs before the options are registered, and then it does not select the possibleOption because it cannot find it.

goetzrobin: Perfect! I'd leave exactly this as a comment 💪💪

goetzrobin (approved):

goetzrobin:
@thatsamsonkid do you think this might resolve the issue you wrote tests for, too?

thatsamsonkid:
> @thatsamsonkid do you think this might resolve the issue you wrote tests for, too?

Maybe I'll check it out

PR #320, opened 2024-06-29
feature(form-field) create new component to handle errors and hints

### LINKED ISSUE #294: Brn and Hlm Error Directive for Form Fields

Don't know / other

### Information

Error messages for form controls currently need to be handled by user instead of having a directive similar to mat-error to handle showing and hiding the error message to the user and maintaining the style consistent with Shadcn

_No response_

- [X] Yes
- [ ] No

cjosue15:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [x] form-field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

None

Closes #294

## What is the new behavior?

Create a new component to handle errors and hints like the material functionality but with the shadcn styles.

- [ ] Yes
- [x] No


.vscode/settings.json
```diff
@@ -3,7 +3,7 @@
 	"eslint.experimental.useFlatConfig": true,
 	"eslint.format.enable": true,
 	"[typescript]": {
-		"editor.defaultFormatter": "biomejs.biome"
+		"editor.defaultFormatter": "esbenp.prettier-vscode"
```
goetzrobin, line 6: what's the reason for this? We are switching to biome for formatting. I might have to update the docs on this

cjosue15: Sorry, I forgot to delete it

libs/ui/core/src/lib/brain/error-options.ts
goetzrobin, line 1: Instead of core, I feel like this should be in a separate brain lib for forms. Thoughts?

libs/ui/core/src/lib/brain/error-state-tracker.ts
```diff
@@ -0,0 +1,31 @@
...
+import type { ErrorStateMatcher } from "./error-options";
+
+export class ErrorStateTracker {
+	/** Whether the tracker is currently in an error state. */
+	errorState = signal(false);
+
+	/** User-defined matcher for the error state. */
+	matcher: ErrorStateMatcher | null = null;
+
+	constructor(
```
goetzrobin, line 12: Let's use the inject function here

libs/ui/core/src/lib/brain/error-state-tracker.ts
```diff
@@ -0,0 +1,31 @@
+import { signal } from "@angular/core";
+import type { AbstractControl, FormGroupDirective, NgControl, NgForm } from "@angular/forms";
+import type { ErrorStateMatcher } from "./error-options";
+
+export class ErrorStateTracker {
+	/** Whether the tracker is currently in an error state. */
+	errorState = signal(false);
```
goetzrobin, line 7: I don't love errorState as a name. I feel like I'd expect an object with more information instead of a boolean, maybe something like hasError?

libs/ui/core/src/lib/brain/error-options.ts
```diff
@@ -0,0 +1,18 @@
+import { Injectable } from '@angular/core';
+import type { AbstractControl, FormGroupDirective, NgForm } from '@angular/forms';
+
+/** Error state matcher that matches when a control is invalid and dirty. */
+@Injectable()
+export class ShowOnDirtyErrorStateMatcher implements ErrorStateMatcher {
+	isErrorState(control: AbstractControl | null, form: FormGroupDirective | NgForm | null): boolean {
```
goetzrobin, line 7: These names should probably be adjusted to fit the variable name

libs/ui/form-field/helm/src/lib/tests/form-field-example.ts
goetzrobin, line 1: I think we can remove the tests and directives folder and just have a flat folder structure. Also, let's rename this file to form-field-example.spec.ts, I understand there are not actual tests, but naming it that way definitely ensures it's not bundled for prod builds

libs/ui/input/helm/src/lib/hlm-input.directive.ts
```diff
@@ -49,4 +58,36 @@ export class HlmInputDirective {
...
+
+	errorStateTracker: ErrorStateTracker;
+
+	private defaultErrorStateMatcher = inject(ErrorStateMatcher);
+	private parentForm = inject(NgForm, { optional: true });
+	private parentFormGroup = inject(FormGroupDirective, { optional: true });
+
+	errorState = computed(() => this.errorStateTracker.errorState());
+
+	constructor(@Inject(Injector) private injector: Injector) {
```
goetzrobin, line 72: Let's not mix the inject function and constructor injector

libs/ui/input/helm/src/lib/hlm-input.directive.ts
```diff
@@ -49,4 +58,36 @@ export class HlmInputDirective {
...
+			() => {
+				if (this.ngControl) {
+					this.error = this.errorStateTracker.errorState();
+				}
+			},
+			{ allowSignalWrites: true },
+		);
+	}
+
+	ngDoCheck() {
```
goetzrobin, line 90: Why is this necessary? It feels like an ant-ipattern to have to do this in every supported component.

libs/ui/select/brain/src/lib/brn-select.component.ts
```diff
@@ -245,6 +267,10 @@ export class BrnSelectComponent implements ControlValueAccessor, AfterContentIni
 		}
 	}
 
+	ngDoCheck() {
+		this.errorStateTracker.updateErrorState();
```
goetzrobin, line 271: Same here

goetzrobin (changes requested):
I added some comments! I really like the idea of this and the overall approach seems fine to me, too. One big question I have: Is there a way to support this with uncontrolled form controls, too? e.g. just add a (host) directive to any form control that's Angular form compatible and automatically get the functionality, instead of having to manually update the form state on every doCheck

cjosue15:
> I added some comments! I really like the idea of this and the overall approach seems fine to me, too. One big question I have: Is there a way to support this with uncontrolled form controls, too? e.g. just add a (host) directive to any form control that's Angular form compatible and automatically get the functionality, instead of having to manually update the form state on every doCheck

When you say uncontrolled form controls, you mean inputs or selects without any of these (FormControl or formcontrolName).
Also I think it is not possible to use this validation as HostDirective because only some components must have the errorState, another point is each component that will be used inside the FormFIeld component should be use the BrnFormFieldControl in the providers, to be identified by the parent(form-field), additionally using the update of the error in the ngDoCheck is the only way, at least that I know of, where I can reevaluate on every change detection when the error is triggered.

libs/ui/core/src/lib/brain/error-state-tracker.ts
cjosue15: I get your point, but this is not a service is just a class.

libs/ui/core/src/lib/brain/error-state-tracker.ts
cjosue15: I can change it, any idea about the content of the object?
You mean something like

```
state = signal<{ hasError: boolean, ... }>({ hasError: false, ... })
```

libs/ui/input/helm/src/lib/hlm-input.directive.ts
cjosue15: Done!

libs/ui/form-field/helm/src/lib/tests/form-field-example.ts
cjosue15: I removed the folder directives and put the files in the `/lib` folder, but regarding the `test` folder, there are 2 files, one the `form-field.spec.ts` and the `form-field-eample.ts`, the first one is the test itself and the second one is the file containing some mock components for the test file.

I saw the same pattern in other packages as select brain

image

libs/ui/core/src/lib/brain/error-options.ts
cjosue15: something like `ui/form/brain/src/lib/error-options.ts` ?

libs/ui/core/src/lib/brain/error-options.ts
cjosue15: I think i can change it to `isInValid`? any suggestion?

libs/ui/core/src/lib/brain/error-options.ts
cjosue15: any thoughts?

libs/ui/input/helm/src/lib/hlm-input.directive.ts
thatsamsonkid: I was seeing that material actually does it this way as well. They have an ngDoCheck in their form components and it's usually to call some sort of function to update the error state. One example from their repo https://github.com/angular/components/blob/ca7c090fb948c2f27f13d76e7cb567d0ba870ab9/src/material/select/select.ts#L678 . Otherwise not sure how else we could it.

thatsamsonkid (commented):
Just added a few comments for now but overall looks great. Thanks for all the hard work so far, this will be a great addition!

cjosue15:
Hi @thatsamsonkid and @goetzrobin thanks for your time, i think i resolve each doubt, please let me know if you have any question.
One point I would like to understand, is how can I test the generator for this new component, I have created it but I am not sure how to test it.

elite-benni (approved):

thatsamsonkid (approved):
LGTM! As far as the generators, I'm not too sure myself. @goetzrobin is there a way to test them before publishing this?

cjosue15:
any thoughts here?, can we merge it? @goetzrobin

PR #354, opened 2024-08-11
fix(select): update initial value properly when using multi-mode and …

guillermoecharri:
…options are added with for-loop

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [x] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

Currently, in the select component, if you use multi-mode, provide an initial value, and dynamically add options in the template (e.g., using a for-loop), the initial value is not applied correctly. Although it initially appears to be set correctly, selecting an option reveals that no initial value was actually applied. Additionally, the component displays as if both the initial value and the selected option are active, even showing duplicates.

Closes #

- [ ] Yes
- [x] No

### DIFF
--- a/libs/ui/select/brain/src/lib/brn-select-value.component.ts
+++ b/libs/ui/select/brain/src/lib/brn-select-value.component.ts
@@ -1,10 +1,13 @@
-import { ChangeDetectionStrategy, Component, Input, computed, inject } from '@angular/core';
+import { AsyncPipe } from '@angular/common';
+import { ChangeDetectionStrategy, ChangeDetectorRef, Component, Input, computed, inject } from '@angular/core';
+import { toObservable } from '@angular/core/rxjs-interop';
+import { map } from 'rxjs';
 import { BrnSelectService } from './brn-select.service';
 
 @Component({
 	selector: 'brn-select-value, hlm-select-value',
 	template: `
-		{{ value() || placeholder() }}
+		{{ value || placeholder() }}
 	`,
 	host: {
 		'[id]': 'id()',
@@ -22,25 +25,38 @@ import { BrnSelectService } from './brn-select.service';
 	],
 	standalone: true,
 	changeDetection: ChangeDetectionStrategy.OnPush,
+	imports: [AsyncPipe],
 })
 export class BrnSelectValueComponent {
 	private readonly _selectService = inject(BrnSelectService);
 
 	public readonly id = computed(() => `${this._selectService.id()}--value`);
 	public readonly placeholder = computed(() => this._selectService.placeholder());
-	public readonly value = computed(() => {
-		const value = this._selectService.selectedOptions();
-		if (value.length === 0) {
-			return null;
-		}
-		const selectedLabels = value.map((selectedOption) => selectedOption?.getLabel());
+	cdr = inject(ChangeDetectorRef);
 
-		if (this._selectService.dir() === 'rtl') {
-			selectedLabels.reverse();
-		}
+	// it's a little bit ugly, but we need to handle selectedOptions as an Observable here
+	// the reason being that computed signals are only evaluated once per event loop cycle
+	// and sometimes (handling initial values) we need to handle two updates per cycle
+	value$ = toObservable(this._selectService.selectedOptions)
+		.pipe(
+			map((value) => {
+				if (value.length === 0) {
+					return null;
+				}
+				const selectedLabels = value.map((selectedOption) => selectedOption?.getLabel());
 
-		return this.transformFn(selectedLabels);
-	});
+				if (this._selectService.dir() === 'rtl') {
+					selectedLabels.reverse();
+				}
+				const result = this.transformFn(selectedLabels);
+				console.log('DINGDINGDING select-value re-computed', result);
+				this.value = result;
+				this.cdr.detectChanges();
+				return result;
+			}),
+		)
+		.subscribe();
+	value?: string = undefined;
 
 	@Input()
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/libs/ui/select/brain/src/lib/brn-select-value.component.ts
+++ b/libs/ui/select/brain/src/lib/brn-select-value.component.ts
@@ -1,7 +1,5 @@
-import { AsyncPipe } from '@angular/common';
 import { ChangeDetectionStrategy, ChangeDetectorRef, Component, Input, computed, inject } from '@angular/core';
-import { toObservable } from '@angular/core/rxjs-interop';
-import { map } from 'rxjs';
+import { takeUntilDestroyed, toObservable } from '@angular/core/rxjs-interop';
 import { BrnSelectService } from './brn-select.service';
 
 @Component({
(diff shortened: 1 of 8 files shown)


guillermoecharri:
Hi there! First time contributor here so sorry if I messed anything up. 
I see that I have some issues with "ci / prettier" and "UI Tests". Those seem to be failing on main as well so I'm not sure if I need to do something to fix those on my end.
Also not sure if I need to do anything to pass the Vercel check.

thatsamsonkid:
> Hi there! First time contributor here so sorry if I messed anything up. I see that I have some issues with "ci / prettier" and "UI Tests". Those seem to be failing on main as well so I'm not sure if I need to do something to fix those on my end. Also not sure if I need to do anything to pass the Vercel check.

Hi @guillermoecharri like you mentioned seems these are some existing issues in main and not related to the select component so I wouldn't worry about it. At first glance I think this LGTM to me and seems all select tests are still passing. Really liking the change actually but want to give a chance for maybe @elite-benni to take a quick look in case I'm not considering something here

Thanks for taking the time to contribute!

thatsamsonkid (approved):
LGTM!

elite-benni (commented):
First of all, thanks for controbuting.

To me this looks like another tricky scenario that initially the _possibleOptionsChanged_ function in the brain service tried to solve.

I am not sure why this does not work, because this was also introduced to solve an issue with options added in a for loop. 

I would like to check why the old solution does not work in that case, but unfortunately i only have access to my phone right now.

The change looks good to me tough.

guillermoecharri:
> First of all, thanks for controbuting.
> 
> To me this looks like another tricky scenario that initially the _possibleOptionsChanged_ function in the brain service tried to solve.
> 
> I am not sure why this does not work, because this was also introduced to solve an issue with options added in a for loop.
> 
> I would like to check why the old solution does not work in that case, but unfortunately i only have access to my phone right now.
> 
> The change looks good to me tough.

Yeah it's a very tricky timing issue. If you want to debug the specific scenario when you have time you can comment-out lines 322-327 and check out the new story that I added. When setInitialSelectedOptions gets called inside writeValue, it won't be able to access the options yet since they are registered just a bit after.

elite-benni:
> Yeah it's a very tricky timing issue. If you want to debug the specific scenario when you have time you can comment-out lines 322-327 and check out the new story that I added. When setInitialSelectedOptions gets called inside writeValue, it won't be able to access the options yet since they are registered just a bit after.

Ok i see. So the problem is, that the initialvalue is not set correctly. 
PossibleOptionsChanged is called, when the Value is really available. In the past there has been problems with setting the possibleOptions when it was registered, because the input value was not set at this time, thats when possibleOptionsChanged was introduced. 
Your solution also seems to work without this problem, so the content children don't have this timing issue.

We now have to options: 

1. use your solution with the options observable and remove the possibleOptionsChanged, because it is not needed anymore. 
2. also save the initial value (string) in the brain service and also set the intialOptions on possibleOptionsChanged and drop the options$ in the component.
    ```
    public setInitialSelectedOptions(value: unknown) {
		this.selectOptionByValue(value);
			...state,
			value: value as string | string[],
			initialValue: value as string | string[],
			initialSelectedOptions: this.selectedOptions(),
		}));
	}
    ....
	public possibleOptionsChanged() {
		this.selectOptionByValue(this.value());
		this.setInitialSelectedOptions(this.initialValue());
	}

```
What I can think of a problem with your solution is, that when the value is written and the options are added afterwards without writing the value again, the select box would again react weirdly.

I hope I was able to explain it good enough. ;)

**Comment by guillermoecharri** [comment]
> ...
> We now have to options:
> 
> 1. use your solution with the options observable and remove the possibleOptionsChanged, because it is not needed anymore.
> 2. also save the initial value (string) in the brain service and also set the intialOptions on possibleOptionsChanged and drop the options$ in the component.
> ...
> What I can think of a problem with your solution is, that when the value is written and the options are added afterwards without writing the value again, the select box would again react weirdly.
> 
> I hope I was able to explain it good enough. ;)

I updated the "ReactiveFormControlWithForAndInitialValueAndMultiple" story to play around with dynamically changing the available options and it seems to react appropriately for the most part. 

The one case that is a bit strange is if an option you've selected gets removed. Maybe this is what you meant by "without writing the value again" since selecting another value would cause that to get cleared out.

It's a strange case but if the expected behavior is that the selected option gets de-selected then maybe this could be handled in the brain service when deregisterOption is called.
Maybe something like:
```typescript
public deregisterOption(option: CdkOption | null) {
	this._possibleOptions.update((options) => options.filter((o) => o !== option));
	// update value if it was just removed
	if (this.multiple()) {
		const values = this.value() as string[];
		const filteredValues = values.filter((v) => v !== option?.value);
		// only update if there was an actual change
		if (values.length !== filteredValues.length) {
			console.log('old values', values, 'new values', filteredValues);
				...state,
				value: filteredValues,
			}));
		}
	} else {
		const value = this.value() as string;
		if (value === option?.value) {
				...state,
				value: '',
			}));
		}
	}
}
```
From some testing it looks like this still isn't quite fixing the issue. Any thoughts on this approach?

elite-benni:
Yeah it seema like a little mess with value selected options and initialvalue. 
I am not quite sure why this was needed. 
Maybe @thatsamsonkid has more insight to this.

thatsamsonkid (changes requested):
Will try to take a closer look at this either today or tomorrow hopefully. Maybe there's an opportunity to clean this up bit as well solve this issue.

guillermoecharri:
I update the PR. The main changes are:
- Now possibleOptionsChanged is no longer needed. I created an effect in the brn-select component that syncs the possibleOptions in the service when the options contentChildren query updates.
- I was also able to get rid of initalSelectedOptions in favor of just using selectedOptions throughout.
- I implemented the things that @elite-benni  brought up about updating the selectedOptions if the possibleOptions are changed

I ran into the headache that @thatsamsonkid mentioned with the pristine/dirty state, but I was able to account for it by doing a 1-time markAsPristine when setting the initial value.

The only other thing of note was that I ran into some trouble getting brn-select-value to render the initial value properly. It seemed like using the signal it was taking an extra frame to render. The only work around I found for that was manually running change detection :( which is always a bummer.

thatsamsonkid:
@guillermoecharri - nice work!, sorry I feel like I haven't been able to give this the attention it deserves. Been busy with personal matters. I will make sure to carve out some time to look at the latest changes later today. Honestly i'd be ok with the manual change detection check. I had run into a similar issue earlier in the development of select so honestly not the worst thing ever

libs/ui/select/brain/src/lib/brn-select-value.component.ts
```diff
@@ -28,21 +29,31 @@ export class BrnSelectValueComponent {
...
+				}
+				const selectedLabels = value.map((selectedOption) => selectedOption?.getLabel());
+
+				if (this._selectService.dir() === 'rtl') {
+					selectedLabels.reverse();
+				}
+				const result = this.transformFn(selectedLabels);
+				this.value = result;
+				cdr.detectChanges();
+				return result;
```
goetzrobin, line 56: I don't think we need the return here anymore? then we can assign result right to `this.value`. Unless I am missing something here

libs/ui/select/brain/src/lib/brn-select.component.ts
```diff
@@ -121,8 +117,24 @@ export class BrnSelectComponent
...
-	@ContentChildren(BrnSelectOptionDirective, { descendants: true })
-	protected options!: QueryList<BrnSelectOptionDirective>;
+
+	protected options = contentChildren(CdkOption, { descendants: true });
+	protected options$ = toObservable(this.options);
+	protected optionsChanged$ = this.options$.pipe(skip(1));
+	/**
+	 * Whenever the CdkOption query is updated we should update the possibleOptions in the select service
+	 */
+	private syncOptions$ = effect(
```
goetzrobin, line 127: can we move this into the constructor just for the sake of consistency?

libs/ui/select/brain/src/lib/brn-select-value.component.ts
guillermoecharri: Oops yeah that is a hold-over. I'll remove it.

libs/ui/select/brain/src/lib/brn-select.component.ts
guillermoecharri: good ideas, yeah I'll do this

guillermoecharri:
Cleaned up according to the feedback :)

libs/ui/select/brain/src/lib/brn-select-value.component.ts
```diff
@@ -28,21 +29,32 @@ export class BrnSelectValueComponent {
...
+
+	constructor(cdr: ChangeDetectorRef) {
+		// In certain cases (when using a computed signal for value) where the value of the select and the options are
+		// changed dynamically, the template does not update until the next frame. To work around this we can use a simple
+		// string variable in the template and manually trigger change detection when we update it.
+		toObservable(this._selectService.selectedOptions)
+			.pipe(takeUntilDestroyed())
+			.subscribe((value) => {
+				if (value.length === 0) {
+					this.value = '';
```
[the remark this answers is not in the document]
guillermoecharri: Good catch. It should be null.

guillermoecharri:
Updated to fix the minor comments

thatsamsonkid (approved):
LGTM🚀🚀

PR #381, opened 2024-09-21
feat: improved tsconfig paths and esbuild config for nitro

benpsnyder:
Visible when running `nx run app:build --verbose`


benpsnyder:
@goetzrobin this is ready to go

PR #382, opened 2024-09-21
chore: formatting and lint improvements

benpsnyder:


benpsnyder:
Does what we want it to do
image

PR #384, opened 2024-09-22
chore: add oxlintrc config and fix circular import lint errors

benpsnyder:


goetzrobin:
I think this might be a solution to avoid the circular dependencies: https://www.bennadel.com/blog/3836-using-abstract-classes-as-dependency-injection-tokens-with-providedin-semantics-in-angular-9-1-9.htm

Create a separate file with an abstract class of e.g. BrnRadioComponent and then use that in the BrnRadioGroupComponent

benpsnyder:
> I think this might be a solution to avoid the circular dependencies: https://www.bennadel.com/blog/3836-using-abstract-classes-as-dependency-injection-tokens-with-providedin-semantics-in-angular-9-1-9.htm
> 
> Create a separate file with an abstract class of e.g. BrnRadioComponent and then use that in the BrnRadioGroupComponent

All for it. I think we should close this PR as-is where I had to consolidate some files. Then we can refactor and break things back out ... easier to do it piece by piece this way.

PR #389, opened 2024-09-27
chore: increase type-safety with stricter tsconfig settings

benpsnyder:


libs/ui/toggle/brain/src/lib/brn-toggle.directive.ts
```diff
@@ -39,15 +38,13 @@ export class BrnToggleDirective {
 	public id = `brn-toggle-${BrnToggleDirective.uniqueId++}`;
 
 	@Input()
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any
-	public value: any;
+	public value: string | number | null = null;
```
alexciesielski, line 41: Should probably also allow undefined, as Signals default to undefined (as opposed to the async pipe which defaults to null) - there will definitely come a follow up ticket about this.

libs/ui/core/src/lib/brain/rx-host-integration.ts
```diff
@@ -27,6 +27,9 @@ function process<T>(element: HTMLElement, prop: string): (value: T) => void {
 	const [key, unit = ''] = prop.replace('attr.', '').replace('style.', '').replace('class.', '').split('.');
 
 	return (value) => {
+		// Ensure that the key is a valid string
+		if (!key) return;
```
alexciesielski, line 31: nitpick: this is confusing. you're saying to ensure that key is a string, but you're not really checking if it is? is the comment necessary even?

libs/ui/toggle/brain/src/lib/brn-toggle.directive.ts
goetzrobin: I think this should remain any. It gives developers the most flexibility and we shouldn't dictate what they use as their value. It's also how the Angular CDK works and I wanna keep that

apps/app/src/app/pages/(components)/components/(dialog)/dialog-dynamic-component.preview.ts
```diff
@@ -107,7 +107,7 @@ export class DialogDynamicComponentPreviewComponent {
 	`,
 })
 class SelectUserComponent {
-	@HostBinding('class') private readonly _class: string = 'flex flex-col gap-4';
+	// @HostBinding('class') private readonly _class: string = 'flex flex-col gap-4';
```
goetzrobin, line 110: we can move this to the host property

libs/cli/src/generators/theme/libs/add-theme-to-application-styles.ts
```diff
@@ -74,7 +74,7 @@ function findStylesEntryPoint(tree: Tree, project: ProjectConfiguration): string
 	}
 
 	// then check for the specified styles in the build configuration if it exists
-	const styles: Array<string | { input: string; inject: boolean }> = project.targets?.build.options?.styles;
+	const styles: Array<string | { input: string; inject: boolean }> = project.targets?.['build'].options?.styles;
```
goetzrobin, line 77:
```suggestion
	const styles: Array<string | { input: string; inject: boolean }> = project.targets?.['build']?.options?.styles;
```
Might not be needed but "seems" logical to keep the chaining all the way through

libs/ui/accordion/brain/src/lib/brn-accordion.directive.ts
```diff
@@ -120,7 +119,7 @@ export class BrnAccordionDirective implements AfterContentInit, OnDestroy {
 	private readonly _el = inject(ElementRef);
 	private _keyManager?: FocusKeyManager<BrnAccordionTriggerDirective>;
 	private _focusMonitor = inject(FocusMonitor);
-	private _ngZone = inject(NgZone);
+	// private _ngZone = inject(NgZone);
```
goetzrobin, line 122: If it's not needed should we delete this?

libs/ui/avatar/brain/src/lib/image/brn-avatar-image.directive.ts
```diff
@@ -8,15 +8,15 @@ import { Directive, HostListener, computed, signal } from '@angular/core';
 export class BrnAvatarImageDirective {
 	private readonly loaded = signal(false);
 
-	@HostListener('error')
```
goetzrobin, line 11: Feels like this needs to be moved to host in the @Directive also?

libs/ui/checkbox/helm/src/lib/hlm-checkbox.component.ts
```diff
@@ -83,12 +82,12 @@ export class HlmCheckboxComponent {
 	protected readonly _disabled = signal(false);
 	public readonly disabled = input(false, { transform: booleanAttribute });
 
-	private disableInput = effect(
```
goetzrobin, line 86: we can move this to the constructor to avoid the unused variable

goetzrobin:
Closing this for now as we go with eslint/prettier until all features are supported

PR #390, opened 2024-09-29
docs(tooltip): add simple string tooltip example

### LINKED ISSUE #371: Simplified tooltip experience

tooltip

### Information

I think, it would be very nice to add a version of tooltip that can be used as a simple string input directive, without declaration of nested components and additional elements for every tooltip. 
The solution could work as the Angular Material tooltip: https://material.angular.io/components/tooltip/overview.
For most use cases where advanced tooltip customizations are not required this solution would be good enough and include much less code/boilerplate.
Directive input could be used to pass a template, innerHtml content or simple string

_No response_

- [X] Yes
- [X] No

OlegSuncrown:
#371

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [x] tooltip
- [ ] typography

Closes #371

- [ ] Yes
- [x] No

### DIFF
--- a/apps/app/src/app/pages/(components)/components/(tooltip)/tooltip--simple.example.ts
+++ b/apps/app/src/app/pages/(components)/components/(tooltip)/tooltip--simple.example.ts
@@ -1,0 +1,42 @@
+import { Component } from '@angular/core';
+import { HlmButtonDirective } from '@spartan-ng/ui-button-helm';
+import { HlmTooltipComponent, HlmTooltipTriggerDirective } from '@spartan-ng/ui-tooltip-helm';
+
+@Component({
+	selector: 'spartan-tooltip-simple',
+	standalone: true,
+	imports: [
+		HlmTooltipComponent,
+		HlmTooltipTriggerDirective,
+		HlmButtonDirective,
+	],
+	template: `
+		<button [hlmTooltipTrigger]="'Simple tooltip'" aria-describedby="Simple tooltip" hlmBtn variant="outline">
+			Simple
+		</button>
+	`,
+})
+export class TooltipSimpleComponent {}
+
+export const simpleCode = `
+import { Component } from '@angular/core';
+import { HlmButtonDirective } from '@spartan-ng/ui-button-helm';
+import { HlmTooltipComponent, HlmTooltipTriggerDirective } from '@spartan-ng/ui-tooltip-helm';
+
+@Component({
+	selector: 'spartan-tooltip-simple',
+	standalone: true,
+	imports: [
+		HlmTooltipComponent,
+		HlmTooltipTriggerDirective,
+		HlmButtonDirective,
+	],
+	template: \`
+		<button [hlmTooltipTrigger]="'Simple tooltip'" aria-describedby="Simple tooltip" hlmBtn variant="outline">
+			Simple
+		</button>
+	\`,
+})
+export class TooltipSimpleComponent {}
+`;
+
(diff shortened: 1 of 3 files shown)


OlegSuncrown:
Maybe it is better to name something like "Default" to "Template tooltip" and "Simple" to "String tooltip"

goetzrobin:
Thanks for your work on this @OlegSuncrown! 

I think I would go with Simple Tooltip and Custom Tooltip. What do you think?

Also, this is not a blocker for docs, but I am wondering if we should add another selector to the TooltipTrigger so one could do something like [hlmTooltipShowing]="'Simple text goes here'".

What do you think?

OlegSuncrown:
@goetzrobin  Thanks for what you are doing. I can change on Simple Tooltip and Custom Tooltip.

OlegSuncrown:
> Also, this is not a blocker for docs, but I am wondering if we should add another selector to the TooltipTrigger so one could do something like [hlmTooltipShowing]="'Simple text goes here'".
> 
> What do you think?

I think we could rename [hlmTooltipTrigger] on something like [tooltipContent], and do it similar to PrimeNg ?
image

elite-benni:
> Thanks for your work on this @OlegSuncrown!
> 
> I think I would go with Simple Tooltip and Custom Tooltip. What do you think?
> 
> Also, this is not a blocker for docs, but I am wondering if we should add another selector to the TooltipTrigger so one could do something like [hlmTooltipShowing]="'Simple text goes here'".
> 
> What do you think?

This means that an additional input is needed, or is there a trick to use the one input for both selectors? 
To Me the PR looks good.

goetzrobin (approved):

OlegSuncrown:
> > This means that an additional input is needed, or is there a trick to use the one input for both selectors? To Me the PR looks good.
> 
> No I agree that we would need two inputs

@goetzrobin @elite-benni sorry for my late answer. It looks like new input is not required, from what I see, this is already working:

pass template as import

```html
<button [hlmTooltipTrigger]="template" hlmBtn variant="outline">Html</button>

<ng-template #template>
	<div class="flex items-center">
		My html tooltip
		<hlm-icon class="ml-2" size="sm" name="lucidePlus" />
	</div>
</ng-template>
```

pass string as import

```html
<button [hlmTooltipTrigger]="'My string tooltip'"hlmBtn variant="outline">
	String
</button>
```

And if you like this version, I think this <hlm-tooltip></hlm-tooltip> can be removed at all.

PR #391, opened 2024-09-29
chore: update readme docs for bun usage

benpsnyder:


nx.json
```diff
@@ -66,7 +66,7 @@
 			"!{projectRoot}/tsconfig.storybook.json",
 			"!{projectRoot}/eslint.config.js"
 		],
-		"sharedGlobals": []
+		"sharedGlobals": ["{workspaceRoot}/package.json", "{workspaceRoot}/bun.lockb"]
```
goetzrobin, line 69: What does this do?

benpsnyder: If those files change, nx cache is invalidated

Issue #362, opened 2024-08-24
Various issues with lint target


Den-dp:
### Please provide the environment you discovered this bug in.

I was unable to run `npx nx run-many -t lint --all` on Windows

### Which area/package is the issue in?

Don't know / other

### Description

I found 3 issues with it:

1. The current `@nx/eslint` package version doesn't work on Windows (cause it duplicates a path), so it requires at least `nx migrate 19.3.2` where it is fixed https://github.com/nrwl/nx/pull/26687
```
> npx nx lint cli

> nx run cli:lint

Linting "cli"...
 NX   ENOENT: no such file or directory, stat 'C:\Users\denis\github\C:\Users\denis\github\spartan\libs\cli\eslint.config.js'
```
2. after `nx migrate 19.3.2`, I found another issue related to the recent `eslint.config.cjs` rename without renaming all usages, like
https://github.com/goetzrobin/spartan/blob/30bf1fa0159b9534823db66cdc65a756b38089cc/libs/cli/eslint.config.js#L2 
```
> npx nx lint cli

> nx run cli:lint

Linting "cli"...
 NX   Cannot find module '../../eslint.config.js'
Require stack:
- C:\Users\denis\github\spartan\libs\cli\eslint.config.js
```
3. after changing all `eslint.config.js` references to `eslint.config.cjs`, I found that there are some packages with lint errors
```
> nx run ui-forms-brain:lint


Linting "ui-forms-brain"...

C:\Users\denis\github\spartan\libs\ui\forms\brain\package.json
  4:3  error  The "ui-forms-brain" project uses the following packages, but they are missing from "peerDependencies":
    - @angular/forms  @nx/dependency-checks
  6:5  error  The "@spartan-ng/ui-core" package is not used by "ui-forms-brain" project                                                     @nx/dependency-checks
  7:5  error  The "class-variance-authority" package is not used by "ui-forms-brain" project                                                @nx/dependency-checks
  8:5  error  The "clsx" package is not used by "ui-forms-brain" project                                                                    @nx/dependency-checks

...

 NX   Running target lint for 66 projects failed

Failed tasks:

- ui-icon-helm:lint
- ui-tooltip-brain:lint
- ui-forms-brain:lint
- ui-form-field-brain:lint
- ui-select-brain:lint
- ui-tooltip-helm:lint
- ui-input-helm:lint
- ui-form-field-helm:lint
- ui-select-helm:lint
- ui-tabs-helm:lint
[...]

PR #392, opened 2024-09-29
chore: update eslint configuration for lint

benpsnyder:
For [Various issues with lint target #362](#362)


goetzrobin (approved):

goetzrobin:
@benpsnyder do you know why e2e tests are failing for this? Might just be a flaky one since I can't see any changes to component code

apps/ui-storybook-e2e/src/integration/dialog/dialog.cy.ts
```diff
@@ -129,8 +129,6 @@ describe('dialog--default', () => {
 				.findByText(/close nested dialog/i)
 				.click();
 
-			cy.wait(100);
```
elite-benni, line 132: Because of this change is the e2e tests are failing. 
When not waiting for the closing the get('.cdk-overlay-backdrop') finds two occurances. 
I would check if the backdrop only has the length of 1, the cy waits for the closing of the dialog. 

`cy.get('.cdk-overlay-backdrop').should('have.length', 1);`

apps/ui-storybook-e2e/src/integration/dialog/dialog.cy.ts
```diff
@@ -197,8 +195,6 @@ describe('dialog--dynamic-component', () => {
...
```
elite-benni, line 200: same here

elite-benni (changes requested):
replacing wait with `cy.get('.cdk-overlay-backdrop').should('have.length', 1);` resolves the problem of the failing dialog e2e tests.
Then we can merge and go on with the linting errors from point 3.

goetzrobin:
Closing this for now as we go with eslint/prettier until all features are supported

PR #394, opened 2024-10-03
feat: add config read/write to store components path

### LINKED ISSUE #355: Memorize libs directory

Don't know / other

### Information

Probably not very important issue, but I'm using a custom target path for installing components and need to type it every time I run the CLI generator. I know that there is the `--directory` parameter for the CLI, but I love how I can easily install components by c&p'ing from your website, so it would be really helpful if the CLI could just remember the first path I entered.

Enter the target directory everytime.

- [X] Yes
- [ ] No

Balastrong:
Heya, here's a naive implementation of a components.json storing a `componentsPath` to be used to install components without asking for the path each time.
Not particularly engineered, but does its job.

I'm new on the project so I might have missed some standards or common practices, feedback is welcome :)

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

Closes #355

- [ ] Yes
- [ ] No

### DIFF
--- a/libs/cli/src/generators/ui/generator.ts
+++ b/libs/cli/src/generators/ui/generator.ts
@@ -1,11 +1,13 @@
 import { type GeneratorCallback, type Tree, addDependenciesToPackageJson, runTasksInSerial } from '@nx/devkit';
 import { prompt } from 'enquirer';
+import { Config, getOrCreateConfig } from '../../utils/config';
 import { SPARTAN_COLLAPSIBLE_BRAIN_VERSION } from '../base/versions';
 import { addDependentPrimitives } from './add-dependent-primitive';
 import type { HlmUIGeneratorSchema } from './schema';
 
 export default async function hlmUIGenerator(tree: Tree, options: HlmUIGeneratorSchema & { angularCli?: boolean }) {
 	const tasks: GeneratorCallback[] = [];
+	const config = await getOrCreateConfig(tree);
 	const availablePrimitives = await import('./supported-ui-libraries.json');
 	const availablePrimitiveNames = [...Object.keys(availablePrimitives), 'collapsible', 'menubar', 'contextmenu'];
 	let response: { primitives: string[] } = { primitives: [] };
@@ -21,7 +23,7 @@ export default async function hlmUIGenerator(tree: Tree, options: HlmUIGenerator
 		});
 	}
 	tasks.push(
-		...(await createPrimitiveLibraries(response, availablePrimitiveNames, availablePrimitives, tree, options)),
+		...(await createPrimitiveLibraries(response, availablePrimitiveNames, availablePrimitives, tree, options, config)),
 	);
 
 	return runTasksInSerial(...tasks);
@@ -35,6 +37,7 @@ async function createPrimitiveLibraries(
 	availablePrimitives,
 	tree: Tree,
 	options: HlmUIGeneratorSchema & { angularCli?: boolean },
+	config: Config,
 ) {
 	const allPrimitivesSelected = response.primitives.includes('all');
 	const primitivesToCreate = allPrimitivesSelected ? availablePrimitiveNames : response.primitives;
@@ -72,7 +75,7 @@ async function createPrimitiveLibraries(
 				primitiveName: '',
 				peerDependencies,
 				skipBrainDependencies: options.skipBrainDependencies,
-				directory: options.directory,
+				directory: config.componentsPath,
 				tags: options.tags,
 				rootProject: options.rootProject,
 				angularCli: options.angularCli,
--- a/libs/cli/src/generators/ui/generator.ts
+++ b/libs/cli/src/generators/ui/generator.ts
@@ -75,7 +75,7 @@ async function createPrimitiveLibraries(
 				primitiveName: '',
 				peerDependencies,
 				skipBrainDependencies: options.skipBrainDependencies,
-				directory: config.componentsPath,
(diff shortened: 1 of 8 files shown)


libs/cli/src/generators/ui/schema.d.ts
```diff
@@ -1,6 +1,5 @@
 export interface HlmUIGeneratorSchema {
 	name?: string;
-	directory?: string;
```
elite-benni, line 3: I don't think it is a good idea to remove the directory from the schema. 
If someone uses the directory param in the CLI this would not work anymore.
I would rather remove the x-prompt and react if the directory is not set.
Also i would then just set the options.directory from the config.

libs/cli/src/utils/config.ts
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	console.log('Configuration file not found, creating a new one...');
+
+	const { componentsPath } = (await prompt({
+		type: 'input',
+		required: true,
+		name: 'componentsPath',
+		message: 'Choose a directory to place your spartan libraries, e.g. libs/ui',
+		initial: 'libs/ui',
```
elite-benni, line 22: I like the initial value, because some people assume that libs/ui is the default if they don't type it out!

elite-benni (changes requested):
This PR improves the DX a lot. 
Would like to have some documentation about the component.json too, so that people know how to change the directory if needed. (e.g. if they had a typo in the first try of generating the components)

libs/cli/src/generators/ui/schema.d.ts
Balastrong: That's right! I'll add this back, just without the prompt :)

Balastrong:
Thanks @elite-benni for the review! I added create a new page for components.json in case we're going to add new configs later :)

libs/cli/src/generators/ui/generator.ts
```diff
@@ -1,11 +1,13 @@
 import { type GeneratorCallback, type Tree, addDependenciesToPackageJson, runTasksInSerial } from '@nx/devkit';
 import { prompt } from 'enquirer';
+import { Config, getOrCreateConfig } from '../../utils/config';
 import { SPARTAN_COLLAPSIBLE_BRAIN_VERSION } from '../base/versions';
 import { addDependentPrimitives } from './add-dependent-primitive';
 import type { HlmUIGeneratorSchema } from './schema';
 
 export default async function hlmUIGenerator(tree: Tree, options: HlmUIGeneratorSchema & { angularCli?: boolean }) {
 	const tasks: GeneratorCallback[] = [];
+	const config = await getOrCreateConfig(tree);
```
elite-benni, line 10: It may be confusing, when i set the --directory param, and then have to specify the directory again. 
I know it is just the first time, if the components.json does not exist. 
Maybe we can skip that step if the directory is set through options.

elite-benni (commented):
Thanks for adding documentation. 
One small comment for a edge case. ;)
LGTM. 
Would like to have @goetzrobin a look over it, as he wrote most of the generators.

libs/cli/src/generators/ui/generator.ts
Balastrong: That's right!
If we think config will have more fields we shouldn't skip this entirely but maybe only skip the step where we ask the directory.

What do you want for now when --directory is provided and config isn't there yet? A couple of options:
A) We entirely skip the config generation (this must be changed when we add more options in the config)
or
B) We try to create the config anyway, but instead of asking the directory we use the provided value from the params (this means today we skip showing prompts entirely, but when we'll add one more no other changes will be required)

elite-benni: Oh, that is a valid point. 
I'd vote for B

Balastrong: I see enquirer (the prompt library) has a `skip` property which also picks the `initial` value if skip is true, exactly what we needed! :)

elite-benni (approved):
LGTM! Thanks!

goetzrobin:
@Balastrong thanks for this PR. we improved the side nav a little and it causes conflicts. this is just for docs and should be an easy fix! happy to merge once resolved!

Great work and thanks so much!

Balastrong:
Thanks! Conflict resolved :)

Balastrong:
Hey! I see the CI fails but seems to be related to other files, is there anything I should do here or is it good? :)

PR #455, opened 2024-11-08
refactor(progress): migration to signals

### LINKED ISSUE #431: refactor(progress): migration to signals

progress

### Information

Migrate away from all Input/Output and View Decorators

_No response_

- [ ] Yes
- [ ] No

ashley-hunter:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [X] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

Closes #431 

- [ ] Yes
- [X] No

## Other information

- Migrating progress components/directives to use signals
- Adding unit tests
- Updating to use provider to retrieve value instead of querying the dom in ngDoCheck as that is quite an expensive task in such a frequently run function.

### DIFF
--- a/libs/ui/progress/brain/src/lib/brn-progress.component.ts
+++ b/libs/ui/progress/brain/src/lib/brn-progress.component.ts
@@ -1,60 +1,65 @@
-import { type NumberInput, coerceNumberProperty } from '@angular/cdk/coercion';
-import { Component, Input, computed, signal } from '@angular/core';
+import { type NumberInput } from '@angular/cdk/coercion';
+import { Component, OnChanges, SimpleChanges, computed, input, numberAttribute } from '@angular/core';
+import { provideBrnProgress } from './brn-progress.token';
 
 @Component({
 	selector: 'brn-progress',
 	standalone: true,
 	template: '<ng-content/>',
+	exportAs: 'brnProgress',
+	providers: [provideBrnProgress(BrnProgressComponent)],
 	host: {
 		role: 'progressbar',
-		'[attr.aria-valuemax]': '_max()',
+		'[attr.aria-valuemax]': 'max()',
 		'[attr.aria-valuemin]': '0',
-		'[attr.aria-valuenow]': 'isNumber() ? _value() : undefined',
-		'[attr.aria-valuetext]': '_value() ? getValueLabel(_value(),_max()) : undefined',
-		'[attr.data-state]': 'progressState()',
-		'[attr.data-value]': '_value() ?? undefined',
-		'[attr.data-max]': '_max()',
+		'[attr.aria-valuenow]': 'value()',
+		'[attr.aria-valuetext]': 'label()',
+		'[attr.data-state]': 'state()',
+		'[attr.data-value]': 'value()',
+		'[attr.data-max]': 'max()',
 	},
 })
-export class BrnProgressComponent {
-	protected readonly _value = signal<number | null | undefined>(undefined);
-	public readonly $value = this._value.asReadonly();
-	@Input()
-	set value(newValue: NumberInput) {
-		if (newValue === undefined || newValue === null || newValue === 'null' || newValue === 'undefined') {
-			this._value.set(null);
+export class BrnProgressComponent implements OnChanges {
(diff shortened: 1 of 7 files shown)


libs/ui/progress/brain/src/lib/brn-progress.component.ts
```diff
@@ -1,60 +1,65 @@
...
+		return value === null || value === undefined ? 'indeterminate' : value === max ? 'complete' : 'loading';
+	});
+
+	ngOnChanges(changes: SimpleChanges): void {
+		if ('value' in changes || 'max' in changes) {
+			this.validate();
+		}
+	}
+
+	private validate(): void {
```
ajitzero, line 46: suggestion/question: (possibly something to be picked up separately, if at all)

I know this is the existing behaviour here to throw a runtime error for invalid input, but is that worth using ngOnChanges? My mental model for change detection is far from perfect, and since we don't have `OnPush` here for some reason, I don't know how often the lifecycle hook runs.

I'm considering having an `isValid` computed signal, that is checked within an `effect` to throw errors in [`isDevMode](https://angular.dev/api/core/isDevMode)` only.

What do you think?

e.g.:
```ts
private readonly validationErrors = computed(() => {
  const value = this.value();
  const max = this.max();

  if (value === null || value === undefined) {
    return null;
  }
  if (value > max || value < 0) {
    return 'Value must be 0 or greater and less or equal to max';
  }
  if (max < 0) {
     return 'max must be greater than 0';
  }
  return null;
})

effect(() => {
  const errorMessage = this.validationErrors();
  if (errorMessage && isDevMode) {
    throw new Error(errorMessage);
  }
})
```

ashley-hunter: Yes, so I was actually considering something very similar. ngOnChanges fires when any input changes (regardless of OnPush), so this would only throw an error whenever value or max input changes. I absolutely could use effect, but I chose not to because effect is not stable yet, that being said, I'm not specifically against using it as it's not writing to any signals in this case.

I was retaining existing behaviour, but alternatively we don't need to throw any errors, we could just clamp the value so it would always be within the valid range.

ajitzero: That's reasonable, and I don't know the original context behind this either. I did see effects being used in a few other places, so I thought of following that, but it's fine. This is, anyhow, something the developer is responsible for.

Also, how are you online 24/7? 😅

ajitzero (approved):

libs/ui/progress/brain/src/lib/brn-progress.component.ts
ashley-hunter: > I did see effects being used in a few other places

Yeah, I'm not against effects in the right place. They still seem to be figuring out the behaviour about writing to signals from effects which they have went back and forward about a few times, but regardless this doesn't apply here. I could use change to use an effect if anyone has any strong opinions, but behaviourally they will work the same as the signals never get updated within the component, only externally.

> Also, how are you online 24/7?

image

😂 I work a bit in the evening, and the rest of the day I just reply on the go.

PR #480, opened 2024-11-13
feat(calendar): calendar component

### LINKED ISSUE #111: RFC: Calendar

calendar

### Information

TBD

_No response_

- [X] Yes
- [ ] No

ashley-hunter:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [X] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

Closes #111 

- [ ] Yes
- [X] No


ashley-hunter:
@thatsamsonkid it would be great to unify our calendar implementations and get this feature over the finish line. Im happy to continue whatever way you think is best, let me know if you have any ideas on what way you think we should proceed 😊

thatsamsonkid:
I'm good to merge your implementation first especially since this is in a much more complete state. The main thing I would be grabbing from my PR is the "material" like navigation of the calendar and I think from that aspect pulling that over to this might not be too bad I think since as far as I can tell our main helm component is laid out somewhat similarly.  Plus I still need to write tests for all that 😅, so I'll just add it as an enhancement on top of this

ashley-hunter:
> I'm good to merge your implementation first especially since this is in a much more complete state. The main thing I would be grabbing from my PR is the "material" like navigation of the calendar and I think from that aspect pulling that over to this might not be too bad I think since as far as I can tell our main helm component is laid out somewhat similarly. Plus I still need to write tests for all that 😅, so I'll just add it as an enhancement on top of this

Sounds good, perhaps since you have been working in this area, perhaps you could review these changes and let me know if you have any feedback - or if there is anything you think in the API needs changed to accommodate your additional functionality, that would be great 😊

> On another note, don't mean to add more work to this PR but should we just go ahead and make this a secondary entry of the lib or should we do it after while wait till we figure out the linting situation?

Good point - I will perhaps start work on the secondary entrypoint setup tonight - I'll do it as part of a separate PR as I don't fully know the extent of the changes and this PR is already quite large. I think I will go ahead with ESLint as part of that setup as that is the default in the Nx generator, if we decide Biome is the desired go forward approach I can always just switch it out - its easier to remove ESLint, than it is to add it. If the secondary entrypoint PR is merged first then I will update this PR to use secondary entrypoints, if this PR is merged first I will raise a follow on PR to move it. Does that sound reasonable to you?

libs/ui/calendar/helm/src/lib/hlm-calendar.component.ts
```diff
@@ -0,0 +1,148 @@
...
+							[attr.aria-label]="i18n.labelWeekday(weekday)">
+						{{ i18n.formatWeekdayName(weekday) }}
+					</th>
+				</tr>
+			</thead>
+
+			<tbody role="rowgroup">
+				<tr *brnCalendarWeek="let week" class="flex w-full mt-2">
+					@for (date of week; track dateAdapter.getTime(date)) {
+					<td brnCalendarCell class='h-9 w-9 text-center text-sm p-0 relative [&:has([aria-selected].day-range-end)]:rounded-r-md data-[selected]:data-[outside]:bg-accent/50 data-[selected]:bg-accent first:data-[selected]:rounded-l-md last:data-[selected]:rounded-r-md focus-within:relative focus-within:z-20'>
```
goetzrobin, line 84: We are missing the brnCalendarCell directive import here

libs/ui/calendar/brain/src/index.ts
```diff
@@ -0,0 +1,37 @@
+export * from './lib/brn-calendar-cell-button.directive';
```
goetzrobin, line 1:
```suggestion
export * from './lib/brn-calendar-cell.directive';
export * from './lib/brn-calendar-cell-button.directive';
```
We need to export BrnCalendarCellDirective here to use it in the HlmCalendarComponent

libs/ui/calendar/brain/src/index.ts
```diff
@@ -0,0 +1,37 @@
...
+export * from './lib/brn-calendar-next-button.directive';
+export * from './lib/brn-calendar-previous-button.directive';
+export * from './lib/brn-calendar-week.directive';
+export * from './lib/brn-calendar-weekday.directive';
+export * from './lib/brn-calendar.directive';
+export * from './lib/brn-calendar.token';
+export * from './lib/i18n/calendar-i18n';
+
+import { NgModule } from '@angular/core';
+import { BrnCalendarCellButtonDirective } from './lib/brn-calendar-cell-button.directive';
```
goetzrobin, line 13:
```suggestion
import { BrnCalendarCellDirective } from './lib/brn-calendar-cell.directive';
import { BrnCalendarCellButtonDirective } from './lib/brn-calendar-cell-button.directive';
```

libs/ui/calendar/brain/src/index.ts
```diff
@@ -0,0 +1,37 @@
...
+import { BrnCalendarGridDirective } from './lib/brn-calendar-grid.directive';
+import { BrnCalendarHeaderDirective } from './lib/brn-calendar-header.directive';
+import { BrnCalendarNextButtonDirective } from './lib/brn-calendar-next-button.directive';
+import { BrnCalendarPreviousButtonDirective } from './lib/brn-calendar-previous-button.directive';
+import { BrnCalendarWeekDirective } from './lib/brn-calendar-week.directive';
+import { BrnCalendarWeekdayDirective } from './lib/brn-calendar-weekday.directive';
+import { BrnCalendarDirective } from './lib/brn-calendar.directive';
+
+export const BrnCalendarImports = [
+	BrnCalendarCellButtonDirective,
```
goetzrobin, line 23:
```suggestion
	BrnCalendarCellDirective,
	BrnCalendarCellButtonDirective,
```

libs/ui/calendar/brain/src/lib/brn-calendar-header.directive.ts
```diff
@@ -0,0 +1,17 @@
+import { Directive, input } from '@angular/core';
+
+let uniqueId = 0;
+
+@Directive({
+	selector: '[brnCalendarHeader]',
+	standalone: true,
+	host: {
+		'[id]': 'id()',
+		'[attr.aria-live]': 'polite',
```
goetzrobin, line 10:
```suggestion
		'[attr.aria-live]': '"polite"',
```

libs/ui/calendar/brain/src/lib/brn-calendar.directive.ts
```diff
@@ -0,0 +1,206 @@
...
+	/** The day the week starts on */
+	public readonly weekStartsOn = input<Weekday, NumberInput>(0, {
+		transform: (v: unknown) => numberAttribute(v) as Weekday,
+	});
+
+	/** The default focused date. */
+	public readonly defaultFocusedDate = input<T>();
+
+	/** @internal Access the header */
+	protected readonly header = contentChild(BrnCalendarHeaderDirective);
```
goetzrobin, line 63:
```suggestion
	public readonly header = contentChild(BrnCalendarHeaderDirective);
```

We need this because we access this variable in [brn-calendar-grid.directive.ts](https://github.com/spartan-ng/spartan/pull/480/files#diff-583e25e213df9cf5c9db39f56be34a967d0879838a7cafae7fe22b12047fa5ad)

goetzrobin (changes requested):
A few minor comments! Looks great! Thanks so much!

ashley-hunter:
Thanks for the feedback @goetzrobin, well spotted! I have made the changes.

thatsamsonkid:
@ashley-hunter 1 small thing can you check the outline on calendar compared to shadcn. It's showing as white in preview site

ashley-hunter:
Thanks @thatsamsonkid, looks like we were missing some styles in the global stylesheets so the border color was never getting set in the documentation or storybook. Should now be resolved.

apps/app/src/styles.css
```diff
@@ -198,3 +198,9 @@ select {
 	--input: 240 3.7% 15.9%;
 	--ring: 142.4 71.8% 29.2%;
 }
+
+@layer base {
+	* {
+		@apply border-border;
```
thatsamsonkid, line 204: @goetzrobin So i remember having a similar issue with select and I had added this previously but I think you had ended up removing it and done something else. I don't recall how you fixed it do you remember? I also never asked at the time but what was the issue with adding this?

thatsamsonkid: Ah ok so I found my original change so I had originally done this because this is what I saw in shadcn repo

* {
	border-color: hsl(var(--border));
}

@goetzrobin removed it and applied border-border to select directly. Your rule is instead just applying that same class in this case so maybe this is will work better?

Linking the change in case it helps jog any memory if there was an issue caused by this https://github.com/thatsamsonkid/spartan/pull/7/files#r1496196592

ashley-hunter: I had updated it to match the recommended global styles on the Shadcn site:

https://ui.shadcn.com/docs/installation/manual

It adds it in a base layer so that it can easily be overridden, but uses it as the default. I can change it to only target the calendar if you prefer.

thatsamsonkid: Yea if you want we can scope it till we find out if there's an issue with the above rule. I don't want to put off this PR just because of that but I'm just more curious why we're not able to have this same rule in our project.

goetzrobin: Yeah originally I did it so border color isn't assigned by default and explicit instead of implicit. Might be more confusing though because it's not in alignment with shadcn?

ashley-hunter: Yeah, my thoughts were to align it with Shadcn, but I can easily switch it if needed, so if anyone has a preference just let me know 😊

thatsamsonkid: I agree I think we should align it

thatsamsonkid (approved):
Lets merge!

PR #506, opened 2024-12-02
build: add pnpm run release script

goetzrobin:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

Robin does some manual search and replace and runs a bunch of commands in his shell

Closes #

## What is the new behavior?

You can set the following environment variables

```
NPM_TOKEN=npm_AND_SO_ON
TAG=latest
VERSION=0.0.1-alpha.370
```
and then execute
```
pnpm run release
```
and it will set the right version for the UI libs and CLI and release them to NPM

- [ ] Yes
- [ ] No


goetzrobin:
I tried setting up semantic-release, but you are forced to start with v1.0.0.
Brandon suggested to just manually update the version until that point.
I think we can also add some logic for the script to automatically increase the alpha version by 1 and set up a nightly build.

Curious to get your feedback and hear your thoughts!
@thatsamsonkid @elite-benni @ashley-hunter

libs/tools/src/executors/release/build-update-publish/executor.spec.ts
```diff
@@ -19,7 +18,6 @@ describe('BuildUpdatePublish Executor', () => {
 		jest.spyOn(projectHelper, 'getProjectName').mockReturnValue(libName);
 
 		// Mock updateVersion and npmPublish to return { success: true }
```
elite-benni, line 20: The comment is now confusing when updateVersion is not here anymore.

libs/tools/src/generators/replace-ui-version/generator.ts
```diff
@@ -0,0 +1,76 @@
...
+	return results;
+}
+
+const getSpartanDependencyKeys = (dependencies?: Record<string, string>): string[] =>
+	Object.keys(dependencies ?? {}).filter((key) => key.startsWith('@spartan-ng'));
+
+const replaceUiVersionInCliVersionsFile = (tree: Tree, oldVersion: string, newVersion: string) => {
+	const filePath = `libs/cli/src/generators/base/versions.ts`;
+	let contents = tree.read(filePath).toString();
+	contents = contents.replaceAll(oldVersion, newVersion);
```
elite-benni, line 27: It is not the case now, but does this have the potential to replace versions of other packages too in the future if some other const has the same version as spartan oldVersion?

release.config.mjs
```diff
@@ -26,15 +25,21 @@ export default {
...
+				prepareCmd: 'TAG=latest,VERSION=${nextRelease.version} pnpm run pre-release',
+				releaseCmd: 'TAG=latest,VERSION=${nextRelease.version} pnpm run release',
 			},
 		],
 		[
 			'@semantic-release/git',
 			{
-				assets: ['libs/**/package.json', 'package.json', 'CHANGELOG.md'],
-				message: 'chore(release): -v${nextRelease.version} [skip ci]\n\n${nextRelease.notes}',
+				assets: [
```
elite-benni, line 35: is the root package.json not changed anymore?

elite-benni (commented):
Just minor comments / questions. 

To answer your question: 
I'd suggest manually updating until release 1.0.0 is fine. 
I hope, that it doesn't take too long to get to stable and with these scripts set up, the manual release should not be too much of an effort?
Also I am not sure how this interferes with the changes of the secondary entrypoints.

release.config.mjs
goetzrobin: It never was, but we could store the current version as a source of truth there. I think keeping CLI and UI in sync also reduces additional complexity as we move forward

libs/tools/src/generators/replace-ui-version/generator.ts
goetzrobin: Good point. We might want to make this a little smarter and only change versions with spartan in their variable name

libs/tools/src/executors/release/build-update-publish/executor.spec.ts
goetzrobin: Good point! I'll remove it

libs/tools/src/generators/replace-ui-version/generator.ts
goetzrobin: @elite-benni added some regex to handle this better and covered it by a unit test to make sure we don't mess up other versions as you pointed out 👍

goetzrobin:
> Just minor comments / questions.
> 
> To answer your question: I'd suggest manually updating until release 1.0.0 is fine. I hope, that it doesn't take too long to get to stable and with these scripts set up, the manual release should not be too much of an effort? Also I am not sure how this interferes with the changes of the secondary entrypoints.

I am working on a follow up PR that would also set up an automated daily release, so we would be set up there.

@ashley-hunter please let me know if there's anything for me to keep in mind to not have this conflict with your secondary entry-points PR.

ashley-hunter:
> > Just minor comments / questions.
> > To answer your question: I'd suggest manually updating until release 1.0.0 is fine. I hope, that it doesn't take too long to get to stable and with these scripts set up, the manual release should not be too much of an effort? Also I am not sure how this interferes with the changes of the secondary entrypoints.
> 
> I am working on a follow up PR that would also set up an automated daily release, so we would be set up there.
> 
> @ashley-hunter please let me know if there's anything for me to keep in mind to not have this conflict with your secondary entry-points PR.

Apologies, haven't had a chance to deep dive into the PR yet, but from the quick look I have had I don't think there is too much to keep in mind - the only real change in the end is there will be one brain package rather than many, so I guess if anything it will just result in less versions needing updated in places.

While on the topic - but not to derail - the secondary entrypoints PR currently sets up the library and has moved across one brain library (checkbox). Would you rather every library just be moved in one large PR, or separate PRs like how we've been approaching signals? For example, if you're happy to do them in separate PRs, then the initial setup and checkbox are migrated and could be merged if you want to verify your release setup with it - if you'd rather it all in one go we can always do a follow on pass at the release scripts once everything is migrated to verify if anything needs changed or can be cleaned up.

I notice one discussion point on CLI vs UI version - I too think it might be a good idea to align these in future, makes it easy to know if versions are compatible with each other, especially in cases where you maybe want to downgrade or upgrade but not to the latest version its easy to know which version of each needs to be installed.

goetzrobin:
> > > Just minor comments / questions.
> > > To answer your question: I'd suggest manually updating until release 1.0.0 is fine. I hope, that it doesn't take too long to get to stable and with these scripts set up, the manual release should not be too much of an effort? Also I am not sure how this interferes with the changes of the secondary entrypoints.
> > 
> > 
> > I am working on a follow up PR that would also set up an automated daily release, so we would be set up there.
> > @ashley-hunter please let me know if there's anything for me to keep in mind to not have this conflict with your secondary entry-points PR.
> 
> Apologies, haven't had a chance to deep dive into the PR yet, but from the quick look I have had I don't think there is too much to keep in mind - the only real change in the end is there will be one brain package rather than many, so I guess if anything it will just result in less versions needing updated in places.
> 
> While on the topic - but not to derail - the secondary entrypoints PR currently sets up the library and has moved across one brain library (checkbox). Would you rather every library just be moved in one large PR, or separate PRs like how we've been approaching signals? For example, if you're happy to do them in separate PRs, then the initial setup and checkbox are migrated and could be merged if you want to verify your release setup with it - if you'd rather it all in one go we can always do a follow on pass at the release scripts once everything is migrated to verify if anything needs changed or can be cleaned up.
> 
> I notice one discussion point on CLI vs UI version - I too think it might be a good idea to align these in future, makes it easy to know if versions are compatible with each other, especially in cases where you maybe want to downgrade or upgrade but not to the latest version its easy to know which version of each needs to be installed.

I am fine with doing one by one. Do we still release them under their current package or would they simply be moved to the brain package and the primitive specific package is discontinued?

ashley-hunter:
> I am fine with doing one by one. Do we still release them under their current package or would they simply be moved to the brain package and the primitive specific package is discontinued?

In its current state the individual primitive libraries are removed entirely, however I have added a generator/schematic that someone can run to automatically update any imports etc.. to the new equivalent. That being said, if you still want to preserve the old packages we could add an additional pre-release step to generate any removed packages in the dist folder and they would simply re-export from the secondary entrypoint. But it depends whether you want to preserve them or whether the generator for migration is sufficient.

goetzrobin:
@ashley-hunter l decided to go with pushing for secondary entry points ASAP! Especially, since you have created such a great migration experience where they can just run a simple command and take advantage of the new structure! I'll adjust this PR to take that into consideration! Thanks so much!

ashley-hunter:
@goetzrobin no problem! Thank you! I'll begin migrating other entry points over the next few days. I have a generator to take care of most of it automatically, so I don't suspect it should take too long to move over entirely.

goetzrobin:
@ashley-hunter @elite-benni released a new version based on these changes. this should get us pretty close to automatic nightly releases!

ashley-hunter:
> @ashley-hunter @elite-benni released a new version based on these changes. this should get us pretty close to automatic nightly releases!

Brilliant!! Great work, this will be an excellent improvement 😊

elite-benni (approved):

PR #537, opened 2024-12-20
feat(icon): icon component migration

ashley-hunter:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [X] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

Currently we wrap the NgIcon component which adds a hard dependency on the library and more specifically can lead to us being tied to a specific version and make it hard to update if there are ever any breaking changes. 

This pull request changes the component we had to a directive that applies the little bit of custom styling we add, this removes the hard dependency on any specific version of NgIcons.

## What is the new behavior?

We use the ng-icon component directly and add an attribute for the styling

- [X] Yes
- [ ] No

## Other information

I have added an automated migration to assist in the process of migration as this is a breaking change


ajitzero:
Hi @ashley-hunter - Considering that the end user "owns" the code for helmet components locally (shadcn-style), I'm not sure I understand this point:

> - adds a hard dependency on the library
> - can lead to us being tied to a specific version
> - hard to update if there are ever any breaking changes

As the end-user, I think I can upgrade whatever I want directly since this is in my local codebase. Or does the new combined brain package have some overlap here? I think that's part of the reason for shadcn/ui's success: direct access to styling + ownership for upgrades/modifications.

I do prefer attribute selectors as they reduce one layer of usually unnecessary host elements. But again, this exists locally for the developer, and I feel it would be better to have a component. That way, if I wanted to, I could replace ng-icon itself with some alternative icon library (though I don't know what yet - ng-icon is pretty comprehensive - great work on that!).

That lets me avoid modifying those usage instances, like in this MR. Similarly, `provideIcons` should potentially be wrapped with `provideHlmIcons` to support switching to other providers internally from a single place.

ashley-hunter:
Hey @ajitzero, sure so to give you a bit of background we have been trying to reduce our hard dependencies on third party library as it can helps reduce friction with updates or libraries becoming unmaintained etc... we have recently moved the scroll area to use an approach similar to this and ng-icons was discussed as being the next potential candidate (https://github.com/spartan-ng/spartan/pull/529#issuecomment-2548306510), commander is another one on the list.

You are absolutely correct, and I guess I did not describe it particularly well in the description above. You are correct, this is most important for brain libraries where the user is not in control of the dependency versions etc.. those scenarios cause many more headaches.

However, there are things to consider on the helm side too when it comes to third party libraries, and these apply to many libraries, not just ng-icons.

Wrapping a component means we have to manually forward all inputs/outputs, if new ones are added we need to update the component generator to have them, but different users may be using different versions of that library (especially as some versions are only compatible with specific Angular versions). So now our generator needs to be aware of exact versions of that library being used, otherwise if we try to use an input that doesn't exist in their version we get a compiler error. The same goes for inputs/outputs that are renamed or removed. We also may have behavioural or type changes with some inputs/outputs across different versions. This could greatly complicate the component generator trying to support all scenarios, and the upgrade path too.

By wrapping a component this also encourages divergent behaviour, and this is still technically present in the icon directive in this PR, where we add some pre-defined custom sizes. In this case I have control over the ng-icon library so I don't plan on causing such a breaking change, but if the method in which sizes are set changed in the underlying library, we have "hacked" in additional functionality to a component that may not be possible in future versions or break at any given time.

Wrapping a component also introduces an additional DOM element, hiding away the underlying element. This can lead to layout issues in some cases, or prevents you from adding attributes to the underlying element which may be necessary for events or accessibility purposes, or many other cases. A prime example of this is we had to have a `ngIconClass` input on the wrapper component to forward any classes to the inner element.

Many of these libraries also have more than one component, so for example, ng-icons also supports an `ng-glyph` component, and `ng-icon-stack`, we don't currently wrap those, or perhaps the icon wouldn't work in an `ng-icon-stack` because of the additional dom element and someone wants to use it, do we now have to wrap `ng-icon-stack` too?

If someone encounters a bug in an underlying library, because it is branded as a helm component, we may get the bug report when in fact the issue is with the third party library. If we don't wrap it or abstract it, this is more obvious and bugs get reported to the correct library.

As a result, for the most part we should only try to apply the necessary styling where possible, and let the underlying libraries handle everything else to avoid any of these issues. This will help reduce the maintenance for us as well as it is less for us to support.

> That way, if I wanted to, I could replace ng-icon itself with some alternative icon library (though I don't know what yet - ng-icon is pretty comprehensive - great work on that!). That lets me avoid modifying those usage instances, like in this MR. Similarly, provideIcons should potentially be wrapped with provideHlmIcons to support switching to other providers internally from a single place.

Thank you btw 😉 I'm not sure this would really be much each easier with a wrapper component either however, all the inputs may be different, even if they had similar inputs, they would likely be using different types. And `provideIcons` accepts a list of icons from `ng-icons` libraries, wrapping the provider wouldn't help here because if the underlying icon library was switched out the icons supplied wouldn't work with a different icon library, and many icon libraries don't use the same approach for registering icons either?

If we don't feel this approach works, I'm happy to leave it as is, however it is important for us to consider the potential impacts of wrapping components going forward, and while we are in alpha we have the opportunity to make such changes which won't really be feasible for much longer as we move toward v1.

ajitzero (approved):

ashley-hunter:
Thankfully we haven't really encountered these kinds of issues as of yet, but it is inevitable so best to prevent them where possible before they happen and it's hard to change.

I completely agree, it would be great if we could spread inputs/outputs and attributes and have fragment style components. Maybe we will get something like this when selectorless lands 🤷‍♂️ kind of like the old AngularJS days where the host element was optional.

goetzrobin:
@ajitzero are you in the discord? I can add you to the contributor channel as you have been a spartan from very early on and it would be great to have you be part of those discussions!

goetzrobin:
@ashley-hunter I think we need to rerun your migration after the radio button changes got merged which still rely on the wrapper component

ashley-hunter:
@goetzrobin, thank you! Will do that now!

ashley-hunter:
Updated now 😊 thank you

PR #542, opened 2024-12-24
chore: add .nvmrc file for Node.js version 22.9.0

JeevanMahesha:
## PR Checklist

---

## PR Type

- [ ] Bugfix  
- [ ] Feature  
- [ ] Refactoring (no functional changes, no API changes)  
- [x] Build related changes  
- [ ] CI related changes  

---

- [x] None (This PR doesn't modify a specific package, it updates the overall project setup.)

---

Currently, there is no `.nvmrc` file in the repository, which can lead to inconsistencies in the Node.js version used by contributors.

Closes: N/A

---

## What is the new behavior?

The `.nvmrc` file has been added to specify the recommended Node.js version, ensuring that all contributors use the same version during development.

---

- [ ] Yes  
- [x] No  

---

## Other information

This is an initial step to improve developer experience by introducing Node.js version management through `.nvmrc`.


.nvmrc
```diff
@@ -0,0 +1 @@
+22.9.0
```
ajitzero, line 1: Thanks for this! I just noticed that we don't mention anything like this in the contributing guide, etc., but should have been added. Maybe with a different PR?

A couple of points:

1. Please include a `.node-version` file with the same contents. Thanks to fragmentation between node version managers, every tool except `nvm` looks for a `.node-version` file. While `nvm` is stable on macOS/Linux, it often enters a problematic state in Windows, so many people have switched to other alternatives.
2. I can't recall where I remember this from, but `.nvmrc` (and `.node-version`) should only include the major version to allow for easy upgrades through minor versions.
```suggestion
22
```

JeevanMahesha: @ajitzero  Thank you for the feedback! 

I’d like to address the points you raised:

1. **Regarding `.node-version`**: We already have a `.node-version` file for Windows users, and this PR extends support for macOS and Linux users by introducing `.nvmrc`. This ensures broader compatibility across platforms while keeping the workflow seamless for contributors.

2. **Specific Version vs. Major Version**: 
   I understand your suggestion to use only the major version for flexibility and ease of upgrades. However, for open-source projects, using a specific version often ensures consistency across development, CI pipelines, and production environments. This avoids potential issues such as:
   - Non-reproducible bugs if contributors or CI pipelines use different patch or minor versions.
   - Unexpected regressions or deprecations introduced in minor or patch updates, which can disrupt the workflow.

While I see the value in using the major version for personal or fast-moving projects, I believe specifying an exact version is more suitable for open-source projects where consistency and reproducibility are critical.

Let me know if you'd like to discuss further or if you'd prefer a separate PR for `.node-version` adjustments. 

Thanks again for your thoughtful review!

ajitzero: > 1. We already have a `.node-version` file

Whoops! I missed that.

> This avoids potential issues such as

So again, I'm really not sure what/who I'm recalling when I say we should only save the major version, so feel free to ignore that part.

The minor versions in Node have been very stable over time and _ideally_ shouldn't introduce any breaking changes. I think its fine if we leave it in though.

---

I'm now considering that nvm users should just symlink `.node-version` to `.nvmrc` instead us maintaining two files for the same thing. A lot of better alternatives exist now, [which all support `.node-version`](https://github.com/shadowspawn/node-version-usage#compatibility-testing).

While the maintainer behind `nvm` refuses to support a standard format ([due to lack of personal bandwidth](https://github.com/nvm-sh/nvm/issues/794)), I guess this project probably doesn't use nvm either, since we have a .node-version file in here.

I personally use `fnm`, but maybe others could chime in? Instead, we could update the contributing guide to mention our preferred node version management tool. / @goetzrobin @ashley-hunter @thatsamsonkid

ashley-hunter: Yeah, the issue seems to be there isn't really a standard, I used NVM for a while, but recently a macOS update made it significantly slower. As a result I switched to Volta, however Volta uses a different convention for specifying the node version again, so adding more and more configs doesn't really seem like a great solution.

We do already define the required version in the engines field in package.json:

https://github.com/spartan-ng/spartan/blob/d7f1338aa062198dccbc482299c2e0a434f0c5f8/package.json#L188

This should throw an error when trying to run PNPM install if you are using an unsupported version, which would result in the user having to use the correct node version in whatever node version manager they use - if any.

So perhaps that is enough? Maybe just add a section to the contributing guide specifying the required version?

goetzrobin: I agree with @ashley-hunter here. Ultimately it's up to the developer to decide which tool they want to use to get the right node version to run the project. I think we can add a section to the contributing guide and if you'd like to describe your setup with `nvm`

dexxiez: Let CI/CD checks have final say, easy

JeevanMahesha:
After review and considering the current project setup, I’ve decided to close this PR. The existing configuration, which specifies the required Node.js versions in the engines field of package.json, already effectively handles the use case this PR aimed to address.

JeevanMahesha:
After review and considering the current project setup, I’ve decided to close this PR. The existing configuration, which specifies the required Node.js versions in the engines field of package.json, already effectively handles the use case this PR aimed to address.

Thank you.

PR #550, opened 2025-01-09
fix: code quality linting improvements

ashley-hunter:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [X] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

N/A

## What is the new behavior?

- Following the discussion on discord earlier where linting errors may occur in an Nx repo, I have went through our code and fixed some areas where eslint was disabled due to things like poor types etc.. This is a pre-cursor to fixing the reported issue, but should reduce the custom settings required to be any shared config we ship.

- Fixing an eslint config using old format.

- I have also removed a checkbox icon component as it can be inlined to reduce amount of code required - and we previously had icon class and icon name inputs for the checkbox. These don't seem very useful and normally the check icon would be the same in all checkboxes. In the rare case that this would need to be changed, the code can be modified in the project as it is a helm component, but it seems cleaner not to pollute to component API for a very rare case that has a simple workaround - I can revert if there is a valid case Im not aware of.

- [x] Yes
- [ ] No

This removes the checkbox icon and checkbox icon class inputs. Typically every checkbox in an app should have the same icon, so customization should be done within the helm component rather than having to do it on every checkbox in an app. I can revert this is we feel its really beneficial, but no other ui libraries seem to offer this so it seems unlikely to have widespread usage.


libs/ui/checkbox/helm/src/lib/hlm-checkbox.component.ts
```diff
@@ -27,9 +30,9 @@ export const HLM_CHECKBOX_VALUE_ACCESSOR = {
 			[aria-labelledby]="ariaLabelledby()"
 			[aria-describedby]="ariaDescribedby()"
 			(changed)="_handleChange()"
-			(touched)="_onTouched()"
+			(touched)="_onTouched?.()"
 		>
-			<hlm-checkbox-checkicon [class]="checkIconClass()" [iconName]="checkIconName()" />
+			<ng-icon hlm size="sm" name="lucideCheck" />
```
elite-benni, line 35: The name should come from the input, some users want to use another icon. 
I see that you removed the input. I would say this is a breaking change.
Would you say, if someone wants another icon he should change the hlm source in his repo?

ashley-hunter: Yes, so I guess the question is, where does it make sense for this kind of customization? Is there really a common case for checkboxes in different parts of an app to have a different check icons? 
In most cases that I have seen e.g. material, shadcn etc.. the icon is always the same for every checkbox. 

I can see a case where a user may want a different icon for every checkbox, in which case customizing the helm component makes perfect sense. But does it make sense to also change it on a per instance basis? 

If so I can certainly add it back in, happy to hear thoughts!

elite-benni: I cannot remember why the inputs were there in the first place.
I just wanted to point out that some users could maybe used this feature. 
Just an example that pops into my head. Someone used checkbox in a tic tac toe game. Maybe not the best example. 😉 But we don't know what users are coming up with when using the library. 

In an normal app i totally agree with you, that checkboxes should always look the same, and maybe these special cases should not be handled with a checkbox. 

Both ways have their valid points and I have a hard time to decide which one is better. 

Just another question. Who provides the icon now, that was previously done by the hlmcheckicon component, which also was a little bit weird if someone used his own icon.  
So maybe the cleaner end result is to remove it and document it as breaking change.

ashley-hunter: The helm checkbox icon component was only used internally so never something that would have been used by a consumer.
The checkbox component now provides the icon directly.
But yes in the case where someone was to provide a custom icon they would have to provide it manually themselves.

I'm happy to restore it if we think it's worth it, just not something any other library I see provides out of the box and most libraries don't copy the components into the project, we do, so that kind of infrequent thing could be added in the project if someone needed it, rather than us needing to provide it by default for everyone, but I can easily restore it if we think it's best.

elite-benni: Oh i missed that!

It is fine for me, if someone needs the inputs he still has the possibility to do this in his own helm component. 

I believe it is the cleaner way, but it has the possibility to be a breaking change. 

Nice work

ashley-hunter: Grand, I have updated the PR description to mention this. If anyone else has an opinion let me know, happy to revert this if needed. Thanks for you feedback!

libs/ui/checkbox/helm/src/lib/hlm-checkbox.component.ts
```diff
@@ -1,9 +1,12 @@
 import { Component, booleanAttribute, computed, forwardRef, input, model, output, signal } from '@angular/core';
 import { NG_VALUE_ACCESSOR } from '@angular/forms';
+import { NgIcon, provideIcons } from '@ng-icons/core';
+import { lucideCheck } from '@ng-icons/lucide';
 import { BrnCheckboxComponent } from '@spartan-ng/brain/checkbox';
 import { hlm } from '@spartan-ng/brain/core';
+import { ChangeFn, TouchFn } from '@spartan-ng/brain/forms';
```
ajitzero, line 7: super minor nit:
```suggestion
import type { ChangeFn, TouchFn } from '@spartan-ng/brain/forms';
```

libs/ui/icon/helm/src/lib/hlm-icon.directive.ts
```diff
@@ -1,12 +1,7 @@
 import { Directive, computed, input } from '@angular/core';
 import { injectHlmIconConfig } from './hlm-icon.token';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const DEFINED_SIZES = ['xs', 'sm', 'base', 'lg', 'xl', 'none'] as const;
-
-type DefinedSizes = (typeof DEFINED_SIZES)[number];
-
-export type IconSize = DefinedSizes | (Record<never, never> & string);
+export type IconSize = 'xs' | 'sm' | 'base' | 'lg' | 'xl' | 'none' | ({} & string);
```
ajitzero, line 4: I can't find a clear/objective reason for this, but using `Record<never, never>` is recommended over `{}`.

- This is the most "complete" request I could find: https://github.com/dotansimha/graphql-code-generator/issues/6068
- And this is also from the above issue: https://github.com/microsoft/TypeScript/wiki/FAQ#why-are-all-types-assignable-to-empty-interfaces

```suggestion
export type IconSize = 'xs' | 'sm' | 'base' | 'lg' | 'xl' | 'none' | (Record<never, never> & string);
```

ajitzero (approved):
This was much needed! Thanks for this!

I left few a very minor comments:

libs/ui/icon/helm/src/lib/hlm-icon.directive.ts
ashley-hunter: Yes, typically that is absolutely the case if you are using it as a type, in this case it doesn't change the type, it's a bit of a hack to get the typescript autocomplete to work nicer.

If we do something like 'sm' | 'xs' | string, typescript collapses the union down to just string and we get no autocomplete as you can provide suggestions for type string. This little addition doesn't affect the type, but does stop the union from collapsing giving us autocomplete suggestions 😊

libs/ui/checkbox/helm/src/lib/hlm-checkbox.component.ts
ashley-hunter: Thanks! Updated!

libs/ui/icon/helm/src/lib/hlm-icon.directive.ts
ajitzero: Yeah, no, I get that part. I meant that using `Record<never, never>` works in the same way as `{}` to enable the autocomplete, but it is generally more type-safe. Having said that, I can't find anything to cite this 😅

From what I can find, the type-safety issue arises only if we use them standalone, not in union. In this case, there appears to be no difference in the expected output, so this is completely fine.

ashley-hunter: Ah apologies, I see what you mean. Correct, yes, we & the empty object type with a string which avoids the type issues normally associated with {} if used standalone 😊

PR #586, opened 2025-02-06
feat: adding support for tailwind 4

ashley-hunter:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

N/A

- [ ] Yes
- [X] No

## Other information

Adding support for Tailwind 4. This adds tailwind version detection to the theme generator and adds the preset import automatically if Tailwind 4 is detected.

### DIFF
--- a/apps/app/src/app/pages/(documentation)/documentation/(ui)/installation.page.ts
+++ b/apps/app/src/app/pages/(documentation)/documentation/(ui)/installation.page.ts
@@ -62,15 +62,14 @@ export const routeMeta: RouteMeta = {
 				also builds on top of angular/cdk. To install it run the following command:
 			</p>
 			<spartan-code class="mt-4" code="npm i @angular/cdk" />
-			<spartan-section-sub-heading id="setting-up-tailwind">Setting up tailwind.config.js</spartan-section-sub-heading>
+			<spartan-section-sub-heading id="setting-up-tailwind">Setup Tailwind</spartan-section-sub-heading>
 			<p class="${hlmP}">
-				You now have to add our spartan-specific configuration to your TailwindCSS setup. To make the setup of your
-				<code class="${hlmCode}">tailwind.config.js</code>
-				as easy as possible, the
+				You now have to add our spartan-specific configuration to your TailwindCSS setup. To make the setup as easy as
+				possible, the
 				<code class="${hlmCode}">&#64;spartan-ng/brain</code>
 				package comes with it own preset.
 			</p>
-			<p class="${hlmP}">Simply add it to the presets array of your config file:</p>
+			<p class="${hlmP}">If you are using Tailwind 3 add the following to your config file:</p>
 			<spartan-cli-tabs
 				language="js"
 				class="mb-6 mt-4"
@@ -106,6 +105,10 @@ module.exports = {
 };
 "
 			/>
+
+			<p class="${hlmP}">If you are using Tailwind 4 add the following import to your global stylesheet:</p>
+
+			<spartan-code class="mt-4" code="@import '@spartan-ng/brain/hlm-tailwind-preset.css';" />
 
 			<spartan-section-sub-heading id="adding-css-vars">Adding CSS variables</spartan-section-sub-heading>
 			<p class="${hlmP}">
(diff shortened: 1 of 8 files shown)


apps/app/src/app/pages/(documentation)/documentation/(ui)/installation.page.ts
```diff
@@ -107,6 +106,10 @@ module.exports = {
 "
 			/>
 
+			<p class="${hlmP}">If you are using Tailwind 4 add the following import to your global stylesheet:</p>
+
+			<spartan-code class="mt-4" code="@import '@spartan-ng/brain/hlm-tailwind-preset.css';" />
```
ajitzero, line 111: That was fast! Such a clean solution :)

I was thinking we could generate this CSS instead - and only if it doesn't exist already - just like we do for our components. That way, if someone strips out hlm styling in favor of their custom styling, we don't need to download this style file along with the brain package.

We could totally look into this with a different PR, if at all.

This also came to my mind only because of [general sentiment/feedback](https://www.reddit.com/r/Angular2/comments/1ihldks/comment/may8m8v/) (frankly, it's just one comment) around users not likely needing to install the brain package either, so having to install it just for some CSS might not go down well. That is an entirely different debate, and I personally like having code centralized in one package. We could maybe even generator the entire brain package locally instead of installing, but that's irrelevant for this PR.

ajitzero (approved):

apps/app/src/app/pages/(documentation)/documentation/(ui)/installation.page.ts
ashley-hunter: Thanks, yes, so there are multiple approaches that could be take here, and I was deliberating between them so thoughts are welcome.

This approach works well for people who are potentially updating to tailwind 4, as the existing variables remain and the names do not change.

We could take the alternative approach which would be to add the Tailwind theme directly to the global stylesheet, the challenges here are:

- If we generate the @theme block in the stylesheet directly, great for new apps, but users who want to upgrade need to do more rework. I guess we could perhaps create a generator to migrate this, but that could be challenging as someone may have custom values, or multiple themes etc...
- Do we support both approaches? In that case our generators become more complex supporting multiple approaches based on a users setup.

Any thoughts are welcome, maybe its fine not to worry about the upgrade case 🤷‍♂️

> This also came to my mind only because of [general sentiment/feedback](https://www.reddit.com/r/Angular2/comments/1ihldks/comment/may8m8v/) (frankly, it's just one comment) around users not likely needing to install the brain package either, so having to install it just for some CSS might not go down well. That is an entirely different debate, and I personally like having code centralized in one package. We could maybe even generator the entire brain package locally instead of installing, but that's irrelevant for this PR.

This is interesting, im not sure how much value there is in spartan without the brain package? That being said, I think the whole setup procedure could be greatly simplified, perhaps just a single `ng add` command to set everything up rather than running multiple steps including some manual. This is something I might look into.

elite-benni: I think this PR supports the users well in terms of upgrading. 

I don't think that that hlm works without brain, we decided that helm without brain is not really valuable. 
This made some implementations a lot easier and we only heard about users that use brain without hlm but not the other way around. 

I would say, for the ``ng add`` command you mentioned we can think of adding the theme directly. 

If a existing user wants to add the styles globally he can still do this. ;)

PR #595, opened 2025-02-13
fix(command): ngModel is recognized, NG0950 error fixed, focus up and…   … down fixed

### LINKED ISSUE #593: Command: Angular 19 Up Down Arrow and small issues

Angular 19

command

### Up Down arrow problem
When no searchtext is set, arrow up and down does not work on angular 19. 
The effect immediately sets the active item back to the first item. 

### Typeahead
Additionally the withtypeahead behavior doesn't match shadcn and doesn't feel right. 

e.g. the selection sometimes does not work.  e.g. type slowly "s e t " then settings is not active. 
I would rather like the first item always to be active when the search is finished.
This is not really a Angular 19 only "problem". 

### ngModel
if a Value is bound to the ngModel of the search input, the items are not filtered.

### NG0950 Error
There is a way the NG0950 Error still occurs because of the valueChange PR.

- [x] Yes
- [ ] No

elite-benni:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [x] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

Typeahead does not feel right and doesn't match shadcn behaviour. 
NG0950 Error can occur still.
Up and down Arrow is not working when searchtext is empty
Setting searchtext via ngModel does not work

Closes #593 

- [ ] Yes
- [x] No

## Other information

Sorry for adding so much additional fixes to this PR.

### DIFF
--- a/libs/brain/command/src/lib/brn-command-search-input.directive.ts
+++ b/libs/brain/command/src/lib/brn-command-search-input.directive.ts
@@ -1,5 +1,6 @@
-import { Directive, ElementRef, HostListener, inject, signal } from '@angular/core';
+import { Directive, ElementRef, inject, signal } from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
+import { NgModel } from '@angular/forms';
 import { startWith } from 'rxjs/operators';
 import { provideBrnCommandSearchInput } from './brn-command-search-input.token';
 import { injectBrnCommand } from './brn-command.token';
@@ -12,9 +13,12 @@ import { injectBrnCommand } from './brn-command.token';
 		role: 'combobox',
 		'aria-autocomplete': 'list',
 		'[attr.aria-activedescendant]': '_activeDescendant()',
+		'(keydown)': 'onKeyDown($event)',
+		'(input)': 'onInput()',
 	},
 })
 export class BrnCommandSearchInputDirective {
+	private readonly _ngModel = inject(NgModel, { optional: true });
 	private readonly _command = injectBrnCommand();
 	private readonly _elementRef = inject<ElementRef<HTMLInputElement>>(ElementRef);
 
@@ -28,16 +32,24 @@ export class BrnCommandSearchInputDirective {
 		this._command.keyManager.change
 			.pipe(startWith(this._command.keyManager.activeItemIndex), takeUntilDestroyed())
 			.subscribe(() => this._activeDescendant.set(this._command.keyManager.activeItem?.id()));
+
+		const ngModel = this._ngModel;
+		if (ngModel) {
+			ngModel.valueChanges?.pipe(takeUntilDestroyed()).subscribe((value) => {
+				this.value.set(value);
+			});
+		}
 	}
 
 	/** Listen for changes to the input value */
-	@HostListener('input')
 	protected onInput(): void {
+		if (this._ngModel) {
+			return;
+		}
 		this.value.set(this._elementRef.nativeElement.value);
 	}
 
 	/** Listen for keydown events */
-	@HostListener('keydown', ['$event'])
 	protected onKeyDown(event: KeyboardEvent): void {
 		this._command.keyManager.onKeydown(event);
 	}
--- a/libs/brain/command/src/lib/brn-command-search-input.directive.ts
+++ b/libs/brain/command/src/lib/brn-command-search-input.directive.ts
@@ -1,6 +1,6 @@
-import { Directive, ElementRef, inject, signal } from '@angular/core';
+import { computed, Directive, effect, ElementRef, Inject, input, Optional, Renderer2, signal } from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
-import { NgModel } from '@angular/forms';
+import { COMPOSITION_BUFFER_MODE, DefaultValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
 import { startWith } from 'rxjs/operators';
 import { provideBrnCommandSearchInput } from './brn-command-search-input.token';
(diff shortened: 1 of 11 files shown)


goetzrobin (approved):
Looks good to me! I'll wait for @ashley-hunter to review before merging!

libs/brain/command/src/lib/brn-command-search-input.directive.ts
```diff
@@ -28,16 +32,24 @@ export class BrnCommandSearchInputDirective {
...
 			.pipe(startWith(this._command.keyManager.activeItemIndex), takeUntilDestroyed())
 			.subscribe(() => this._activeDescendant.set(this._command.keyManager.activeItem?.id()));
+
+		const ngModel = this._ngModel;
+		if (ngModel) {
+			ngModel.valueChanges?.pipe(takeUntilDestroyed()).subscribe((value) => {
+				this.value.set(value);
+			});
+		}
 	}
```
ashley-hunter, line 42: Is this something we could implement `ControlValueAccessor` to handle? Just curious, I haven't checked, but for example would this approach work if someone was using a reactive form instead to ngModel - in which case perhaps ControlValueAccessor might support both? I havent had a chance to check so maybe it is fine as is.

elite-benni: You are right and I didn't also want to use a ngModel in the first place, since the typed value is not really the value that i want to save into the form.
I hope this makes sense. 

I just think that many users use ngModel just to bind to the value to an input, even if they don't have a form. 

Is setting the initial searchvalue something i would create a ``FormControl`` for?

alternative approach, we handle the value by ourselves like this and also remove the ngModel. 
```
public readonly value = input<string>('');
public readonly mutableValue = computed(() => signal(this.value()));
public readonly valueState = computed(() => this.mutableValue()());

effect(() => {
  this._elementRef.nativeElement.value = this.valueState();
});

protected onInput(): void {
  this.mutableValue().set(this._elementRef.nativeElement.value);
}
```
What do you think?

ashley-hunter: Well, one thing you could do is just extend the `DefaultValueAccessor` that gets applied to all imports when FormsModule is imported, and hook into it:

```ts
@Directive({
   ...
  providers: [
    {
      provide: NG_VALUE_ACCESSOR,
      useExisting: BrnCommandSearchInputDirective,
      multi: true,
    },
  ],
})
export class BrnCommandSearchInputDirective extends DefaultValueAccessor {
  private readonly _element = inject<ElementRef<HTMLInputElement>>(ElementRef);

  /** @internal Expose the current text value */
  public readonly value = signal<string>('');

  /** The id of the active option */
  protected readonly _activeDescendant = signal<string | undefined>(undefined);

  override writeValue(value: string | null): void {
    super.writeValue(value);

    if (value) {
      this.value.set(value);
    }
  }

  /** Listen for changes to the input value */
  protected onInput(): void {
    this.value.set(this._element.nativeElement.value);
  }
}```

That solves the initial value issue and means you don't need to do the whole form control setup

https://stackblitz.com/edit/stackblitz-starters-azvgfsbp?description=An%20angular-cli%20project%20based%20on%20@angular/animations,%20@angular/common,%20@angular/compiler,%20@angular/core,%20@angular/forms,%20@angular/platform-browser,%20@angular/platform-browser-dynamic,%20@angular/router,%20core-js,%20rxjs,%20tslib%20and%20zone.js&file=src/main.ts&template=node&title=Angular%20Starter

elite-benni: I can add this Accessor. 
One downside is, that you need FormsModule to bind whereas on my Example this would not be the case. 

But to be honest, I guess if a user sees an input he would expect that ngModel binding is working. 
So i will add it like you said!

ashley-hunter: Ah, ok, so are you referring to the case where someone binds using [value] directly?

elite-benni: I guess we have multiple options now. 
Maybe @goetzrobin  and @thatsamsonkid could you also tell what you think? 
1) only implement binding to [ngModel] 
2) implement binding to [value] and [ngModel]
3) binding to [ngModel] and formControl
4) implement all cases, binding to [value] and defaultAccessor

From our discussion here, my vote would be number 2, but I am open to whatever you think its the best. ;)

elite-benni: just added all cases. 
Only binding to ngModel like i did, seemed a bit strange while implementing. ;)

ashley-hunter (approved):
Looks good, thanks

AlonMiz (approved):
great

PR #610, opened 2025-02-19
feat(dialog): add more dialog options to BrnDialogDefaultOptions

### LINKED ISSUE #608: Dialog: add provideBrnDialogConfig to override input values

dialog

### Information

I am implementing an `AutocompleteTriggerDirective` and the inputs could be set in the constructor before #599.

```ts
@Directive({
  selector: '[psAutocompleteTrigger]',
  standalone: true,
  host: {
    '(click)': 'open()',
  },
})
export class AutocompleteTriggerDirective {
  private readonly _host = inject(ElementRef, { host: true });

  private _brnDialog = inject(BrnDialogComponent, { optional: true });

  public readonly disabled = input<boolean, BooleanInput>(false, {
    transform: booleanAttribute,
  });

  constructor() {
    if (!this._brnDialog) return;

    this._brnDialog.attachTo = this._host.nativeElement;
    // otherwise the input field will loose focus
    this._brnDialog.autoFocus = 'first-heading';
    this._brnDialog.closeOnOutsidePointerEvents = true;
  }

  open() {
    if (this.disabled()) return;

    this._brnDialog?.open();
  }
}
```

Some inputs have a mutable signal like `attachTo` and `closeOnOutsidePointerEvents`, but `autoFocus` is just an signal input. I can update the inputs to use the new `mutable...` signal states, but `autoFocus` is not included.

```diff
- this._brnDialog.attachTo = this._host.nativeElement;
+ this._brnDialog.mutableAttachTo().set(this._host.nativeElement);
- this._brnDialog.autoFocus = 'first-heading'; // no option to change the input
- this._brnDialog.closeOnOutsidePointerEvents = true;
+ this._brnDialog.mutableCloseOnOutsidePointerEvents().set(true);
```

A `provideBrnDialogConfig` would allow to update config values like `closeOnOutsidePointerEvents`, `autoFocus` and more programmatically.

_No response_

- [x] Yes
- [ ] No

marcjulian:
* change closeDelay to 100 default
* use closeDelay input value
* use defaultOptions values for input signal defaults

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [x] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

Closes #608

- [x] Yes
- [ ] No

`provideBrnDialogDefaultOptions` needs to be added to `HlmAlertDialogComponent`

```ts
import {
	BRN_ALERT_DIALOG_DEFAULT_OPTIONS,
} from '@spartan-ng/brain/alert-dialog';
import { provideBrnDialogDefaultOptions } from '@spartan-ng/brain/dialog';

providers: [
		...,
		provideBrnDialogDefaultOptions({
			...BRN_ALERT_DIALOG_DEFAULT_OPTIONS,
		}),
	],
```

### DIFF
--- a/libs/brain/alert-dialog/src/lib/brn-alert-dialog.component.ts
+++ b/libs/brain/alert-dialog/src/lib/brn-alert-dialog.component.ts
@@ -1,5 +1,11 @@
 import { ChangeDetectionStrategy, Component, forwardRef, ViewEncapsulation } from '@angular/core';
-import { BrnDialogComponent } from '@spartan-ng/brain/dialog';
+import { BrnDialogComponent, BrnDialogDefaultOptions, provideBrnDialogDefaultOptions } from '@spartan-ng/brain/dialog';
+
+export const BRN_ALERT_DIALOG_DEFAULT_OPTIONS: Partial<BrnDialogDefaultOptions> = {
+	closeOnBackdropClick: false,
+	closeOnOutsidePointerEvents: false,
+	role: 'alertdialog',
+};
 
 @Component({
 	selector: 'brn-alert-dialog',
@@ -12,16 +18,10 @@ import { BrnDialogComponent } from '@spartan-ng/brain/dialog';
 			provide: BrnDialogComponent,
 			useExisting: forwardRef(() => BrnAlertDialogComponent),
 		},
+		provideBrnDialogDefaultOptions(BRN_ALERT_DIALOG_DEFAULT_OPTIONS),
 	],
 	changeDetection: ChangeDetectionStrategy.OnPush,
 	encapsulation: ViewEncapsulation.None,
 	exportAs: 'brnAlertDialog',
 })
-export class BrnAlertDialogComponent extends BrnDialogComponent {
-	constructor() {
-		super();
-		this.mutableCloseOnOutsidePointerEvents().set(false);
-		this.mutableCloseOnBackdropClick().set(false);
-		this.mutableRole().set('alertdialog');
-	}
-}
+export class BrnAlertDialogComponent extends BrnDialogComponent {}
--- a/apps/app/src/app/pages/(components)/components/(card)/card.preview.ts
+++ b/apps/app/src/app/pages/(components)/components/(card)/card.preview.ts
@@ -58,7 +58,7 @@ type Framework = { label: string; value: string };
 				<label class="my-4 mb-1.5 block" hlmLabel>
 					Framework
 
-					<brn-popover [state]="state()" (stateChanged)="stateChanged($event)" sideOffset="5" closeDelay="100">
+					<brn-popover [state]="state()" (stateChanged)="stateChanged($event)" sideOffset="5">
 						<button
 							class="mt-1.5 w-full justify-between"
 							id="edit-profile"
@@ -206,7 +206,7 @@ type Framework = { label: string; value: string };
               <label class="my-4 mb-1.5 block" hlmLabel>
                   Framework
 
-                  <brn-popover [state]="state()" (stateChanged)="stateChanged($event)" sideOffset="5" closeDelay="100">
+                  <brn-popover [state]="state()" (stateChanged)="stateChanged($event)" sideOffset="5">
                       <button
                               class="w-full mt-1.5 justify-between"
                               id="edit-profile"
--- a/apps/app/src/app/pages/(components)/components/(combobox)/combobox.preview.ts
+++ b/apps/app/src/app/pages/(components)/components/(combobox)/combobox.preview.ts
@@ -1,4 +1,3 @@
-import { NgForOf } from '@angular/common';
 import { Component, signal } from '@angular/core';
 import { NgIcon, provideIcons } from '@ng-icons/core';
 import { lucideCheck, lucideChevronsUpDown, lucideSearch } from '@ng-icons/lucide';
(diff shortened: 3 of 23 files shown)


elite-benni:
Maybe i have to check it again. But the reason why i didn't add more peoperties to the default options is because is felt misleading.

You would have to add it to the helm component as well, otherwise it won't work. 
Atleast thats what I experienced on my tests. 

I am a big fan of the inject defaultOptions pattern but it doesnt work that good for components.

Maybe I am wrong and had some other mistakes in my tests, but thats why i refactored back to the mutableSignals instead of the default options.

marcjulian:
I thought this is common practices to provide default input, where the component inputs cannot be used. 
We have `provideBrnButtonConfig`, `provideHlmDatePickerConfig` etc.

Not sure why it didn't work before, but looks like that the only option `closeDelay` wasn't even used from `defaultOptions` nor from the closeDelay input.

For me it seems to work, try it out with hm-alert-dialog and add state open to `provideBrnDialogDefaultOptions` and the alert dialog should be open when navigating to the alert-dialog page. 
However, for this case setting the state via `provideBrnDialogDefaultOptions` causes issues for the outputs on initial page load.

I guess we will still need the mutableSignals for inputs which may change for other inputs like in the trigger directives.

elite-benni:
I just nearly went crazy because it really works. 
My problem was, that I wanted to add the provideconfig to ``BrnAlertDialogComponent`` but it is not reused automatically for ``HlmAlertDialogComponent``. 
This means that you have to also Change the HlmComponent to make this feature work again, which is maybe a bigger breaking change. 

But maybe it was seen too serious by me.

As written in my first comment, i like the defaultOptions pattern a lot and would love this to be the default way to set the options. 

You are right, not all of them can be removed!
I am pretty sure we can remove some of the mutableInputs, because they are just used by helm component to set it. 

The following should be able to remove:

``mutableCloseOnBackdropClick``
``mutableRole``
``_mutableHasBackdrop``
``mutableScrollStrategy``

libs/brain/alert-dialog/src/lib/brn-alert-dialog.component.ts
```diff
@@ -1,5 +1,11 @@
 import { ChangeDetectionStrategy, Component, forwardRef, ViewEncapsulation } from '@angular/core';
-import { BrnDialogComponent } from '@spartan-ng/brain/dialog';
+import { BrnDialogComponent, BrnDialogDefaultOptions, provideBrnDialogDefaultOptions } from '@spartan-ng/brain/dialog';
+
+export const BRN_ALERT_DIALOG_DEFAULT_OPTIONS: Partial<BrnDialogDefaultOptions> = {
+	closeOnBackdropClick: false,
+	closeOnOutsidePointerEvents: false,
+	role: 'alertdialog',
+};
```
marcjulian, line 8: I expose the defaults for alert dialog here and use it in brn-alert-dialog as well as in hlm-alert-dialog. Maybe that solves it a little bit that it has to be used in hlm-alert-dialog as well

elite-benni: yeah thanks! thats a good idea!

libs/brain/alert-dialog/src/lib/brn-alert-dialog.component.ts
```diff
@@ -12,16 +18,10 @@ import { BrnDialogComponent } from '@spartan-ng/brain/dialog';
...
 })
-export class BrnAlertDialogComponent extends BrnDialogComponent {
-	constructor() {
-		super();
-		this.mutableCloseOnOutsidePointerEvents().set(false);
-		this.mutableCloseOnBackdropClick().set(false);
-		this.mutableRole().set('alertdialog');
-	}
-}
+export class BrnAlertDialogComponent extends BrnDialogComponent {}
```
elite-benni, line 27: I love it when all the imperative set calls are gone ;).

elite-benni:
Thanks for this improvement. 

LGTM!!!

I just wanted to point out, that it is a potential breaking change for users that use their own helm components, because they have to add the ``provideBrnDialogDefaultOptions``, but I think it is ok.
We had some of these "breaking changes" while refactoring the inputs.

elite-benni (approved):

libs/brain/alert-dialog/src/lib/brn-alert-dialog.component.ts
marcjulian: Yes, its so clean!

marcjulian:
I updated the breaking change section

PR #624, opened 2025-03-07
fix(select): use toggleSelect to prevent duplicate values

### LINKED ISSUE #645: Select: Multi Select - Cannot deselect a option and can reselect same option multiple times

Behavior can be replicated in the documentation example of the multi-select.

https://www.spartan.ng/components/select

select

### Description

When using the multi-select you cannot deselect an option that is already selected. You can also click the same option multiple times and it will add it as selected.

Steps to reproduce;

1. Use a multi-select
2. Select a value
3. Click the same value to deselect it
4. Actual Outcome: Value is added again as selected rather than deselected.
5. Expected Outcome: Value should be deselected.

As for the same value being selected multiple times, a check should be added to check if an item is already selected before adding it as selected. This case won't be as obvious when the select/deselect issue is fixed, but it would be possible when interacting with the form control programmatically in the real world.

```true

```

### Other information

_No response_

- [ ] Yes
- [x] No

yackinn:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [x] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

**Multilple=false:**
If you click an already checked option nothing happens except for closing the dropdown. The option value will still be the selected value.

**Multiple=true:**
If you click an already checked option it will add the same value again to the select values again and again.

## What is the new behavior?
**Multilple=false:**
If you click an already checked option it will uncheck the option and close the dropdown.

**Multiple=true:**
If you click an already checked option it will remove the value from the selection and it will keep the dropdown opened.

- [ ] Yes
- [x] No

## Other information
- I don't think a documentation update or test update is required, but let me know if you think otherwise.
- The calendar e2e tests are failing

Closes #645

### DIFF
--- a/libs/brain/select/src/lib/brn-select-option.directive.ts
+++ b/libs/brain/select/src/lib/brn-select-option.directive.ts
@@ -48,7 +48,7 @@ export class BrnSelectOptionDirective<T> implements Highlightable {
 			return;
 		}
 
-		this._select.selectOption(this.value());
+		this._select.toggleSelect(this.value());
 	}
 
 	/** Get the label for this element which is required by the FocusableOption interface. */


ashley-hunter:
Thank you! You are absolutely correct there is an issue in multi select, the value should not get repeated, toggling makes more sense.

I don't think however that single select should toggle on selection, if we take a look at [ShadCN](https://ui.shadcn.com/docs/components/select) or [Angular Material](https://material.angular.io/components/select/examples) or the native html select, none of them toggle on single select, so I think this update should only affect multiple selects.

A test case to cover the multi select toggle behaviour would be nice so we have it covered. The bug slipped through the tests before because none of the existing tests covered it.

Any questions or issues just reach out, thanks!

goetzrobin:
@yackinn please let us know if you need any assistance for this

yackinn:
@goetzrobin Will do once I find some time to get into it. Thank you.

ashley-hunter:
Thank you. I'm going to close this PR in favour of #646 which fixes the same issue.

PR #639, opened 2025-03-21
feat: add input-otp

marcjulian:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [x] input-otp
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] tooltip
- [ ] typography

## What is the new behavior?

CleanShot 2025-03-21 at 10 35 57

CleanShot 2025-03-21 at 10 36 12

- [ ] Yes
- [ ] No

## Other information

Ring colors differ 

shadcn/ui `--ring: 240 5% 64.9%;`
spartan/ui `--ring: 240 5.9% 10%;`


marcjulian:
Not sure how to fix the lint issue `Parsing error: /Users/m/Dev/forkes/spartan/libs/ui/input-otp/helm/src/test-setup.ts was not found by the project service. Consider either including it in the tsconfig.json or including it in allowDefaultProject`

libs/brain/input-otp/src/lib/brn-input-otp.component.ts
```diff
@@ -0,0 +1,127 @@
...
+		if (newValue.length > maxLength) {
+			// Replace the last character when max length is exceeded
+			newValue = newValue.slice(0, maxLength - 1) + newValue.slice(-1);
+		}
+
+		this.value.set(newValue);
+		this._onChange?.(newValue);
+	}
+
+	/** CONROL VALUE ACCESSOR */
```
ajitzero, line 108: minor: typo
```suggestion
	/** CONTROL VALUE ACCESSOR */
```

libs/brain/input-otp/src/lib/brn-input-otp.component.ts
```diff
@@ -0,0 +1,127 @@
...
+	standalone: true,
+	imports: [FormsModule],
+	template: `
+		<ng-content />
+		<div style="position: absolute; inset: 0; pointer-events: none;">
+			<input
+				[class]="inputClass()"
+				autocomplete="one-time-code"
+				data-slot="input-otp"
+				style="position: absolute; inset: 0; width: 100%; height: 100%; display: flex; textAlign: left; opacity: 1; color: transparent; pointerEvents: all; background: transparent; caret-color: transparent; border: 0px solid transparent; outline: transparent solid 0px; box-shadow: none; line-height: 1; letter-spacing: -0.5em; font-family: monospace; font-variant-numeric: tabular-nums;"
```
ajitzero, line 35: Do we need this styling at the brn level? I feel we could move this to hlm and not lose out on much, and we could recommend using the above `inputClass` input for this. Similarly, we could also expose an input for classes for the container div on Line 30 & host on Line 46.

What do you think?

libs/brain/input-otp/src/lib/brn-input-otp.component.ts
```diff
@@ -0,0 +1,127 @@
...
+import { ChangeFn, TouchFn } from '@spartan-ng/brain/forms';
+import { provideBrnInputOtp } from './brn-input-otp.token';
+
+export const BRN_INPUT_OTP_VALUE_ACCESSOR = {
+	provide: NG_VALUE_ACCESSOR,
+	useExisting: forwardRef(() => BrnInputOtpComponent),
+	multi: true,
+};
+
+export type InputMode = 'text' | 'tel' | 'url' | 'email' | 'numeric' | 'decimal' | 'search';
```
ajitzero, line 22: In case we need to support more (seeing form TODO comment on Line 61-62):
```suggestion
export type InputMode = 'text' | 'tel' | 'url' | 'email' | 'numeric' | 'decimal' | 'search' | (string & Record<never, never>);
```
This TypeScript trick would let users pass arbitrary strings in _addition_ to these strings.

apps/app/src/app/pages/(components)/components/(label)/label.page.ts
```diff
@@ -59,7 +59,7 @@ export const routeMeta: RouteMeta = {
 
 			<spartan-page-bottom-nav>
 				<spartan-page-bottom-nav-link href="menubar" label="Menubar" />
-				<spartan-page-bottom-nav-link direction="previous" href="input" label="Input" />
+				<spartan-page-bottom-nav-link direction="previous" href="input-otp" label="Input OTP" />
```
ajitzero, line 62: I think we need to also update the "Next" button in the input.page.ts

libs/brain/input-otp/src/lib/brn-input-otp.component.ts
```diff
@@ -0,0 +1,127 @@
...
+				[class]="inputClass()"
+				autocomplete="one-time-code"
+				data-slot="input-otp"
+				style="position: absolute; inset: 0; width: 100%; height: 100%; display: flex; textAlign: left; opacity: 1; color: transparent; pointerEvents: all; background: transparent; caret-color: transparent; border: 0px solid transparent; outline: transparent solid 0px; box-shadow: none; line-height: 1; letter-spacing: -0.5em; font-family: monospace; font-variant-numeric: tabular-nums;"
+				[disabled]="state().disabled()"
+				[inputMode]="inputMode()"
+				[ngModel]="value()"
+				(input)="onInputChange($event)"
+				(focus)="focused.set(true)"
+				(blur)="focused.set(false)"
```
ajitzero, line 41: We should also add a `(paste)` handler in case users copy the OTP from a 2FA app, email, or wherever. We can take this up in a separate PR though, or if I missed it somewhere in this PR then please ignore this.

libs/brain/input-otp/src/lib/brn-input-otp.component.ts
marcjulian: I think this is similar to the brn-radio styles on the input. They belong to the input, thus it is invisible but focusable in the ui. Input classes should be used for additional styles like for disabled state.

https://github.com/spartan-ng/spartan/blob/a3cd75d62a4b1729a588d5d373ce11a35420ae22/libs/brain/radio-group/src/lib/brn-radio.component.ts#L59

libs/brain/input-otp/src/lib/brn-input-otp.component.ts
marcjulian: I forgot to remove the TODO when i only had `numeric`. The types should be complete (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode), i only left out `none`. Thus are the same types supported by https://github.com/guilhermerodz/input-otp?tab=readme-ov-file#props

apps/app/src/app/pages/(components)/components/(label)/label.page.ts
marcjulian: Yes its updated on the input too

libs/brain/input-otp/src/lib/brn-input-otp.component.ts
marcjulian: Copy and paste already works, but we can add `(paste)` event and handle it a bit more.
What would be the use case for `(paste)`? Like removing whitespaces or keeping only the last part of a long text matching the maxLength?

marcjulian:
@ajitzero thanks for your feedback

libs/brain/input-otp/src/lib/brn-input-otp.component.ts
ajitzero: Makes sense!

apps/app/src/app/pages/(components)/components/(input-otp)/input-otp--form.example.ts
```diff
@@ -0,0 +1,221 @@
...
+	private resetCountdown() {
+		this.countdown.set(60);
+		this.startCountdown();
+	}
+
+	private startCountdown() {
+		this.stopCountdown();
+		this._intervalId = setInterval(() => {
+			if (this.countdown() > 0) {
+				this.countdown.set(this.countdown() - 1);
```
rifayetuxbd, line 96: Would it be possible to use the signal’s update method here? For example, you could try:
`this.countdown.update(currentCount => currentCount - 1)`

This approach might simplify the logic. What do you think?

marcjulian: Good idea

how about

```ts
private startCountdown() {
		this.stopCountdown();
		this._intervalId = setInterval(() => {
			this.countdown.update((countdown) => Math.max(0, countdown - 1));
			if (this.countdown() === 0) {
				console.log('Countdown finished');
				this.stopCountdown();
			}
		}, 1000);
	}
```

libs/brain/input-otp/src/lib/brn-input-otp.component.ts
marcjulian: Good idea. I added `transformPaste` input to allow the for custom transformation functions

`transformPaste: (pastedText: string, maxLength: number) => pastedText.replaceAll('-', '').slice(0, maxLength)`

apps/app/src/app/pages/(components)/components/(input-otp)/input-otp--form.example.ts
rifayetuxbd: Looks good. 

Update may not immediately propagated to next if call. 

What about:

```ts
private startCountdown() {
		this.stopCountdown();
		this._intervalId = setInterval(() => {
			this.countdown.update((countdown) => {
                            if(countdown === 0) {
                                this.stopCountdown();
                            }
                            return Math.max(0, countdown - 1)
                        });
		}, 1000);
}
```

marcjulian: Signal updates are synchronous and reads are immediate. 
The countdown signal will be updated (`this.countdown.update(...)`) synchronous and the read (`this.countdown()`) returns the recent value. 

I think using `this.countdown.update` to do more than updating the signal introduces unintended side effects.

```ts
this.countdown.update((countdown) => {
 if(countdown === 0) {
    this.stopCountdown(); // this is a side effect, not the concern of the update() function
 }
 return Math.max(0, countdown - 1)
});
```

goetzrobin:
@marcjulian Thanks so much for this! Looks great! is this ready to go? I see a lint error about a version dependency conflict with the brain package!

marcjulian:
@goetzrobin fixing the version conflict. Yes this is ready to go.

goetzrobin:
Looks great. A couple of things. Should we make the styles overridable like so?
```ts
@Component({
	selector: 'brn-input-otp',
	standalone: true,
	imports: [FormsModule],
	template: `
		<ng-content />
		<div [style]="containerStyles()">
			<input
				[class]="inputClass()"
				autocomplete="one-time-code"
				data-slot="input-otp"
				[style]="inputStyles()"
				[disabled]="state().disabled()"
				[inputMode]="inputMode()"
				[ngModel]="value()"
				(input)="onInputChange($event)"
				(paste)="onPaste($event)"
				(focus)="focused.set(true)"
				(blur)="focused.set(false)"
			/>
		</div>
	`,
	host: {
		'[style]': 'hostStyles()',
		'data-input-otp-container': 'true',
	},
	providers: [BRN_INPUT_OTP_VALUE_ACCESSOR, provideBrnInputOtp(BrnInputOtpComponent)],
})
```

and then in the component:

```ts
	public readonly hostStyles = input<string>(
		'position: relative; cursor: text; user-select: none; pointer-events: none;',
	);

	public readonly inputStyles = input<string>(
		'position: absolute; inset: 0; width: 100%; height: 100%; display: flex; textAlign: left; opacity: 1; color: transparent; pointerEvents: all; background: transparent; caret-color: transparent; border: 0px solid transparent; outline: transparent solid 0px; box-shadow: none; line-height: 1; letter-spacing: -0.5em; font-family: monospace; font-variant-numeric: tabular-nums;',
	);

	public readonly containerStyles = input<string>('position: absolute; inset: 0; pointer-events: none;');
```

apps/app/src/app/pages/(components)/components/(input-otp)/input-otp--form.example.ts
```diff
@@ -0,0 +1,239 @@
...
+
+	/** Overrides global formatDate  */
+	public transformPaste = (pastedText: string) => pastedText.replaceAll('-', '');
+
+	public form = this._formBuilder.group({
+		otp: [null, [Validators.required, Validators.minLength(this.maxLength), Validators.maxLength(this.maxLength)]],
+	});
+
+	constructor() {
+		this.startCountdown();
```
goetzrobin, line 78:
```suggestion
		afterNextRender(() => this.startCountdown());
```

apps/app/src/app/pages/(components)/components/(input-otp)/input-otp--form.example.ts
```diff
@@ -0,0 +1,239 @@
+import { Component, computed, inject, OnDestroy, signal } from '@angular/core';
```
goetzrobin, line 1:
```suggestion
import { afterNextRender, Component, computed, inject, OnDestroy, signal } from '@angular/core';
```

apps/app/src/app/pages/(components)/components/(input-otp)/input-otp--form.example.ts
```diff
@@ -0,0 +1,239 @@
...
+	private stopCountdown() {
+		if (this._intervalId) {
+			clearInterval(this._intervalId);
+			this._intervalId = undefined;
+		}
+	}
+}
+
+export const inputOtpFormCode = `
+import { Component, computed, inject, OnDestroy, signal } from '@angular/core';
```
goetzrobin, line 121:
```suggestion
import { afterNextRender, Component, computed, inject, OnDestroy, signal } from '@angular/core';
```

apps/app/src/app/pages/(components)/components/(input-otp)/input-otp--form.example.ts
```diff
@@ -0,0 +1,239 @@
...
+
+	/** Overrides global formatDate  */
+	public transformPaste = (pastedText: string) => pastedText.replaceAll('-', '');
+
+	public form = this._formBuilder.group({
+		otp: [null, [Validators.required, Validators.minLength(this.maxLength), Validators.maxLength(this.maxLength)]],
+	});
+
+	constructor() {
+		this.startCountdown();
```
goetzrobin, line 198:
```suggestion
		afterNextRender(() => this.startCountdown());
```

We need to put this into the afterNextRender hook. Or it will start the interval on the server during SSR and never return the page. Using afterNextRender we only start the interval in the browser

goetzrobin (changes requested):
Looks great. Couple minor comments and ideas! Curious to hear your thoughts!

marcjulian:
@goetzrobin yes we can make the styles overridable. I followed `brn-radio` where styles are also hardcoded to the input field. Should we consider this change for other components too?

https://github.com/spartan-ng/spartan/blob/a3cd75d62a4b1729a588d5d373ce11a35420ae22/libs/brain/radio-group/src/lib/brn-radio.component.ts#L59

apps/app/src/app/pages/(components)/components/(input-otp)/input-otp--form.example.ts
marcjulian: Good catch! i will apply thus changes, thank you!

marcjulian:
@goetzrobin all changes are committed 🤙

goetzrobin (approved):

PR #654, opened 2025-04-04
fix(tooltip): allow programmatic control of tooltipContentClasses

MatanShushan:
### ✅ PR Checklist  
- [x] The commit message follows our [guidelines](https://github.com/goetzrobin/spartan/blob/main/CONTRIBUTING.md#-commit-message-guidelines)  
- [ ] Tests for the changes have been added (not needed for this directive-level change)  
- [ ] Docs have been added / updated (not applicable)

---

### 🔧 PR Type  

- [x] Bugfix

---

### 📦 Which package are you modifying?

- [x] tooltip

---

### ❓ What is the current behavior?

When using `BrnTooltipTriggerDirective` as a `hostDirective`, it's not possible to programmatically set `tooltipContentClasses` since it is defined using `input()`, which is readonly in Angular's signal system.

---

### ✅ What is the new behavior?

This PR updates the `tooltipContentClasses` field in `BrnTooltipTriggerDirective` from an `input()` to a `model()`.

This allows:
- Two-way binding via the `Input`
- Programmatic updates using `.set()` — which is currently not possible with `input()`

---

### 💡 Example Use Case

```ts
@Directive({
  selector: '[myCustomTooltipTrigger]',
  standalone: true,
  hostDirectives: [{
    directive: BrnTooltipTriggerDirective,
    inputs: ['brnTooltipTrigger'],
  }]
})
export class MyCustomTooltipTriggerDirective {
  private readonly variant = input<'default' | 'dark'>('default');
  private readonly classes = computed(() => this.getClassesByVariant(this.variant()));
  private readonly _tooltipDirective = inject(BrnTooltipTriggerDirective, { self: true });

  constructor() {
    effect(() => {
      this._tooltipDirective.tooltipContentClasses.set(this.classes());
    });
  }
}

### DIFF
--- a/libs/brain/tooltip/src/lib/brn-tooltip-trigger.directive.ts
+++ b/libs/brain/tooltip/src/lib/brn-tooltip-trigger.directive.ts
@@ -271,7 +271,7 @@ export class BrnTooltipTriggerDirective implements OnDestroy, AfterViewInit {
 			const ariaDescribedBy = this.ariaDescribedBy()();
 			this._ariaDescriber.removeDescription(
 				this._elementRef.nativeElement,
-				untracked(() => this.ariaDescribedByPrevious()),
+				untracked(() => this.ariaDescribedByPrevious()()),
 				'tooltip',
 			);
 
--- a/libs/brain/tooltip/src/lib/brn-tooltip-trigger.directive.ts
+++ b/libs/brain/tooltip/src/lib/brn-tooltip-trigger.directive.ts
@@ -75,8 +75,8 @@ export const BRN_TOOLTIP_SCROLL_STRATEGY_FACTORY_PROVIDER: Provider = {
 	deps: [Overlay],
 	useFactory:
 		(overlay: Overlay): (() => ScrollStrategy) =>
-			() =>
-				overlay.scrollStrategies.reposition({ scrollThrottle: SCROLL_THROTTLE_MS }),
+		() =>
+			overlay.scrollStrategies.reposition({ scrollThrottle: SCROLL_THROTTLE_MS }),
 };
 
 const PANEL_CLASS = 'tooltip-panel';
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by goetzrobin** [comment]
I think we should keep classes as inputs to stay consistent with everywhere else in the Codebase. However, what we have done before is using a linkedSignal (or computed returning a signal) that acts as a mutable container for classes and provide a method to then programatically set that value. Kind of similar to this:

```ts
	public readonly brnHoverCardTriggerFor = input<TemplateRef<unknown> | BrnHoverCardContentDirective | undefined>(
		undefined,
	);
	public readonly mutableBrnHoverCardTriggerFor = computed(() => signal(this.brnHoverCardTriggerFor()));
	private readonly _brnHoverCardTriggerForState = computed(() => this.mutableBrnHoverCardTriggerFor()());

	constructor() {
		effect(() => {
			const value = this._brnHoverCardTriggerForState();
			untracked(() => {
				if (value) {
					this._contentService.setContent(value, this._vcr);
				}
			});
		});
	}
```

I am imagining a setTooltipContentClasses you can call. Also curious to hear your thoughts on this @ashley-hunter @elite-benni

ashley-hunter:
Yes, I agree, we should avoid the use of model here because that adds an output to the API, whereas outputs are typically for informing about changes due to an interaction within a component, whereas this is a case where we want to programmatically change a value from outside.

There is also a task underway to document component APIs automatically from the code, so this would also be documenting the output which isn't really what we want. 

The suggestions above sound good to me.

MatanShushan:
@ashley-hunter @goetzrobin 
Thank you for you response! 

@ashley-hunter 
Im not sure Im following, could you please provide an example? 
As i see it even if I'll expose a function that set the value programmatically I won't be able to set it due the fact input could not be seted

ashley-hunter:
Yes, you are absolutely correct, what what we can do is use linkedSignal to essentially convert the input into a mutable signal.

This is just a pseudocode example:

```ts
@Component({
   host: {
      'class': 'userClass()' 
   }
})
export class SomeComponent {
   readonly _userClass = input('', { alias: 'class' });

   protected readonly userClass = linkedSignal(() => this._userClass());

  setTooltipClass(class: string) {
    this.userClass.set(class);
  }
}
```

hopefully that should give you a little bit of an idea, apologies for any typos or syntax errors. I'm typing it on my phone which isn't all that easy 😂

if you have any issues, let me know and I'll provide some further guidance.

MatanShushan:
@ashley-hunter
Great idea and design!
Done, I would appreciate your review and let me know what you are think about it now :)

ashley-hunter (approved):
Thank you, the change looks good to me. The computed approach was a workaround before we dropped support for Angular 18, now that we support Angular 19+ we can use linkedSignal instead, but they both achieve the same thing, so it's not a big deal which approach we take. Thanks

goetzrobin:
@MatanShushan I think there is an error with ariaDescribedByPrevious and that's why the build is failing. can you check that?

MatanShushan:
@goetzrobin 
For sure.
On it, will update PR soon

PR #663, opened 2025-04-11
fix(checkbox): bring checkIconName back

dongphuong0905:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] carousel
- [x] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-otp
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

For now, we can't add custom checked icon. And the own icon example always show lucideCheckIcon

image

## What is the new behavior?

Bring the option `checkIconName` back, so we can override default icon

image

- [ ] Yes
- [x] No

### DIFF
--- a/libs/cli/src/generators/ui/libs/ui-checkbox-helm/files/lib/hlm-checkbox.component.ts.template
+++ b/libs/cli/src/generators/ui/libs/ui-checkbox-helm/files/lib/hlm-checkbox.component.ts.template
@@ -31,7 +31,7 @@ export const HLM_CHECKBOX_VALUE_ACCESSOR = {
 			(changed)="_handleChange()"
 			(touched)="_onTouched?.()"
 		>
-			<ng-icon [class]="_computedIconClass()" hlm size="sm" name="lucideCheck" />
+			<ng-icon hlm size="sm" [class]="_computedIconClass()" [name]="checkIconName()" />
 		</brn-checkbox>
 	`,
 	host: {
@@ -74,6 +74,9 @@ export class HlmCheckboxComponent {
 
 	/** The checked state of the checkbox. */
 	public readonly checked = model<CheckboxValue>(false);
+
+	/** Used to set the icon is displayed when checkbox is checked. Default is 'lucideCheck' */
+	public readonly checkIconName = input<string>('lucideCheck');
 
 	/** The name attribute of the checkbox. */
 	public readonly name = input<string | null>(null);
--- a/libs/ui/checkbox/helm/src/lib/hlm-checkbox.component.ts
+++ b/libs/ui/checkbox/helm/src/lib/hlm-checkbox.component.ts
@@ -31,7 +31,7 @@ export const HLM_CHECKBOX_VALUE_ACCESSOR = {
 			(changed)="_handleChange()"
 			(touched)="_onTouched?.()"
 		>
-			<ng-icon [class]="_computedIconClass()" hlm size="sm" name="lucideCheck" />
+			<ng-icon hlm size="sm" [class]="_computedIconClass()" [name]="checkIconName()" />
 		</brn-checkbox>
 	`,
 	host: {
@@ -74,6 +74,9 @@ export class HlmCheckboxComponent {
 
 	/** The checked state of the checkbox. */
 	public readonly checked = model<CheckboxValue>(false);
+
+	/** Used to set the icon is displayed when checkbox is checked. Default is 'lucideCheck' */
+	public readonly checkIconName = input<string>('lucideCheck');
 
 	/** The name attribute of the checkbox. */
 	public readonly name = input<string | null>(null);


ashley-hunter:
Hey, thanks for the PR! Just curious to understand the use case here. 

I can understand wanting to change the default icon to something other than lucideCheck, but typically you'd want all checkboxes in the application to have the same icon for their checked state. It's not all that common to have different checked icons in different checkboxes, nor is it particularly efficient to have to specify the desired icon on each instance of the checkbox.

We previously removed this input because this seemed to be an uncommon case, and you can modify the hlm-checkbox component in your codebase to change the default icon without needing to extend the component API.

Is there a use case for different checked icons in each checkbox? And is it likely to be something that would benefit a lot of people, or is this something that is fairly niche to your project, in which case adding the input can be done in your codebase?

Thanks!

dongphuong0905:
@ashley-hunter Thanks for your thoughtful response. I now see why this option was removed.

I agree with you that it’s a niche use case, as most of the time we don’t need multiple different icons for checkboxes. However, when I first saw the checkbox component, I really appreciated how easy it was to customize the icon. I even built a few showcase components using custom icons to explore its potential.

In my opinion, having more flexibility through options like this can make development smoother, especially for teams working on more dynamic or varied UIs. Since this addition doesn’t impact performance and keeps the default behavior unchanged, I think it’s a nice-to-have enhancement for those who need it.

ashley-hunter:
Thanks for your feedback ☺️ its good discuss the pros and cons of anything so we do what is best, so you input is valuable.

So, yes, adding an input to customize the check icon is fairly simple, but its good to look at it in the context of both how it will be used, what is easiest, and to ensure some level of consistency across our components.

I guess the following are the things that come to mind:

**Where best to make the change**

One benefit of Spartan is that we generate the files in your project so this gives you complete control to customize the components in any way you need, such as changing default icons or adding any custom inputs you see fit. These components are designed to be extended by you as you need, so we don't need to cater for every edge case out of the box, we provide the common cases.

**Usage**

It makes most sense for the default icon to be set within the hlm-checkbox component, otherwise the html is going to look like this, where we have to specify it on each checkbox, which increases the amount you have to type but also makes it easy to forget leading to different icons on different checkboxes inadvertantly.

```html
<label hlmLabel>
    <hlm-checkbox checkIconName="heroCheck" />
    Subscribe to newsletter
</label>

<label hlmLabel>
    <hlm-checkbox checkIconName="heroCheck" />
    Accept terms and conditions
</label>

<label hlmLabel>
    <!-- Whoops, forgot to add the input, now its a slightly different icon -->
    <hlm-checkbox />
    Receive spam phone calls
</label>
```

The additional issue is that simply specifying the name of the icon as an input is not enough, this by default will show no icon because you must also register the icon you want to use in the component provider too. This would likely be a source of confusion for people as to why the icon they have defined doesn't show. This would not be the case if the default icon was simply updated in the hlm-checkbox component.

**Consistency**

Every input we add is one more thing we have to support, maintain, test and document etc... every additional piece of code is potential to introduce bugs. In this case, sure, the amount of code is small and could be seen as trivial, but we have to consider how this change aligns with our other components. Many of our components contain icons, e.g. breadcrumbs, calendar, carousel, dialog, radio buttons, select etc... now that we have introduced this pattern of customizing icons via inputs, why not make all of these icons customizable via inputs in a similar way? If we add inputs for all these things there is a lot more code to add, test, document and maintain - if we only add it to checkboxes we now have one component that does things differently than the rest, and we will inevitably get requests to add the same functionality to other components.

**Alternatives**

One alternative pattern we do already use in places is having a `provideXyzConfig` function. So for checkbox this would be `provideCheckboxConfig`. You can use this to set defaults across your entire app, so hypothetically you could do something like this:

```ts
providers: [
   provideCheckboxConfig({ checkIcon: 'heroCheck' })
]
```

This allows customization without modifying the component, but means you don't have to specify it as an input on each instance of the checkbox.

That being said, Im not sure that this brings much benefit either over just modifying the hlm-component in the project, as this approach too still requires you to register the icon with `provideIcons` either at the root of your app or in every component that uses the checkbox any way.

Interested to hear your thoughts, and if anyone else has any I'd be happy to hear everyone's opinions on the matter 😊

@goetzrobin, @elite-benni, @ajitzero

goetzrobin:
I agree with @ashley-hunter here. I understand wanting it to be easy to customize the icon, but it is in reality not as easy as to just change the input and for things to just work, because you do need the icon registered. 

So I think this is something we do not provide by default, but because you own the code can add on a project basis. If you'd like you could add a section in the docs that has a "hlm-checkbox-custom" component where you explain how to add the input and why one still has to provide the icon and just changing the name isn't all that's needed for it to work?

dongphuong0905:
@goetzrobin @ashley-hunter  Yes, I think that’s a great idea!

I created this when upgrading to a newer version of Spartan and noticed that the custom checkbox in SimUI no longer worked as expected. So I’ll go ahead and update the SimUI documentation to include a hlm-checkbox-custom example, showing how to add the input and explaining the need to register the icon properly.

I’ll go ahead and close this PR. Thanks so much for the thoughtful feedback and discussion! <3

PR #669, opened 2025-04-19
fix(calendar): rendering weekdays should follow changes of rendering days

namdien177:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [x] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-otp
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

Weekdays don't render correctly if the rendering days are changed (due to various reasons, for example, `weekStartsOn`)

## What is the new behavior?

Weekdays will re-render if the rendering days change. The logic is similar to `BrnCalendarWeekDirective`.

- [ ] Yes
- [x] No

### DIFF
--- a/libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
+++ b/libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
@@ -37,7 +37,7 @@ export class BrnCalendarWeekdayDirective<T> implements OnDestroy {
 	protected readonly weekdays = computed(() => this._calendar.days().slice(0, 7));
 
 	/** Store the view refs */
-	private readonly _viewRefs: EmbeddedViewRef<BrnWeekdayContext>[] = [];
+	private _viewRefs: EmbeddedViewRef<BrnWeekdayContext>[] = [];
 
 	// Make sure the template checker knows the type of the context with which the
 	// template of this directive will be rendered
@@ -48,9 +48,6 @@ export class BrnCalendarWeekdayDirective<T> implements OnDestroy {
 	constructor() {
 		// Create a new view for each day
 		effect(() => {
-			// Clear the view container before creating new views
-			this._viewContainerRef.clear();
-
 			// Get the weekdays to display
 			const weekdays = this.weekdays();
 			// Render the weekdays
@@ -64,7 +61,7 @@ export class BrnCalendarWeekdayDirective<T> implements OnDestroy {
 			viewRef.destroy();
 		}
 
-		this._viewRefs.length = 0;
+		this._viewRefs = [];
 
 		// Create a new view for each day
 		for (const day of weekdays) {
--- a/libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
+++ b/libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
@@ -65,12 +65,9 @@ export class BrnCalendarWeekdayDirective<T> implements OnDestroy {
 
 		// Create a new view for each day
 		for (const day of weekdays) {
-			const viewRef = this._viewContainerRef.createEmbeddedView(
-				this._templateRef,
-				{
-					$implicit: this._dateAdapter.getDay(day),
-				},
-			);
+			const viewRef = this._viewContainerRef.createEmbeddedView(this._templateRef, {
+				$implicit: this._dateAdapter.getDay(day),
+			});
 			this._viewRefs.push(viewRef);
 		}
 
--- a/libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
+++ b/libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
@@ -1,4 +1,15 @@
-import { Directive, EmbeddedViewRef, OnDestroy, TemplateRef, ViewContainerRef, computed, inject } from '@angular/core';
+import {
+	ChangeDetectorRef,
+	computed,
+	Directive,
+	effect,
+	EmbeddedViewRef,
+	inject,
+	OnDestroy,
+	TemplateRef,
+	untracked,
+	ViewContainerRef,
+} from '@angular/core';
 import { injectDateAdapter } from '@spartan-ng/brain/date-time';
 import { injectBrnCalendar } from './brn-calendar.token';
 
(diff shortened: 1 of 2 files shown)


libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
```diff
@@ -33,12 +47,37 @@ export class BrnCalendarWeekdayDirective<T> implements OnDestroy {
 
 	constructor() {
 		// Create a new view for each day
-		for (const day of this.weekdays()) {
-			const viewRef = this._viewContainerRef.createEmbeddedView(this._templateRef, {
-				$implicit: this._dateAdapter.getDay(day),
-			});
+		effect(() => {
+			// Clear the view container before creating new views
+			this._viewContainerRef.clear();
```
goetzrobin, line 52: why are you clearing the ref here instead of doing in the renderWeekdays function?

libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
```diff
@@ -33,12 +47,37 @@ export class BrnCalendarWeekdayDirective<T> implements OnDestroy {
...
+		});
+	}
+
+	private _renderWeekdays(weekdays: T[]): void {
+		// Destroy all the views when the directive is destroyed
+		for (const viewRef of this._viewRefs) {
+			viewRef.destroy();
+		}
+
+		this._viewRefs.length = 0;
```
goetzrobin, line 67: Would this._viewRefs = [] be better? I am curious because I have never seen it done that way

libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
```diff
@@ -33,12 +47,37 @@ export class BrnCalendarWeekdayDirective<T> implements OnDestroy {
...
+			const viewRef = this._viewContainerRef.createEmbeddedView(
+				this._templateRef,
+				{
+					$implicit: this._dateAdapter.getDay(day),
+				},
+			);
 			this._viewRefs.push(viewRef);
 		}
+
+		this._changeDetector.detectChanges();
```
goetzrobin, line 80: Why is this needed?

libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
namdien177: This one was copied from the `renderWeeks` function. I also noticed this "hack" because `_viewRefs` is readonly, and the attempt here is to reset the array length. Actually, it would be better if we had this one be like `this.viewRefs = []` instead, but it would be a bit inconsistent because we have to make this variable not `readonly`.

Maybe we can change them both?

libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
namdien177: Also from the `renderWeeks` but I think this is a more safety thing, to make sure the view is reflected with the new context of the embedded view.

libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
namdien177: this one was my first attempt to clear the view before recreating the embedded view.
I think it's not necessary now since we already loop through the weekdays to destroy the DOM; therefore, I can remove this one.

(the only edge case I can think of is we have something that is not in our control to render in this view, and that "something" will not be cleared when we re-render the container)

libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
goetzrobin: @marcjulian do you know what the reasoning behind having viewRefs be readonly is? I am cool with keeping it that way, just curious.

libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
goetzrobin: I am okay with keeping it to have things be consistent!

libs/brain/calendar/src/lib/brn-calendar-weekday.directive.ts
marcjulian: I am also not sure if its necessary to keep viewRefs readonly, maybe @ashley-hunter knows a bit more about this.

ashley-hunter: No there's no particular reason this needs to be readonly, we can make it mutable if necessary

namdien177: Well, it's not necessary, but if we are good with that, I will make them mutable in the next commit to avoid this hack.

namdien177:
@goetzrobin could you or someone help me with the CI failure? I checked the log and it doesn't provide much information, and even running it on my local, it doesn't output anything...

ashley-hunter:
Hey, it's just the formatting of one of the files in incorrect, run 'pnpm format' and that should resolve the issue.

namdien177:
> Hey, it's just the formatting of one of the files in incorrect, run 'pnpm format' and that should resolve the issue.

image

Tried it but it doesn't output any diff to commit. No message to fix at all.

ashley-hunter:
Not sure why this wasn't being detected on your machine, but I have pushed the formatting fix for you.

PR #671, opened 2025-04-25
fix(switch): fix accessibility issues

### LINKED ISSUE #648: Switch - Accessibility issue of placement of `role: switch`

"@angular/core": "18.2.10"
"@spartan-ng/brain": "0.0.1-alpha.404"

switch

### Description

switch

```true
The brn-switch component is misconfigured in terms of accessibility, resulting in inability to test the component effectively.

Specifically, the matcher test matcher of `getByRole('switch')` is unable to be used due to the role of `switch` being placed in a nested <input> (see https://github.com/spartan-ng/spartan/blob/main/libs/brain/switch/src/lib/brn-switch.component.ts#L47).

Doing a `getByRole('switch').click()` does not actually toggle the switch. In order to fix this, the `role` of `switch` should be on the actual `brn-switch` component. Thanks!
```

### Other information

_No response_

- [ ] Yes
- [x] No

goetzrobin:
putting the role in the right place
association aria and data attributes with the container using native attributes (checked,required,disabled) on hidden input

here is where it gets fun:
identified issue where checked is not applied correctly and found this gem: https://github.com/angular/components/issues/26101 current approach uses preventDefault in toggle and breaks. this current version works but not if we click on the switch (background) and not the thumb what's even more fun: e2e tests pass because cypress clicks right in the middle where our thumb is.

open to suggestions and help!

fixes #648

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-otp
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [x] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

Closes #648

- [ ] Yes
- [x] No

## Other information

HELP. PLEASE

### DIFF
--- a/libs/brain/switch/src/lib/brn-switch.component.ts
+++ b/libs/brain/switch/src/lib/brn-switch.component.ts
@@ -1,6 +1,6 @@
 import { FocusMonitor } from '@angular/cdk/a11y';
 import { BooleanInput } from '@angular/cdk/coercion';
-import { isPlatformBrowser } from '@angular/common';
+import { DOCUMENT, isPlatformBrowser } from '@angular/common';
 import {
 	type AfterContentInit,
 	booleanAttribute,
@@ -14,6 +14,7 @@ import {
 	forwardRef,
 	inject,
 	input,
+	linkedSignal,
 	model,
 	type OnDestroy,
 	output,
@@ -39,25 +40,24 @@ let uniqueIdCounter = 0;
 
 @Component({
 	selector: 'brn-switch',
-	standalone: true,
 	template: `
 		<button
 			#switch
 			role="switch"
 			type="button"
 			[class]="class()"
-			[id]="forChild(state().id) ?? ''"
-			[name]="forChild(state().name) ?? ''"
+			[id]="getSwitchButtonId(state().id) ?? ''"
+			[name]="getSwitchButtonId(state().name) ?? ''"
 			[value]="checked() ? 'on' : 'off'"
 			[attr.aria-checked]="checked()"
 			[attr.aria-label]="ariaLabel() || null"
-			[attr.aria-labelledby]="mutableAriaLabelledby()() || null"
+			[attr.aria-labelledby]="mutableAriaLabelledby() || null"
 			[attr.aria-describedby]="ariaDescribedby() || null"
 			[attr.data-state]="checked() ? 'checked' : 'unchecked'"
 			[attr.data-focus-visible]="focusVisible()"
 			[attr.data-focus]="focused()"
 			[attr.data-disabled]="state().disabled()"
-			[disabled]="disabled()"
+			[disabled]="state().disabled()"
 			[tabIndex]="tabIndex()"
 			(click)="$event.preventDefault(); toggle()"
 		>
@@ -87,6 +87,7 @@ export class BrnSwitchComponent implements AfterContentInit, OnDestroy {
 	private readonly _elementRef = inject<ElementRef<HTMLElement>>(ElementRef);
 	private readonly _focusMonitor = inject(FocusMonitor);
 	private readonly _cdr = inject(ChangeDetectorRef);
+	private readonly _document = inject(DOCUMENT);
 
 	protected readonly focusVisible = signal(false);
 	protected readonly focused = signal(false);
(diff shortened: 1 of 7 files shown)


ashley-hunter:
Could you elaborate on what exactly the issue is when `preventDefault` is used? I notice clicking on the switch background does not work in some cases, but when I add the following to the `host:` property:

```ts
'(click)': 'toggle(); $event.preventDefault()',
```

That does appear to resolve the issue - am I missing something else?

I guess an alternative would be to evaluate if the `input` element is required at all? For example, Material does not include one in their implementation:

https://github.com/angular/components/blob/0e391700cb55786a74a3aa16757e90596eb55e6c/src/material/slide-toggle/slide-toggle.html

goetzrobin:
Yeah sorry I should have been more specific.

If you inspect the input element in your browser you will see that adding $event.preventDefault() messes with the checked attribute being applied to the input element.

It's not applied correctly once you add that line. The behavior is described in the linked issue in the PR, which has a stackblitz that shows this behavior also!

Does that help? 
Interesting to see that they are using a button. Do you think that's a better way to go about it?

ashley-hunter:
Yes, binding to `[attr.checked]` instead of `[checked]` seems to resolve the issue for me.

goetzrobin:
> Yes, binding to `[attr.checked]` instead of `[checked]` seems to resolve the issue for me.

Is that with the $event.preventDefault()? That would be a lovely fix!

ashley-hunter:
Yes 😊 that is with preventDefault. Seems to work, but if I have missed anything I'm happy to look again.

libs/brain/switch/src/lib/brn-switch.component.ts
```diff
@@ -36,48 +35,46 @@ export const BRN_SWITCH_VALUE_ACCESSOR = {
 
 const CONTAINER_POST_FIX = '-switch';
 
+let uniqueIdCounter = 0;
+
 @Component({
 	selector: 'brn-switch',
-	imports: [NgStyle],
+	standalone: true,
```
ashley-hunter, line 42: I think this can be removed now that we target Angular v19+ this is the default.

libs/brain/switch/src/lib/brn-switch.component.ts
```diff
@@ -36,48 +35,46 @@ export const BRN_SWITCH_VALUE_ACCESSOR = {
...
-		<ng-content select="brn-switch-thumb" />
+			[attr.aria-checked]="checked()"
+			[attr.aria-label]="ariaLabel() || null"
+			[attr.aria-labelledby]="mutableAriaLabelledby()() || null"
+			[attr.aria-describedby]="ariaDescribedby() || null"
+			[attr.data-state]="checked() ? 'checked' : 'unchecked'"
+			[attr.data-focus-visible]="focusVisible()"
+			[attr.data-focus]="focused()"
+			[attr.data-disabled]="state().disabled()"
+			[disabled]="disabled()"
```
ashley-hunter, line 60: I think this should also use the disabled state signal.

```suggestion
			[disabled]="state().disabled()"
```

libs/brain/switch/src/lib/brn-switch.component.ts
```diff
@@ -141,33 +175,23 @@ export class BrnSwitchComponent implements AfterContentInit, OnDestroy {
...
-			const label = parent?.querySelector(`label[for="${this.forChild(this.state().id)}"]`);
-			if (!label) return;
-			this._renderer.setAttribute(label, 'data-disabled', this.state().disabled() ? 'true' : 'false');
+			const isDisabled = this.state().disabled();
+			const state = this.state();
+			if (!this._elementRef.nativeElement || !this._isBrowser) return;
+
+			const switchElement = this._elementRef.nativeElement;
+
+			const labelElement = this._findLabelForSwitch(switchElement);
```
ashley-hunter, line 184: Rather than needing this `_findLabelForSwitch` method, could we just use `this._elementRef.nativeElement.closest('label')` to search the parents for a label?

ashley-hunter (changes requested):
Thanks! Looks great and works well! I agree, I think removing the input is probably the best option! Had a few very small suggestions, but really nice improvement!

ashley-hunter (approved):
Looks great, thank you!

PR #679, opened 2025-05-01
fix(radio-group): decoupling the label (#655)

### LINKED ISSUE #655: Radio component tightly couples the label, limiting flexibility

`radio-group`

---

### Information

Currently, the `Radio` component **"swallows" the `<label>` element**, meaning the label is rendered internally by the component itself:

```ts
<label [for]="inputId()" data-slot="label">
  <ng-content />
</label>
```

While this pattern is common in libraries like Angular Material, it introduces limitations when building a design system and composing custom UIs. Specifically:

- It **inconsistently diverges** from other components in Spartan like `Checkbox` and `Switch`, which do **not** include an internal label and allow external labeling.
- It **prevents flexible layout composition**, especially when integrating radio buttons into higher-level UI components like layout blocks, wrappers, or form field layouts.

Here’s a look at the current internal structure of the `Radio` component:

```ts
<input
  #input
  type="radio"
  [id]="inputId()"
  [checked]="checked()"
  [disabled]="disabledState()"
  [tabIndex]="tabIndex()"
  [attr.name]="radioGroup.name()"
  [attr.value]="value()"
  [required]="required()"
  (change)="onInputInteraction($event)"
  (click)="onInputClick($event)"
/>
<label [for]="inputId()" data-slot="label">
  <ng-content />
</label>
```

### Why this matters

Libraries like [Taiga UI](https://taiga-ui.dev/components/block) and [shadcn](https://ui.shadcn.dev/docs/components/radio-group) which Spartan NG proudly builds on top of—treat the radio and label as separate elements. This is much more composable and developer-friendly:

Shadcn: 
```tsx
<RadioGroup defaultValue="comfortable">
  <div className="flex items-center space-x-2">
    <RadioGroupItem value="default" id="r1" />
    <Label htmlFor="r1">Default</Label>
  </div>
  <div className="flex items-center space-x-2">
    <RadioGroupItem value="comfortable" id="r2" />
    <Label htmlFor="r2">Comfortable</Label>
  </div>
</RadioGroup>
```

Taiga: 
```angular2html
<label tuiBlock="s">
        <input
            formControlName="testValue5"
            size="s"
            tuiRadio
            type="radio"
            value="qiwi"
        />
        <span tuiTitle>
            <span>
                Qiwi
                <tui-icon tuiTooltip="Not the bird" />
            </span>
            <span tuiSubtitle>Green and sour</span>
        </span>
</label>
```

This separation allows for:

- Custom alignment and positioning of radio buttons and labels
- Uniform composition across different selection controls
- Easier accessibility testing and configuration

### Suggested Change

Remove the internal `<label>` and expose only the hidden native input. Developers should then be responsible for providing an external label using `<label for="...">`, just like `Checkbox` and `Switch`.

This would bring consistency, flexibility, and better alignment with Spartan's compositional philosophy.

---

The current workaround is to **style the projected content carefully** and override internal structure via CSS. But this is brittle and limits the ability to integrate `Radio` into larger, reusable UI layouts.

---

- [x] Yes
- [ ] No

josueggh:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] button
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-otp
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [x] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

Previously, the Radio component internally handled the <label> element, effectively 'swallowing' it by rendering the label within the component itself. With the update, the label has been decoupled and is now managed separately from the component.

Closes #655 

- [ ] Yes
- [x] No


libs/ui/radio-group/helm/src/lib/hlm-radio.component.ts
```diff
@@ -55,8 +72,24 @@ export class HlmRadioComponent<T = unknown> {
...
 
+	protected readonly state = computed(() => ({
+		disabled: signal(this.disabled()),
+	}));
 	/**
 	 * Event emitted when the checked state of this radio button changes.
 	 */
 	public readonly change = output<BrnRadioChange<T>>();
+
+	constructor() {
```
goetzrobin, line 83: We recently merged an improvement to the switch component that is also looking for a label and setting the right state on it. How do you feel about also using this approach:
https://github.com/spartan-ng/spartan/blob/5b4fd88dfd515f53902581245b28a7aef305e01e/libs/brain/switch/src/lib/brn-switch.component.ts#L178C2-L201C3

It also takes care of setting the right ariaLabelledby. Let me know if that makes sense and if you have any questions

goetzrobin (changes requested):
Looks great! Left one comment about a11y and label changes we can adopt from the switch component!

libs/ui/radio-group/helm/src/lib/hlm-radio.component.ts
josueggh: Make sense , I'll update the PR 👍

goetzrobin:
@maxhov do you know why the test would be failing in CI?

OmerGronich:
Ah you beat me to it!
Looks great👏 
Thanks for doing this

maxhov:
> @maxhov do you know why the test would be failing in CI?

Seems that it ran slower than expected in CI. I would suggest increasing the tolerance there.

Issue #324, opened 2024-07-08
Checkboxes within a scrollarea


elite-lucas:
### Please provide the environment you discovered this bug in.

Angular 18.0.1

### Which area/package is the issue in?

checkbox

### Description

If you have a checkbox within a scrollable area and this is in a dialogue, the hidden native HTML inputs increase the window size of the dialogue and thus create an empty area.

![pickA](https://github.com/goetzrobin/spartan/assets/115133536/e7461d17-1905-484d-8ec1-2353b28a1447)
![pickB](https://github.com/goetzrobin/spartan/assets/115133536/fff396ef-c963-41b9-9a71-3f9c14dcddaf)
![pickC](https://github.com/goetzrobin/spartan/assets/115133536/f677e62a-b033-4a2d-85fc-f7bc355bdad9)

### I would be willing to submit a PR to fix this issue

- [X] Yes
- [ ] No

goetzrobin:
This should be addressed by: https://github.com/spartan-ng/spartan/pull/682

ashley-hunter:
#682 has now been, merged. Closing as this now should be resolved.

PR #682, opened 2025-05-05
fix: rewrite checkbox to use button instead of input

### LINKED ISSUE #644: Checkbox - Accessibility Issue in brn-checkbox Component

"@angular/core": "18.2.10"
"@spartan-ng/brain": "0.0.1-alpha.404"

checkbox

### Description

The brn-checkbox component is failing accessibility tests due to improper implementation of ARIA attributes.

**Issue Description**
When running accessibility tests on the checkbox component, the following error is reported:
`'aria-conditional-attr' | 'serious' | "Ensure ARIA attributes are used as described in the specification of the element's role"`

**Root Cause**
The input element in the checkbox component has conflicting role and state management attributes:

1. Native <input type="checkbox"> is being used
2. Redundant explicit role="checkbox" is added
3. Using aria-checked="true" instead of the native checked attribute.

`<input tabindex="-1" type="checkbox" role="checkbox" value="on" id="default-checkbox" name="default-checkbox" aria-label="Default Checkbox" aria-checked="true" style="position: absolute; width: 1px; height: 1px; padding: 0px; margin: -1px; overflow: hidden; clip: rect(0px, 0px, 0px, 0px); white-space: nowrap; border-width: 0px;">`

**Expected Behavior**
The checkbox should comply with ARIA specifications, which state that when using native HTML elements, their implicit roles should be relied upon rather than explicitly declared, and their native state attributes should be used instead of ARIA equivalents when available.

```true
`'aria-conditional-attr' | 'serious' | "Ensure ARIA attributes are used as described in the specification of the element's role"`
```

### Other information

_No response_

- [ ] Yes
- [x] No

goetzrobin:
I switched the underlying implementation from a hidden input to a button. similar to the switch primitive. this fixes some a11y issues that occured with the previous implementation.

closes #644

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] calendar
- [ ] card
- [ ] carousel
- [x] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] form-field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-otp
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toast
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

Closes #644, #324

- [ ] Yes
- [ ] No


libs/brain/checkbox/src/lib/brn-checkbox.component.ts
```diff
@@ -30,72 +33,62 @@ export const BRN_CHECKBOX_VALUE_ACCESSOR = {
...
 		'[attr.data-disabled]': 'state().disabled()',
-		'[attr.aria-labelledby]': 'null',
-		'[attr.aria-label]': 'null',
-		'[attr.aria-describedby]': 'null',
-		'[attr.id]': 'hostId()',
-		'[attr.name]': 'hostName()',
 	},
 	providers: [BRN_CHECKBOX_VALUE_ACCESSOR],
 	changeDetection: ChangeDetectionStrategy.OnPush,
 	encapsulation: ViewEncapsulation.None,
```
ashley-hunter, line 79: I think this can be removed here as this component has no styles.

ashley-hunter (approved):
Thanks! A great update, really simplifies things! I left a few comments on things - none of which you actually changed, but just "tidy" up things in this area - but they can be done in a separate PR if you'd rather.

I noticed this example is still present but isn't relevant any more as we removed this input a while back.

image

libs/brain/checkbox/src/lib/brn-checkbox.component.ts
goetzrobin: I will check it out. I thought this was necessary to "redirect: the aria-label and aria-describedby inputs to the child, but I could be wrong

ashley-hunter: Hmm, GitHub seems to have highlighted different lines than I selected, I was only referring to the view encapsulation line, the rest are perfect 😊

goetzrobin: That makes perfect sense! I'll remove it!

ashley-hunter (approved):
Looks great, thank you for the updates! Big improvement!

PR #711, opened 2025-06-06
feat: new york theme

### LINKED ISSUE #167: New York style

Don't know / other

### Information

Right now the project only copies the "Default" style of shadcn/ui. I think it would be a great addition to also copy the "New York" style.

Although this is a lot of work, I think it will be better for the project in the long run since it will be more similiar to shadcn/ui itself and other ports (i.e. https://github.com/huntabyte/shadcn-svelte or https://github.com/radix-vue/shadcn-vue), which would allow developers some freedom to move between frameworks for different projects while still having the same style.

Some use cases for this:
- The design team designs a site using shadcn/ui (by using the figma file for example), but the development team prefers to use Angular based on the requirements or just out of preference.
- There are multiple projects for your site that use different technologies and frontend frameworks (for various reasons, like different development teams), but you want your "entire" site to have the same look and feel for the end users.

_No response_

- [ ] Yes
- [ ] No

ashley-hunter:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [x] accordion
- [x] alert
- [x] alert-dialog
- [ ] aspect-ratio
- [x] avatar
- [x] badge
- [x] breadcrumb
- [x] button
- [ ] calendar
- [ ] card
- [ ] carousel
- [x] checkbox
- [x] collapsible
- [x] combobox
- [x] command
- [x] context-menu
- [ ] data-table
- [ ] date-picker
- [x] dialog
- [x] dropdown-menu
- [ ] form-field
- [ ] hover-card
- [ ] icon
- [x] input
- [x] input-otp
- [x] label
- [x] menubar
- [ ] navigation-menu
- [x] pagination
- [x] popover
- [x] progress
- [x] radio-group
- [ ] scroll-area
- [x] select
- [ ] separator
- [x] sheet
- [ ] skeleton
- [x] slider
- [ ] sonner
- [ ] spinner
- [x] switch
- [ ] table
- [x] tabs
- [x] textarea
- [x] toggle
- [x] toggle-group
- [x] tooltip
- [ ] typography

Closes #167 

- [x] Yes
- [ ] No

## Other information

ShadCN has dropped support for the legacy default theme and removed it entirely from their docs. It doesn't make much sense to launch it v1 with a deprecated theme. This PR updates the helm components to use the New York theme.

Styles have been updated, and some of the features that previously were present are no longer present and have been removed.

Updated the CLI styles.

This is quite a large PR - I'm sure there is stuff I have missed.


ashley-hunter:
Thanks for the great feedback! 

I realised some of the colors were incorrect, I was using the old New York colors rather than the v4 colors - I have updated it, however there may still be slight differences, the reason is ShadCN now uses Tailwind 4, which uses oklch colors, whereas in Tailwind 3 we can't use them. As a result I have converted the colors to HSL, but this is not a 1:1 conversion due to them being in a different color space. This will improve if we update the docs to use Tailwind 4 - but I'll save that task for another day - but they should be close enough now.

> Background
> The background in darkmode is often grey on shadcn, i found this for multiple components
> Card, Combobox, Command, Context Menu, Date Picker (Trigger and Popup), DropDown, Hover Card, Menubar, Popover, Sonner

This should be fixed now the colors are updated - might be slightly different due to color space issue mentioned above

> Calendar
> image
> In dark mode the text color seems missing on the calendar when hovering.
> Also the square seems a little too big, it overlaps with selected day when hovering over the next.

Fixed

> Checkbox
> Nitpick: shadcn does not have a cursor_pointer. I saw this is the case for more components (select, date picker)
> I am not even sure if I like it..

Fixed

> Combobox
> The selected Checkmark is on the right in shadcn and on the left on spartan.
> It is a little bit inconvenient, because on the menubar it is no the left side.

Fixed

> Command
> Shadcn reacts on hover.

I will raise a follow on issue for this because this requires a change to the command component implementation

> Context Menu
> This maybe was already a thing before this change, I just noticed, that on click for a submenu, shadcn does not close popup for the next level.

Yep, pre-existing issue, will raise an issue

> Label
> DarkMode: the Color of the label is too bright. --foreground: rgb(250,250,250)
> on shadcn it comes from text-neutral-300 which is rgb(212,212,212)

This should be fixed now the colors are updated - might be slightly different due to color space issue mentioned above

> Scrollarea
> The scrollbar is hidden on shadcn and is inside the padding.
> I assume this should be a separate issue.

Now shows on hover, but the padding I don't think is configurable in ngx-scrollbar 🤔 unless i am mistaken

> Tabs
> image
> Shadcn has a border around the selected Tab

Fixed

> Input and Textarea
> The background color for input and Textarea seem a little bit off too. rgb(18,18,21) for spartan and rgb(21,21,21) for shadcn. (atleast this is what the color Picker says)
> Also the border seems to be a little bit darker on spartan.

This should be fixed now the colors are updated - might be slightly different due to color space issue mentioned above

elite-benni:
> > Hey Ashley, nice! one thing i noticed when running your PR, on serve and build i see 3 generated snippets. combobox, datepicker and scrollarea where generated. can you please check if they are generated on your machine too?
> 
> Ah yes, well spotted - I have them updated now. I wonder should we make the generated files ignored - as they will be automatically generated as part of the build so we don't really need them committed? Any thoughts?

thats a good point, we also ignored the api docs, so it would be the same for the generated code snippets. 
As you pointed this out, i was thinking if I should refactor this to a single json file too. But it has nothing to do with this issue.

elite-benni (approved):
I found some defaultCode exports in the preview files. 
It seems that there were some problems while merging. 
Not sure if we should tackle this in a follow up PR, because this one is  already extremly big.

I think it looks good now. 
It would be interesting, what the breaking changes are. 
Some of the variants were removed?

I like it, that you added additional userClass input. 

great job

ashley-hunter:
@goetzrobin @thatsamsonkid any further comments or are we happy to merge 😊?

goetzrobin:
Merged 😂

ashley-hunter:
😂 Thank you

PR #716, opened 2025-06-11
docs: autodocs improvements

### LINKED ISSUE #718: docs: AutoDocs issues

Prod Doc Site

Don't know / other

### Description

Defects:  
- Default values presented in the table seems to include casting in the value if used
- Column alignment/text overflow issues in certain cases

```true

```

### Other information

_No response_

- [x] Yes
- [ ] No

thatsamsonkid:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] form-field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-otp
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

Updates the glob pattern from using .directive.ts and .component.ts in anticipation of updating the component files to use the latest style guide file naming convention. Instead we are leveraging using 'hlm-' and 'brn-' file prefixes and excluding mock in file name

Closes #718 

- [ ] Yes
- [x] No


thatsamsonkid:
I simplified the test for the api docs generator, I think what I had was really testing implementation much more than actually just ensuring the actual generation was working and the all the mocks honestly was overly complex. It is now snapshot based so this should make actually catching any issues/differences when generating the docs. 

Lastly I also just made a target for the docs generation within the tools project.json and that will generate a separate json under dist folder, because the app related one gets generated directly in the public folder of the app. I also added a script that should make it easy to update the snapshot when intentional

libs/tools/src/executors/docs/generate-ui-docs/executor.ts
```diff
@@ -16,10 +16,10 @@ export default async function runExecutor(options: GenerateUiDocsExecutorSchema,
 
 	const project = new Project();
 	project.addSourceFilesAtPaths([
-		`${brainDir}/**/*.component.ts`,
-		`${brainDir}/**/*.directive.ts`,
-		`${helmDir}/**/*.component.ts`,
-		`${helmDir}/**/*.directive.ts`,
+		`${brainDir}/**/brn-!(*.spec).ts`,
+		`${helmDir}/**/hlm-!(*.spec).ts`,
+		// aspect ratio is odd one out
```
goetzrobin: should we fix aspect ratio?

goetzrobin:
@thatsamsonkid there is a conflict with ui-docs-section.component.ts but apart from that looks awesome!

libs/tools/src/executors/docs/generate-ui-docs/executor.ts
thatsamsonkid: I figured maybe we could do it as part of the style-guide changes we have planned since I assume that would need a migration script of some sort as well with all the file renames, what do you think

thatsamsonkid: At least i'm pretty sure in this case right since it's a helm file, if we change the name we need a migration script go with it right in this case or am i wrong

goetzrobin: no I think you are right!

thatsamsonkid:
@goetzrobin Lets go ahead and merge @ashley-hunter PR first, I don't want to throw that PR in conflict again, I'll resolve it on my end!

PR #727, opened 2025-06-17
fix: respect @nx/angular:library config in nx.json

koenigderluegner:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] form-field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-otp
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

Currently generating ui libraries always uses jest as default unit test runner and eslint as linter among other defaults because the project configuration is not taken into account (@nx/angular:library).

Closes #

## What is the new behavior?

Configuration from nx.json is used as base for creating libraries (@nx/angular:library, if existent).

- [ ] Yes
- [x] No

### DIFF
--- a/libs/cli/src/generators/base/lib/initialize-angular-library.ts
+++ b/libs/cli/src/generators/base/lib/initialize-angular-library.ts
@@ -1,4 +1,4 @@
-import { type Tree, joinPathFragments } from '@nx/devkit';
+import { type Tree, joinPathFragments, readNxJson } from '@nx/devkit';
 import type { HlmBaseGeneratorSchema } from '../schema';
 
 export async function initializeAngularLibrary(tree: Tree, options: HlmBaseGeneratorSchema) {
@@ -9,6 +9,7 @@ export async function initializeAngularLibrary(tree: Tree, options: HlmBaseGener
 			'@nx/angular/generators'
 		)
 	).libraryGenerator(tree, {
+		...(readNxJson(tree).generators?.['@nx/angular:library'] || {}),
 		name: options.publicName,
 		skipFormat: true,
 		simpleName: true,


libs/cli/src/generators/base/lib/initialize-angular-library.ts
```diff
@@ -9,6 +9,7 @@ export async function initializeAngularLibrary(tree: Tree, options: HlmBaseGener
 			'@nx/angular/generators'
 		)
 	).libraryGenerator(tree, {
+		...(readNxJson(tree).generators?.['@nx/angular:library'] || {}),
```
goetzrobin, line 12: should this go in the end? so one can set importpath / prefix etc? or would we need to make this configurable through the options object?

koenigderluegner: I‘ve put it in the beginning because it seems like the other existing values are mandatory for the rest of the generation to work properly. I don’t know which implications it might have to let users overwrite everything - especially import path.

edit: for example in this file the selector is hard coded instead of checking for the prefix, this would cause at least lint warnings/errors and maybe worse depending on the rest.

https://github.com/spartan-ng/spartan/blob/main/libs/cli/src/generators/ui/libs/ui-accordion-helm/files/lib/hlm-accordion-content.component.ts.template

(Sorry for the odd linking, currently on mobile)

goetzrobin: Good point! need to think through these things if we want to allow our options to be customizable!

PR #766, opened 2025-07-15
refactor: use the latest angular naming conventions

### LINKED ISSUE #717: Update primitive file names to follow new Angular v20 style guide updates

Don't know / other

### Information

Update the repo and the end user primitive files to follow the new naming convention described in angular latest  [style guide](https://angular.dev/style-guide) update. 

Minor subtask update aspect-ratio helm files to be prefixed with 'hlm' instead of 'helm' and remove specisl glob from docs executor. Best opportunity to do it since migration script will most likely be required for the above.

_No response_

- [ ] Yes
- [ ] No

ashley-hunter:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [x] accordion
- [x] alert
- [x] alert-dialog
- [x] aspect-ratio
- [x] avatar
- [x] badge
- [x] breadcrumb
- [x] button
- [x] calendar
- [x] card
- [x] carousel
- [x] checkbox
- [x] collapsible
- [x] combobox
- [x] command
- [x] context-menu
- [x] data-table
- [x] date-picker
- [x] dialog
- [x] dropdown-menu
- [x] form-field
- [x] hover-card
- [x] icon
- [x] input
- [x] input-otp
- [x] label
- [x] menubar
- [x] navigation-menu
- [x] pagination
- [x] popover
- [x] progress
- [x] radio-group
- [x] scroll-area
- [x] select
- [x] separator
- [x] sheet
- [x] skeleton
- [x] slider
- [x] sonner
- [x] spinner
- [x] switch
- [x] table
- [x] tabs
- [x] textarea
- [x] toggle
- [x] toggle-group
- [x] tooltip
- [x] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [X] cli

Closes #717

- [X] Yes
- [ ] No

## Other information

Apologies for yet another large PR - I think this should be the last. This updates everything to use the latest Angular style guide conventions.

While this is a breaking change, we have a healthcheck as part of this PR which should rename all identifiers to use the new naming conventions


apps/app/src/app/pages/(components)/components.page.ts
```diff
@@ -57,7 +44,7 @@ export const routeMeta: RouteMeta = {
 		<spartan-page />
 	`,
 })
-export default class ComponentsPageComponent {
+export default class sPage {
```
ajitzero:
```suggestion
export default class ComponentsPage {
```

ashley-hunter: Well spotted! Thank you!

apps/app/src/app/pages/(components)/components/(accordion)/accordion.preview.ts
```diff
@@ -72,7 +72,7 @@ import { HlmIconDirective } from '@spartan-ng/helm/icon';
 		</div>
 	`,
 })
-export class AccordionPreviewComponent {}
+export class AccordionPreview {}
```
ajitzero, line 76: I think Line 80 needs a renaming `HlmAccordionContentComponent` -> `HlmAccordionContent`
And all the *.template file.

Similar changes needed in multiple *.preview.ts files (I just did a global search for "Component"/"Directive"). I mentioned some of them here that have a slight documentation change needed.
- apps/app/src/app/pages/(components)/components/(breadcrumb)/breadcrumb.page.ts Line 118
- apps/app/src/app/pages/(components)/components/(breadcrumb)/breadcrumb.preview.ts Line 56, 71, 76

ajitzero: Try this regex: `(Component|Directive|Pipe),?$`

ajitzero: image

ajitzero (approved):
LGTM, Thanks for the changes!

goetzrobin:
@ashley-hunter no need to apologize at all! thanks so much!

elite-benni:
I found some issues in here, i try to figure them out exactly by running it against a nx repo. 
 
### brn-radio 
the brn-radio healthcheck is broken - `healthchecks/brn-radio.ts`			
`if (file.endsWith('hlm-radio.component.ts')) {`
change to
`if (file.endsWith('src/lib/hlm-radio.ts')) {`

Additionally I think we also should update modulespecifier helm too.  (in generator and healthcheck.) 
otherwise it breaks a all helm imports. 
-			if (moduleSpecifier.includes('@spartan-ng/brain/')) {
+			if (moduleSpecifier.includes('@spartan-ng/brain/') || moduleSpecifier.includes('@spartan-ng/helm/')) {

ashley-hunter:
Thanks very much for the feedback! Conflicts should be resolved!

- Fixed brn-radio healthcheck
- As for migrating helm components/directives - I wasn't sure if we want to do this - that is user code rather than ours - so do we want to make these changes as they are convention updates, not to avoid breaking changes like our other migrations? Or should that happen just if they regenerate?

> why is the renaming called with directive?	BrnSelectValueDirective: 'BrnSelectValueTemplateDirective',

That was a mistake 😊 well spotted!

- HlmMenuItemRadio - I did rename the file to hlm-menu-item-radio-indicator - were you seeing something different?

elite-benni:
**Generating Helm**
you are right, completely forgot about this again. I think having the option to also migrate the helm Component Imports in the users codebase would be nice for users that use Helm as is. It would be nice to have it either as option in the healthchecks or as part of `migrate-helm-libraries`.
Now that you pointed that out, I think the correct place would be when they regenerate. 

**HlmMenuItemRadio** 
Sorry that I did not write that correctly. I meant the generator, which did not replace helm at this time.
When we decide to provide this migration for helm, we have to add the `renamedIdentifiers`. 
There was a HlmMenuItemRadioDirective and a HlmMenuItemRadioComponent
Both would be renamed to HlmMenuItemRadio with the migrator, this causes errors.

ashley-hunter:
Thank you! In that case I think everything should be updated that needs to - we could add a separate health check in future to migrate the helm libraries to use the new conventions, although that may be better as a separate PR as this is large enough, and it would need some additional logic to rename files and update import paths etc.

elite-benni:
Ok now I see where the problem was.
I was not on the latest version of spartan, so many errors came from that and not from the migration of this PR.

one thing I still would want to be addressed or have checked is the change of the `FormGroupDirective` to `FormGroup`.
Was this change intended?

you are right, this PR is already pretty large and thats the reason i comment everything here instead of the code directly. :/

We should then be pretty quick with the hlm change or merge it at the same time, if possible, because running the helm-migration results in a completely broken codebase.

ashley-hunter:
Ah ok, I see what you mean, of course yeah all the imports will be different. I guess you're right tben, we should include that, I'll make the updates in this PR, as you said otherwise it could break. Thanks for the feedback

libs/cli/src/generators/migrate-naming-convention/generator.ts
```diff
@@ -78,4 +81,38 @@ function updateIdentifiers(tree: Tree) {
...
 
+/**
+ * Helm identifiers are not necessarily just imports as the class declarations
+ * are defined in the users codebase so we need to update all identifiers that
+ * start with `Hlm` and end with `Directive` or `Component`.
+ * We also must check for known identifiers that have been renamed.
+ */
+function updateHelmIdentifiers(tree: Tree) {
+	visitFiles(tree, '/', (path) => {
+		let content = tree.read(path).toString('utf-8');
```
elite-benni, line 100: I guess we should add something like this here. 

```
		if (!path.endsWith('.html') && !path.endsWith('.ts')) {
			return;
		}
```

otherwise we overwrite binaryfiles as utf-8 string.

Also it should speed up the generator a lot.  

This does not happen on `updateBrainIdentifiers`  because we have a early return if it does not find importDeclarations, but i would also add such a check to the brain updater

ashley-hunter: Good idea! I have added this in for both to only process .ts files, thank you

elite-benni:
LGTM! :)
Thanks for all the changes!

PR #782, opened 2025-07-25
fix(select): voiceover navigatability and typehead restored

### LINKED ISSUE #773: select: not navigatable with VO screenreader

Latest prod version

This may not be limited to Voiceover, it could be an issue with Jaws/NVDA but would need to check and or if someone else could verify that.

select

### Description

When trying to navigate the Select component with just the keyboard with Voiceover, the select dropdown opens fine and VO reads the first option but from there the down arrow does not take you to the next option, instead VO simply re-reads the same option and then the up arrow seems to randomly take me outside of the select dropdown list (in my case it jumped to my browser bookmarks section). 

Keyboard navigation works fine without Voiceover, i think unfortunately it's some weirdness with the way VO is handling active descendant and I believe it simply doesnt like that we have an intermediate div element between the element that has the activedescendant and listbox role and the actual list. Idk if that's the only issue, in some research I found VO and active descendant in safari can be an issue but not sure if that's old news or not. The radix select works fine and we have a very similar html structure but I believe they rely on the tab-index focus strategy for the listbox.

```true

```

### Other information

_No response_

- [x] Yes
- [ ] No

thatsamsonkid:
- fix focus styling

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] form-field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-otp
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [x] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli

Closes #773 

- [ ] Yes
- [ ] No

Not sure if we consider the outline change a breaking change? Might mess with peoples styles?

### DIFF
--- a/libs/brain/select/src/lib/brn-select-content.ts
+++ b/libs/brain/select/src/lib/brn-select-content.ts
@@ -12,13 +12,12 @@ import {
 	effect,
 	inject,
 	signal,
-	untracked,
 	viewChild,
 } from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { BrnSelectOption } from './brn-select-option';
 
-import { ActiveDescendantKeyManager } from '@angular/cdk/a11y';
+import { FocusKeyManager } from '@angular/cdk/a11y';
 import { Directive } from '@angular/core';
 import { Subject, fromEvent, interval } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
@@ -88,11 +87,9 @@ export class BrnSelectScrollDown {
 	changeDetection: ChangeDetectionStrategy.OnPush,
 	host: {
 		role: 'listbox',
-		tabindex: '0',
 		'[attr.aria-multiselectable]': '_select.multiple()',
 		'[attr.aria-disabled]': '_select.disabled() || _select.formDisabled()',
 		'aria-orientation': 'vertical',
-		'[attr.aria-activedescendant]': 'keyManager?.activeItem?.id()',
 		'[attr.aria-labelledBy]': '_select.labelId()',
 		'[attr.aria-controlledBy]': "_select.id() +'--trigger'",
 		'[id]': "_select.id() + '--content'",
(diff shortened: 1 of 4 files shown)


thatsamsonkid:
Ok I confirmed this is working as expected with Windows + NVDA and on Mac

libs/brain/select/src/lib/brn-select-content.ts
```diff
@@ -170,33 +167,39 @@ export class BrnSelectContent<T> implements AfterContentInit {
 	}
 
 	ngAfterContentInit(): void {
-		this.keyManager = new ActiveDescendantKeyManager(this._options, this._injector)
+		this.keyManager = new FocusKeyManager(this._options())
```
ashley-hunter: The focus key manager now allows a signal to be passed directly for the options ensuring things stay updated if the options change, we just need to also pass the injector.

```suggestion
		this.keyManager = new FocusKeyManager(this._options, this._injector)
```

ashley-hunter (approved):
Looks good to me! Had one small suggestion - but its just a nice to have 😊

libs/brain/select/src/lib/brn-select-content.ts
thatsamsonkid: @ashley-hunter - thanks for this, I wasn't aware. I'll update it!

Issue #813, opened 2025-08-08
`"strict": false` gets added to `@nx/angular:library` generator config


Ban117:
### Please provide the environment you discovered this bug in.

Using NodeJS 22.17.1, no globals, see `package.json` at the end

Created new Nx/Angular project:
`npx create-nx-workspace@latest some-project --preset=angular-monorepo`

Installed SpartanUI:
`npm i -D @spartan-ng/cli`

Generated a component
`npx nx g @spartan-ng/cli:ui button`

Notice `nx.json` has been modified and that `"strict": false` has been added to the config for the `@nx/angular:library` generator
```
// `nx.json`
"@nx/angular:library": {
      "linter": "eslint",
      "unitTestRunner": "none",
      "strict": false // 👈 this is the issue
    },
```

```
{
  "name": "@some-project/source",
  "version": "0.0.0",
  "license": "MIT",
  "scripts": {},
  "private": true,
  "dependencies": {
    "@angular/cdk": "18.0.0",
    "@angular/common": "~20.1.0",
    "@angular/compiler": "~20.1.0",
    "@angular/core": "~20.1.0",
    "@angular/forms": "~20.1.0",
    "@angular/platform-browser": "~20.1.0",
    "@angular/platform-browser-dynamic": "~20.1.0",
    "@angular/router": "~20.1.0",
    "@spartan-ng/brain": "0.0.1-alpha.497",
    "class-variance-authority": "^0.7.0",
    "clsx": "^2.1.1",
    "rxjs": "~7.8.0",
    "zone.js": "~0.15.0"
  },
  "devDependencies": {
    "@angular-devkit/core": "~20.1.0",
    "@angular-devkit/schematics": "~20.1.0",
    "@angular/build": "~20.1.0",
    "@angular/cli": "~20.1.0",
    "@angular/compiler-cli": "~20.1.0",
    "@angular/language-service": "~20.1.0",
    "@eslint/js": "^9.8.0",
    "@nx/angular": "21.3.11",
    "@nx/devkit": "21.3.11",
    "@nx/eslint": "21.3.11",
    "@nx/eslint-plugin": "21.3.11",
    "@nx/jest": "21.3.11",
    "@nx/js": "21.3.11",
    "@nx/playwright": "21.3.11",
    "@nx/web": "21.3.11",
    "@nx/workspace": "21.3.11",
    "@playwright/test": "^1.36.0",
    "@schematics/angular": "~20.1.0",
[...]

PR #814, opened 2025-08-08
cli: Configurable Library Structure in component.json

### LINKED ISSUE #752: One entrypoint for hlm components

Don't know / other

### Information

It would be great to have the ability to configure the library structure in the component.json file. Specifically, the Spartan CLI currently creates a new library for each component by default. However, in some project structures like ours we prefer to group all components within a single library.

Adding a configuration option (e.g. libraryMode: "single" | "multi") would allow teams to choose between:

Multi-library mode: The current default behavior—each component lives in its own dedicated library.

Single-library mode: All components are generated into a shared library (e.g., @my-org/ui), matching common monorepo setups.

This would offer more flexibility for teams with different architectural preferences, and help align Spartan NG with a broader range of real-world repository structures.

_No response_

- [ ] Yes
- [x] No

MerlinMoos:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] form-field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-otp
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [x] cli

currenlty its only possible to generate the components into multi libs

Closes #752 #813 

## What is the new behavior?

Components can be generated into a single library

- [ ] Yes
- [x] No

### DIFF
--- a/libs/cli/src/generators/base/generator.ts
+++ b/libs/cli/src/generators/base/generator.ts
@@ -6,7 +6,6 @@ import {
 	joinPathFragments,
 	runTasksInSerial,
 	updateJson,
-	visitNotIgnoredFiles,
 } from '@nx/devkit';
 import { addTsConfigPath } from '@nx/js';
 import * as path from 'node:path';
@@ -18,6 +17,8 @@ import { deleteFiles } from './lib/deleteFiles';
 import { getTargetLibraryDirectory } from './lib/get-target-library-directory';
 import { initializeAngularLibrary } from './lib/initialize-angular-library';
 
+import { librarySecondaryEntryPointGenerator } from '@nx/angular/generators';
+import { singleLibName } from './lib/single-lib-name';
 import type { HlmBaseGeneratorSchema } from './schema';
 import { FALLBACK_ANGULAR_VERSION } from './versions';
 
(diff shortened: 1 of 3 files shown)


MerlinMoos:
the "nx g @spartan-ng/cli:migrate-helm-libraries" is currenlty failing, I'm working on that

MerlinMoos:
I found a issue with the buildable option true, the build is not including the files, maybe someone is having an idea

MerlinMoos:
@goetzrobin @ashley-hunter with the current codebase the libs are getting generated like 

image


this is not the default behavior from nx, from my point of view we should do it more like 

image


because inside of the src folder should be all generated components, this is the nx behavior. Its open to discuss.

libs/cli/src/generators/base/schema.d.ts
```diff
@@ -8,4 +8,5 @@ export interface HlmBaseGeneratorSchema {
 	peerDependencies?: Record<string, string>;
 	angularCli?: boolean;
 	buildable: boolean;
+	multiLibs: boolean;
```
ashley-hunter: Could we make this something like `mode: 'library' | 'entrypoint'` which would let us potentially support additional strategies in future. `mode` may not be the best name, open to suggestions here.

MerlinMoos: Yes this is a great idea, i will chance that

ashley-hunter: Thank you, just could be useful as there was some discussion in the past about potentially supporting Angular CLI projects, so just in case we ever did want to support that it keeps it flexible.

ashley-hunter:
> Okay then I will keep it like it is, maybe you have an idea why the build is not working?

I haven't checked, but I'm going to guess in the tsconfig.lib.json file it's only including files in the src directory, if so that just needs updated.

MerlinMoos:
I added the **/*.ts in the config. I will check it tomorrow and change the config flag name :)

ashley-hunter:
> I added the **/*.ts in the config. I will check it tomorrow and change the config flag name :)

Awesome; thank you! Any issues let me know 😊

MerlinMoos:
hey @ashley-hunter i fixed the build the ng-package.json was missing and I renamed the config flag, now its ready to review :)

libs/cli/src/generators/base/generator.ts
```diff
@@ -5,55 +5,124 @@ import {
...
 	}
 
-	// we need to delete the src/lib/{publicName} directory because it is generated by nx and cannot be disabled
-	deleteFiles(tree, joinPathFragments(options.directory, options.publicName, 'src', 'lib', options.publicName));
+	if (options.buildable) {
+		const ngPackageJson = {
+			lib: {
+				entryFile: 'index.ts',
+			},
+		};
```
ashley-hunter: This file should be here regardless of whether it is buildable or not - the only thing that would matter would be whether there is a "build" target in the library project.json

libs/cli/src/generators/base/generator.ts
```diff
@@ -5,55 +5,124 @@ import {
...
+		tree.rename(libDir, srcDir);
 	}
 
-	if (options.angularCli) {
-		addTsConfigPath(tree, tsConfigAliasToUse, [
-			`./${joinPathFragments(targetLibDir, 'src', 'index.ts').replace(/\\/g, '/')}`,
-		]);
-	} else {
-		tasks.push(await initializeAngularLibrary(tree, options));
+	// Replace @spartan-ng/helm/<generic> with ../../ui-<generic>-helm in all src files
```
ashley-hunter: Could it be simpler if we named the folder `lib/ui/button` rather than `lib/ui/ui-button-helm` that way we wouldn't have to rename all the import paths? If that would work it seems like this would be less "risky" as we wouldn't have to modify source code.

libs/cli/src/generators/base/generator.ts
```diff
@@ -5,55 +5,124 @@ import {
...
+	return alias in existingPaths;
+}
+
+function setupTsConfigAlias(tree: Tree, alias: string, targetLibDir: string) {
+	addTsConfigPath(tree, alias, [`./${joinPathFragments(targetLibDir, 'src', 'index.ts').replace(/\\/g, '/')}`]);
+}
 
-	const existingPathsByAlias = readTsConfigPathsFromTree(tree);
-	const tsConfigAliasToUse = `@spartan-ng/helm/${options.primitiveName}`;
+function generateEntrypointFiles(tree: Tree, targetLibDir: string, alias: string, options: HlmBaseGeneratorSchema) {
```
ashley-hunter: Can we use the Nx generator for creating the secondary entry point rather than doing it ourselves?

```ts
import { librarySecondaryEntryPointGenerator } from '@nx/angular/generators';
```

MerlinMoos: yes but sadly its only working for buildable or publishable libraries, I changed the logic and now we don't do any replacements

ashley-hunter: Thank you so much! No problem, I'll look into it! Thanks

PR #853, opened 2025-08-25
feat: autocomplete

### LINKED ISSUE #186: RFC: Autocomplete

Don't know / other

### Information

Like `select` but with input.

_No response_

- [X] Yes
- [ ] No

marcjulian:
* add example for async filter w/ new resource
* add example for form

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [x] autocomplete
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [x] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] form-field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-otp
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli

Closes #186

- [ ] Yes
- [x] No


apps/app/src/app/pages/(components)/components/(autocomplete)/autocomplete.page.ts
```diff
@@ -0,0 +1,112 @@
...
+import { AutocompletePreview, defaultImports, defaultSkeleton } from './autocomplete.preview';
+
+export const routeMeta: RouteMeta = {
+	data: { breadcrumb: 'Autocomplete' },
+	meta: metaWith('spartan/ui - Autocomplete', 'Autocomplete input and dropdown selection with filtering options.'),
+	title: 'spartan/ui - Autocomplete',
+};
+
+@Component({
+	selector: 'spartan-combobox',
```
MerlinMoos: The selector should not be combobox

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
```diff
@@ -0,0 +1,197 @@
...
+	changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class HlmAutocomplete implements ControlValueAccessor {
+	protected readonly _elementRef = inject(ElementRef<HTMLElement>);
+	protected readonly _popoverState = signal<BrnDialogState | null>(null);
+
+	/** The list of options to display in the autocomplete. */
+	public readonly options = input<string[]>([]);
+
+	/** The selected value. */
```
MerlinMoos, line 165: Please change this to an input and a output for two way binding. You are set the value in the writeValue and this is trigger a change which is not correct :)

marcjulian: Updated it to use input, linkedSignal and output.

libs/cli/src/generators/ui/libs/ui-autocomplete-helm/files/lib/hlm-autocomplete.ts.template
```diff
@@ -0,0 +1,202 @@
...
+			[align]="'start'"
+			[autoFocus]="'first-heading'"
+			[closeOnOutsidePointerEvents]="true"
+			[state]="_popoverState()"
+			(stateChanged)="_stateChanged($event)"
+			sideOffset="5"
+			closeDelay="100"
+		>
+			<hlm-command class="border-border w-full border">
+				<hlm-command-search class="h-full" hlmAutocompleteTrigger [disabled]="!_filter()">
```
ashley-hunter, line 76: There seems to be a double border here, noticeably slightly thicker on the bottom - is this intentional?

image

libs/helm/autocomplete/src/lib/hlm-autocomplete-trigger.ts
```diff
@@ -0,0 +1,31 @@
+import { BooleanInput } from '@angular/cdk/coercion';
+import { booleanAttribute, Directive, ElementRef, inject, input } from '@angular/core';
+import { BrnDialog } from '@spartan-ng/brain/dialog';
+
+@Directive({
+	selector: '[hlmAutocompleteTrigger]',
+	host: {
+		'(click)': 'open()',
```
ashley-hunter, line 8: Should we also open on arrow down?

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
```diff
@@ -0,0 +1,202 @@
...
+		HlmAutocompleteTrigger,
+
+		NgIcon,
+		HlmIcon,
+	],
+	providers: [HLM_AUTOCOMPLETE_VALUE_ACCESSOR, provideIcons({ lucideSearch, lucideChevronDown })],
+	template: `
+		<brn-popover
+			[align]="'start'"
+			[autoFocus]="'first-heading'"
```
ashley-hunter: We can remove thing bindings for these static values.

```suggestion
			align="start"
			autoFocus="first-heading"
```

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
```diff
@@ -0,0 +1,202 @@
...
+			[closeOnOutsidePointerEvents]="true"
+			[state]="_popoverState()"
+			(stateChanged)="_stateChanged($event)"
+			sideOffset="5"
+			closeDelay="100"
+		>
+			<hlm-command class="border-border w-full border">
+				<hlm-command-search class="h-full" hlmAutocompleteTrigger [disabled]="!_filter()">
+					<ng-icon name="lucideSearch" hlm />
+					<input
```
ashley-hunter, line 77: If I press enter in the text input whenever the dropdown is closed I get the following:

image

apps/app/src/app/pages/(components)/components/(autocomplete)/autocomplete.page.ts
```diff
@@ -0,0 +1,112 @@
...
+			/>
+
+			<spartan-section-sub-heading id="usage">Usage</spartan-section-sub-heading>
+			<div class="space-y-4">
+				<spartan-code [code]="_defaultImports" />
+				<spartan-code [code]="_codeSkeleton" />
+			</div>
+
+			<spartan-section-sub-heading id="examples">Examples</spartan-section-sub-heading>
+			<h3 id="examples__async" class="${hlmH4} mb-2 mt-6">Asynchronos</h3>
```
ashley-hunter:
```suggestion
			<h3 id="examples__async" class="${hlmH4} mb-2 mt-6">Asynchronous</h3>
```

apps/app/src/app/pages/(components)/components/(autocomplete)/autocomplete--async.example.ts
```diff
@@ -0,0 +1,46 @@
+import { ChangeDetectionStrategy, Component, inject, resource, signal } from '@angular/core';
+import { FormBuilder, Validators } from '@angular/forms';
+import { HlmAutocomplete } from '@spartan-ng/helm/autocomplete';
+
+@Component({
+	selector: 'spartan-autocomplete-async',
+	imports: [HlmAutocomplete],
+	template: `
+		<hlm-autocomplete [options]="options.value()" (filterChange)="_filter.set($event)" />
```
ashley-hunter: When we start searching it shows "No options found" - is there a way to showing a loading state to differentiate between no options because we are loading and no options because there were no matches?

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
```diff
@@ -0,0 +1,202 @@
...
+			[align]="'start'"
+			[autoFocus]="'first-heading'"
+			[closeOnOutsidePointerEvents]="true"
+			[state]="_popoverState()"
+			(stateChanged)="_stateChanged($event)"
+			sideOffset="5"
+			closeDelay="100"
+		>
+			<hlm-command class="border-border w-full border">
+				<hlm-command-search class="h-full" hlmAutocompleteTrigger [disabled]="!_filter()">
```
ashley-hunter: The input height is larger than all other controls like input, select, combobox etc.. and there is no box shadow:

image

Is this intentional? Or should they align?

libs/cli/src/generators/ui/libs/ui-autocomplete-helm/files/lib/hlm-autocomplete.ts.template
marcjulian: Ah yes, I removed the additional `border-b` from `hlm-command-search`

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
marcjulian: Updated the styles to align with input

libs/helm/autocomplete/src/lib/hlm-autocomplete-trigger.ts
marcjulian: I added the keydown event to the input, as this won't be triggered here.

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
marcjulian: The error comes from `brn-command`, I will have a look what to do here

https://github.com/spartan-ng/spartan/blob/c7e1c9c0c97db601f40eabe6771b1e1702bc4dfb/libs/brain/command/src/lib/brn-command.ts#L98-L101

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
marcjulian: CleanShot 2025-08-27 at 13 14 45

apps/app/src/app/pages/(components)/components/(autocomplete)/autocomplete--async.example.ts
marcjulian: Good idea! I added an option for loading indicator which changes the content in the empty state. I added a spinner to the example. When you enter `empty` it should show "No options found" instead of the loading spinner.

What do you think of the approach?

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
```diff
@@ -0,0 +1,236 @@
...
+					<button
+						class="flex items-center justify-center outline-none disabled:cursor-not-allowed [&>_ng-icon]:opacity-50"
+						tabindex="-1"
+						type="button"
+						aria-label="Toggle dropdown"
+						[disabled]="_disabled()"
+						(click)="_toggleOptions()"
+					>
+						<ng-icon name="lucideChevronDown" hlm />
+					</button>
```
marcjulian: @ashley-hunter do you might have an idea on how to avoid this behavior?

ashley-hunter: Sure, I think if we avoid trying to sync the state and instead get a view child of the popover and call the open and close methods on it directly, that should avoid this issue.

apps/app/src/app/pages/(components)/components/(autocomplete)/autocomplete--async.example.ts
ashley-hunter: I think that's great! Thanks!

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
marcjulian: Awesome idea! That works excellent

libs/helm/autocomplete/src/index.ts
```diff
@@ -1,11 +1,23 @@
...
+export * from './lib/hlm-autocomplete-search';
+export * from './lib/hlm-autocomplete-search-input';
 export * from './lib/hlm-autocomplete-trigger';
 
-export const HlmAutocompleteImports = [HlmAutocomplete, HlmAutocompleteTrigger] as const;
+export const HlmAutocompleteImports = [
+	HlmAutocomplete,
+	HlmAutocompleteCommand,
+	HlmAutocompleteSearch,
+	HlmAutocompleteSearchInput,
```
marcjulian: @ashley-hunter I copied over the command, command search and command search input for autocomplete. This way we don't need to override existing styles from command components and won't risk future styles to break autocomplete.

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
```diff
@@ -0,0 +1,244 @@
...
+		value.toLowerCase().includes(search.toLowerCase());
+
+	/** A custom filter function to determine which items are visible when searching. */
+	public readonly filter = input<CommandFilter>(this._defaultFilter);
+
+	/** The search query. */
+	public readonly search = input<string>('');
+	protected readonly _search = linkedSignal(() => this.search());
+
+	protected readonly _filteredOptions = computed(() =>
```
marcjulian: @ashley-hunter I am thinking about changing `_filteredOptions` to an input to replace `options`. This would move the filter obligation to the ui when the autocomplete is used. What do you think about that? Do you think this makes it more complicated? Otherwise we could add another input to override the filter logic or like with supporting object selection the autocomplete must be copied and adjusted to the developer needs.

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
```diff
@@ -0,0 +1,244 @@
...
+	providers: [HLM_AUTOCOMPLETE_VALUE_ACCESSOR, provideIcons({ lucideSearch, lucideChevronDown })],
+	template: `
+		<brn-popover
+			align="start"
+			autoFocus="first-heading"
+			sideOffset="5"
+			closeDelay="100"
+			[closeOnOutsidePointerEvents]="true"
+		>
+			<hlm-autocomplete-command [filter]="filter()">
```
marcjulian: @ashley-hunter this filter is used to set the items to visible/invisible. Should we just set it to true = visible, because options are filtered already?

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
```diff
@@ -0,0 +1,244 @@
...
+					class="max-h-60 overflow-y-auto p-0"
+					[style.width.px]="_elementRef.nativeElement.offsetWidth"
+				>
+					<hlm-command-list>
+						<hlm-command-group>
+							@for (option of _filteredOptions(); track option) {
+								<button hlm-command-item [value]="option" (selected)="_optionSelected(option)">
+									{{ option }}
+								</button>
+							}
```
marcjulian: @ashley-hunter What if we want to allow the autocomplete to filter not just `string[]` but any objects. I guess this will add complexity to the autocomplete, not sure if copy pasting it for user specific use-cases would be better in this case. Keen to hear your feedback.

For example `countries: {name: string, code: string}[]` where the value would be the `code` and `name` the display label.

ashley-hunter: Yeah, I like allowing other data types, and I think your other suggestion of moving filtering to userspace would allow this much more easily.

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
ashley-hunter: Yeah, I think that makes sense

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
ashley-hunter: Yes, I like that, allows for much more complex filtering and on your other point this would help support other option types.

libs/helm/autocomplete/src/index.ts
ashley-hunter: Sounds good, makes sense!

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
marcjulian: I added provideBrnCommandConfig and injectBrnCommandConfig to override the default filter easily

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
marcjulian: I am handling this with an input `focusActiveOnEnter` which is by default true, but only true for the autocomplete when the popover is open.

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
marcjulian: I am trying to pull the for-loop out so the value and selected option can be changed easily from a string to an object. However, when i use a `ng-content` and place the for loop in side of `hlm-autocomplete` for content projection i receive an error `No provider for InjectionToken BrnCommandToken!`

```html
<!-- hlm-autocomplete -->
<hlm-command-list>
  <hlm-command-group>
     <ng-content />
  </hlm-command-group>
</hlm-command-list>

<!-- autocomplete-preview -->
<hlm-autocomplete #autocomplete [options]="options">
  @for (option of filteredOptions(); track option) {
     <button hlm-command-item [value]="option" (selected)="autocomplete.optionSelected(option)">
        {{ option }}
    </button>
  }
</hlm-autocomplete>
```

Any suggestions on how to solve this?

marcjulian:
One other styling issue I found is the scrollbar is light in dark mode

CleanShot 2025-08-29 at 11 24 35

@ashley-hunter I saw in Angular Primivites the scroll bar changes based on the theme, but couldn't find the reason for it 😅

ashley-hunter:
> One other styling issue I found is the scrollbar is light in dark mode
> 
> CleanShot 2025-08-29 at 11 24 35
> @ashley-hunter I saw in Angular Primitives the scrollbar changes based on the theme, but couldn't find the reason for it 😅

I believe its to do with the `color-scheme` CSS property being set. In Angular Primitives we set it to light or dark accordingly, but I don't think this happens in the spartan docs

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
ashley-hunter: There is a way to do this. You can create new provider eg. provideBrnCommandHost which you would add to your helm autocomplete component providers.
Now inside the provideBrnCommand function you would optionally inject the BrnCommandHost and if found would register itself there too.
Then you update the injectBrnCommand function to optionally inject BrnCommand and if not found, check if there is a BrnCommandHost and if an instance has been registered with it return it.

I don't know if I have explained that badly - it's not the easiest to describe 😅

I do something similar in the Angular Primitives state library, might be useful to reference. But if your really confused after my poor explanation I can help out when I get some time 😊

marcjulian: Thanks for tipp, I will try it out. I will be on holiday starting next week, so I am not sure if I find time this week.

I am not sure if we should use hlm-command for the autocomplete, because `hlm-command-item` requires the value to be set as string, but we wouldn't use that actually. I am thinking about create brn-autocomplete parts based on brn-command which could support value as `T`. What do you think?

ashley-hunter: Yeah, that might be a good idea. It would solve a lot of these issues. The command was originally a third-party library that we rewrote ourselves while trying to keep the API quite similar to avoid too many breaking changes, but as a result, it does have some limitations like only supporting strings, so I think it'd be nice if this didn't have those kinds of limitations.

hope you enjoy your holiday 😊

marcjulian: I created a secondary entrypoint for `@spartan-ng/brain/autocomplete` and used the command code as base for the brain autocomplete. Autocomplete has a generic so the value doesn't need to be string only.

I couldn't figure out how to do the thing you proposed with provideBrnCommandHost (or provideBrnAutocompleteHost). Not sure if its still necessary to pull out the for-loop for the options, I added `optionTemplate` as input to change the option display content. For this I added an example for countries as an object. I guess we could still pull out `options` and `filter` and only add `filteredOptions`, because I think passing around the options and filter is not necessary. 

I will clean it up after my holiday! Thanks for your feedback already!

libs/brain/autocomplete/src/lib/brn-autocomplete.ts
```diff
@@ -0,0 +1,109 @@
...
+	public readonly keyManager = new ActiveDescendantKeyManager(this.items, this._injector);
+
+	constructor() {
+		this.keyManager
+			.withVerticalOrientation()
+			.withHomeAndEnd()
+			.withWrap()
+			.skipPredicate((item) => item.disabled);
+
+		// TODO change active to first when searchInput changes
```
marcjulian: @ashley-hunter do you have an idea how we can keep the first item active for the following scenarios?

1. Open popover -> first active
2. Clear search -> first active
3. Updating search, items change -> first active

libs/helm/autocomplete/src/lib/hlm-autocomplete-command.ts
```diff
@@ -0,0 +1,30 @@
+import { ChangeDetectionStrategy, Component, computed, input } from '@angular/core';
+import { BrnAutocomplete } from '@spartan-ng/brain/autocomplete';
+import { hlm } from '@spartan-ng/helm/utils';
+import type { ClassValue } from 'clsx';
+
+// TODO rename selector and class - remove comand
+@Component({
+	selector: 'hlm-autocomplete-command',
```
marcjulian: @ashley-hunter do you have suggestions for the naming? This implements `BrnAutocomplete`, but the main component already uses `HlmAutocomplete`

ashley-hunter: Ah, the hardest thing in programming 😂 naming things...

I wonder is this component needed? Could we just use the BrnAutocomplete directive inline in the template rather than using it as a host directive?

libs/brain/autocomplete/src/lib/brn-autocomplete.ts
ashley-hunter: I have not thoroughly tested this, but if we remove the two existing effects handling this and do something like this:

```ts
    effect(() => {
      const items = this.items();

      untracked(() => {
        // if there is no active item or the active item no longer exists then reset to the first item
        if (!this.keyManager.activeItem || !items.includes(this.keyManager.activeItem)) {
          this.keyManager.setFirstItemActive();
        }
      });
    });
```

If this doesn't work let me know and I'll take another look .

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
ashley-hunter: I took some time to look into it myself, I did get it working, but the code was getting very cumbersome, so I think we should avoid it. This approach sounds fine to me 👍

libs/helm/autocomplete/src/lib/hlm-autocomplete-command.ts
marcjulian: will always be the hardest thing 😆 

Probably you are right and the component is not actually needed. No additional styles are applied.

libs/helm/autocomplete/src/lib/hlm-autocomplete.ts
marcjulian: Thanks for trying it out! so far the current approach works fine.

libs/brain/autocomplete/src/lib/brn-autocomplete.ts
marcjulian: Good idea, I also added a third condition to check if the first item in items is not the active item. Maybe you can try it out once more.

marcjulian:
@ashley-hunter thank you for your reviews! I cleaned up the examples and only `filteredOptions` are used as input to display the options. The user need to use `[(search)]` value from the input to filter static options or use it to make async API requests.

apps/app/src/app/pages/(components)/components/(autocomplete)/autocomplete--countries.example.ts
```diff
@@ -0,0 +1,61 @@
...
+	flag: string;
+};
+
+@Component({
+	selector: 'spartan-autocomplete-countries',
+	imports: [HlmAutocomplete],
+	template: `
+		<hlm-autocomplete [filteredOptions]="filteredCountries()" [optionTemplate]="option" [(search)]="search" />
+
+		<ng-template #option let-option>{{ option.flag }} {{ option.name }}</ng-template>
```
marcjulian, line 17: @ashley-hunter `option` variable is here types as `any`, maybe its because the `HlmAutocomplete<any>` is also inferred as any even though `filteredOptions` is from type `Country[]`. Anything we can help angular to infer the correct type?

ashley-hunter: Hmm, so the only way this is potentially possible that I am aware of is rather than passing the option template as an input, would be to have it as a child and a structural directive, something like:

```html
<hlm-autocomplete [filteredOptions]="filteredCountries()" [(search)]="search">
	<span *hlmAutocompleteOption="let option">{{ option.flag }} {{ option.name }}</span>
</hlm-autocomplete>
```

You can see how you can implement type guards here:
https://angular.dev/guide/directives/structural-directives#typing-the-directives-context

That being said, in this case we are trying to infer from a generic on the parent component, so I'm not sure if that kind of inference would be possible unfortunately.

apps/app/src/app/pages/(components)/components/(autocomplete)/autocomplete.page.ts
```diff
@@ -0,0 +1,131 @@
...
+			<spartan-section-sub-heading id="examples">Examples</spartan-section-sub-heading>
+			<h3 id="examples__objects" class="${hlmH4} mb-2 mt-6">Objects</h3>
+			<spartan-tabs firstTab="Preview" secondTab="Code">
+				<div spartanCodePreview firstTab>
+					<spartan-autocomplete-countries />
+				</div>
+				<spartan-code secondTab [code]="_countriesCode()" />
+			</spartan-tabs>
+
+			<h3 id="examples__async" class="${hlmH4} mb-2 mt-6">Asynchronous</h3>
```
ashley-hunter, line 99: The empty text seems a bit off center, it looks like the hidden list element still has some height:

image

Also I notice if I open the dropdown when there is no search term it shows "no options", I wonder in this case should we instead prompt them to search? Not a big deal as it is just an example, but just a thought.

apps/app/src/app/pages/(components)/components/(autocomplete)/autocomplete--countries.example.ts
marcjulian: Thanks for the tipp! I will leave it as is for now. We can explore that later perhaps.

apps/app/src/app/pages/(components)/components/(autocomplete)/autocomplete.page.ts
marcjulian: Well spotted! Autocomplete list will be hidden when options are empty. The popover also doesn't open anymore when there is no search term or options. So for the async example it shouldn't open on keydown arrowDown or the toggle button when the search is empty, but the other examples will open because options are available.

PR #907, opened 2025-10-02
feat: input/textarea polishing

marcjulian:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] autocomplete
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] form-field
- [ ] hover-card
- [ ] icon
- [x] input
- [ ] input-group
- [ ] input-otp
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] sidebar
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [x] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli

- [ ] Yes
- [ ] No

### DIFF
--- a/libs/helm/input/src/lib/hlm-input-error.ts
+++ b/libs/helm/input/src/lib/hlm-input-error.ts
@@ -1,13 +1,6 @@
 import { Directive, computed, input } from '@angular/core';
 import { hlm } from '@spartan-ng/helm/utils';
-import { type VariantProps, cva } from 'class-variance-authority';
 import type { ClassValue } from 'clsx';
-
-export const inputErrorVariants = cva('text-destructive text-sm font-medium', {
-	variants: {},
-	defaultVariants: {},
-});
-export type InputErrorVariants = VariantProps<typeof inputErrorVariants>;
 
 @Directive({
 	selector: '[hlmInputError]',
@@ -17,5 +10,5 @@ export type InputErrorVariants = VariantProps<typeof inputErrorVariants>;
 })
 export class HlmInputError {
 	public readonly userClass = input<ClassValue>('', { alias: 'class' });
-	protected readonly _computedClass = computed(() => hlm(inputErrorVariants(), this.userClass()));
+	protected readonly _computedClass = computed(() => hlm('text-destructive text-sm font-medium', this.userClass()));
 }
--- a/apps/app/src/app/pages/(components)/components/(input)/input--button.preview.ts
+++ b/apps/app/src/app/pages/(components)/components/(input)/input--button.preview.ts
@@ -1,13 +1,13 @@
 import { Component } from '@angular/core';
-import { HlmButton } from '@spartan-ng/helm/button';
-import { HlmInput } from '@spartan-ng/helm/input';
+import { HlmButtonImports } from '@spartan-ng/helm/button';
+import { HlmInputImports } from '@spartan-ng/helm/input';
 
 @Component({
 	selector: 'spartan-input-button',
-	imports: [HlmInput, HlmButton],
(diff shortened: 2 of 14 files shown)


libs/helm/input/src/lib/hlm-input-error.ts
```diff
@@ -1,14 +1,7 @@
...
 import type { ClassValue } from 'clsx';
 
-export const inputErrorVariants = cva('text-destructive text-sm font-medium', {
-	variants: {},
-	defaultVariants: {},
-});
-export type InputErrorVariants = VariantProps<typeof inputErrorVariants>;
-
 @Directive({
 	selector: '[hlmInputError]',
```
marcjulian: I removed the unused error variants and apply the classes directly in `hlm`. I noticed two things

1. There is no example for `hlmInputError` - should I add one?
2. `hlmInputError` has the same styles as `hlm-error` from `form-field` component - do we need both or do we only need one?

MerlinMoos: I think we need both, when we don't have the styling in the input the component is not working correct without the form field. We can think about if we should implement the BrnFormField control in the following components. Autocomplete, select, combobox, datepicker. Im not sure if they have a invalid style but I believe consumers are needing this. What is your opinion?

marcjulian: Why is the input component not working correct without it? It doesn't even use this directive. The input has the error styles for the input field in the `inputVariants`.

For me the styles look like to be used for a error message, but that is already part of form-field with `hlm-error`. I think the purpose is this, but there is not usage of it in the docs.

```html
<div>
  <input hlmInput />
  <div hlmInputError>Email is required</div>
</div>
```

I think we can remove it and `hlm-error` is the way to go.

MerlinMoos: You are right. We can remove them I had something different in my mind, sorry for that.

marcjulian: @thatsamsonkid what do you think about the `hlmInputError` directive as you are currently improving the form field integration in #911 😄

thatsamsonkid: Sorry clicked 'resolved' by accident and also sorry I'm just seeing this until now. I think we don't need it. I made brn directives of both error and hint. So if you need the need the unstyled functionality you would just use that and if you want the styling you'll use the hlm version. I'm going to work on integrating some of the other primitives but i have this working on input and select currently

marcjulian: Thanks for your feedback @thatsamsonkid. I agree, using `HlmError` from Form Field makes more sense, when it also supports more features. I removed the `HlmInputError` directive.

PR #917, opened 2025-10-08
feat(button-group): add new button-group component

dongphuong0905:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] autocomplete
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [x] button-group
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] dropdown-menu
- [ ] form-field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-group
- [ ] input-otp
- [ ] kbd
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] sidebar
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli

Shadcn just present new button-group behavior

Closes #

## What is the new behavior?

This PR will:
- Add new button group component
- Remove shadow from button component for variant: default, destructive and secondary (shadcn PR: https://github.com/shadcn-ui/ui/pull/8252)
- Add two new variants for button icon: icon-sm and icon-lg (shadcn PR: https://github.com/shadcn-ui/ui/pull/8334)

- [ ] Yes
- [x] No

### DIFF
--- a/.github/ISSUE_TEMPLATE/01-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/01-bug-report.yml
@@ -50,6 +50,7 @@ body:
         - input
         - input-group
         - input-otp
+        - kbd
         - label
         - menubar
         - navigation-menu
--- a/.github/ISSUE_TEMPLATE/02-feature-request.yml
+++ b/.github/ISSUE_TEMPLATE/02-feature-request.yml
@@ -41,6 +41,7 @@ body:
         - input
         - input-group
         - input-otp
+        - kbd
         - label
         - menubar
         - navigation-menu
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -54,6 +54,7 @@ What kind of change does this PR introduce?
 - [ ] input
 - [ ] input-group
 - [ ] input-otp
+- [ ] kbd
 - [ ] label
 - [ ] menubar
 - [ ] navigation-menu
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,5 +17,6 @@
 	"editor.defaultFormatter": "esbenp.prettier-vscode",
 	"editor.formatOnPaste": true,
 	"editor.formatOnSave": false,
-	"nxConsole.generateAiAgentRules": true
+	"nxConsole.generateAiAgentRules": true,
+	"tailwindCSS.classFunctions": ["hlm", "cva"]
 }
--- a/apps/app/src/app/core/models/primitives-snippets.model.ts
+++ b/apps/app/src/app/core/models/primitives-snippets.model.ts
@@ -27,6 +27,7 @@ export type PrimitiveSnippet =
 	| 'input'
 	| 'input-group'
 	| 'input-otp'
+	| 'kbd'
 	| 'label'
 	| 'menubar'
 	| 'pagination'
--- a/apps/app/src/app/pages/(components)/components/(autocomplete)/autocomplete--config.example.ts
+++ b/apps/app/src/app/pages/(components)/components/(autocomplete)/autocomplete--config.example.ts
@@ -9,22 +9,20 @@ type Country = {
 };
 
 @Component({
-	selector: 'spartan-autocomplete-countries',
+	selector: 'spartan-autocomplete-config',
 	imports: [HlmAutocompleteImports],
 	template: `
-		<hlm-autocomplete [filteredOptions]="filteredCountries()" [optionTemplate]="option" [(search)]="search" />
-
-		<!-- custom option template with access to the option item -->
-		<ng-template #option let-option>{{ option.flag }} {{ option.name }}</ng-template>
+		<hlm-autocomplete [filteredOptions]="filteredCountries()" [(search)]="search" />
 	`,
 	providers: [
 		provideHlmAutocompleteConfig({
+			transformOptionToString: (option: Country) => `${option.flag} ${option.name}`,
 			transformValueToSearch: (option: Country) => `${option.flag} ${option.name}`,
 		}),
 	],
 	changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class AutocompleteCountries {
+export class AutocompleteConfig {
 	private readonly _countries: Country[] = [
 		{ name: 'Argentina', code: 'AR', flag: '🇦🇷' },
 		{ name: 'Australia', code: 'AU', flag: '🇦🇺' },
--- a/apps/app/src/app/pages/(components)/components/(autocomplete)/autocomplete.page.ts
+++ b/apps/app/src/app/pages/(components)/components/(autocomplete)/autocomplete.page.ts
@@ -1,6 +1,6 @@
 import type { RouteMeta } from '@analogjs/router';
 import { Component, computed, inject } from '@angular/core';
-import { hlmH4, hlmP } from '@spartan-ng/helm/typography';
+import { hlmCode, hlmH4, hlmP, hlmUl } from '@spartan-ng/helm/typography';
 import { Code } from '../../../../shared/code/code';
 import { CodePreview } from '../../../../shared/code/code-preview';
 import { MainSection } from '../../../../shared/layout/main-section';
@@ -17,6 +17,7 @@ import { UIApiDocs } from '@spartan-ng/app/app/shared/layout/ui-docs-section/ui-
 import { Tabs } from '../../../../shared/layout/tabs';
 import { metaWith } from '../../../../shared/meta/meta.util';
 import { AutocompleteAsync } from './autocomplete--async.example';
+import { AutocompleteConfig } from './autocomplete--config.example';
 import { AutocompleteCountries } from './autocomplete--countries.example';
 import { AutocompleteForm } from './autocomplete--form.example';
 import { AutocompletePreview, defaultImports, defaultSkeleton } from './autocomplete.preview';
@@ -45,6 +46,7 @@ export const routeMeta: RouteMeta = {
 		AutocompleteAsync,
 		AutocompleteForm,
 		AutocompleteCountries,
+		AutocompleteConfig,
 	],
 	template: `
 		<section spartanMainSection>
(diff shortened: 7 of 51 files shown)


libs/helm/button/src/lib/hlm-button.ts
```diff
@@ -6,16 +6,16 @@ import type { ClassValue } from 'clsx';
 import { injectBrnButtonConfig } from './hlm-button.token';
 
 export const buttonVariants = cva(
-	'focus-visible:border-ring focus-visible:ring-ring/50 aria-invalid:ring-destructive/20 dark:aria-invalid:ring-destructive/40 aria-invalid:border-destructive inline-flex shrink-0 cursor-default items-center justify-center gap-2 whitespace-nowrap rounded-md text-sm font-medium outline-none transition-all focus-visible:ring-[3px] data-[disabled]:pointer-events-none data-[disabled]:opacity-50 [&_ng-icon]:pointer-events-none [&_ng-icon]:shrink-0',
```
ajitzero, line 9: I think cursor-default is required but I don't remember the context. shadcn also added this at some point. We should revert this from the global HlmButton and only add to the Button group, if needed.

apps/app/src/app/pages/(components)/components/(button)/button.page.ts
```diff
@@ -149,7 +149,7 @@ export const routeMeta: RouteMeta = {
 			</spartan-tabs>
 
 			<spartan-page-bottom-nav>
-				<spartan-page-bottom-nav-link href="calendar" label="Calendar" />
+				<spartan-page-bottom-nav-link href="button-group" label="Button Group" />
 				<spartan-page-bottom-nav-link direction="previous" href="breadcrumb" label="Breadcrumb" />
```
ajitzero, line 153: The previous link should be updated for the Calendar page as well.

apps/app/src/app/pages/(components)/components/(button-group)/button-group--dropdown-menu.example.ts
```diff
@@ -0,0 +1,90 @@
...
+import { HlmButtonGroup } from '@spartan-ng/helm/button-group';
+import { HlmIcon } from '@spartan-ng/helm/icon';
+import { HlmMenu, HlmMenuGroup, HlmMenuItem, HlmMenuSeparator } from '@spartan-ng/helm/menu';
+
+@Component({
+	selector: 'spartan-button-group-dropdown-menu',
+	imports: [
+		HlmIcon,
+		NgIcon,
+		HlmButton,
```
MerlinMoos: Hey can you please add all imports as HlmButtonImports and HlmMenuImports and not as a single import :)

libs/helm/button-group/src/lib/hlm-button-group.ts
MerlinMoos, line 1: can we please create a brain for that? then we can handle the inputs and the attr bindings inside of the brain and use it as a host directive

libs/helm/button-group/src/lib/hlm-button-group-separator.ts
MerlinMoos, line 1: Can you please convert that to a directive a component is not needed for that and use the BrnSeparator as a host directive. Check the HlmSeperator as a example

libs/helm/button-group/src/lib/hlm-button-group-text.ts
```diff
@@ -0,0 +1,19 @@
...
+		'[class]': '_computedClass()',
+	},
+})
+export class HlmButtonGroupText {
+	public readonly userClass = input<ClassValue>('', { alias: 'class' });
+
+	protected readonly _computedClass = () =>
+		hlm(
+			"bg-muted shadow-xs flex items-center gap-2 rounded-md border px-4 text-sm font-medium [&_svg:not([class*='size-'])]:size-4 [&_svg]:pointer-events-none",
+			this.userClass(),
```
MerlinMoos, line 17: here we should use the ng-icon instead of the svg in the classes

libs/helm/button/src/lib/hlm-button.ts
dongphuong0905: Reverted

apps/app/src/app/pages/(components)/components/(button)/button.page.ts
dongphuong0905: Updated. Thanks for your good catch <3

apps/app/src/app/pages/(components)/components/(button-group)/button-group--dropdown-menu.example.ts
dongphuong0905: Updated. Thanks a lot ❤️

libs/helm/button-group/src/lib/hlm-button-group-text.ts
dongphuong0905: Updated

libs/helm/button-group/src/lib/hlm-button-group-separator.ts
dongphuong0905: I created a separated component because I want to set the default value of `orientation` is `vertical` instead of `horizontal`. Do you have any better idea to do it? 🤔

MerlinMoos: You can inject the BrnSeperator when you have it as a host directive and set the input then in the constructor. Maybe you need to convert the input to a model :)

MerlinMoos: I looked into the separator. Please add a linked signal and apply the orientation to it. Then provide a public method where you can change the orientation. Then we don't get a output :)

apps/app/src/app/pages/(components)/components/(button-group)/button-group--dropdown-menu.example.ts
MerlinMoos: Can you please do it with all hlm imports? This would be awesome over all components

libs/helm/button-group/src/lib/hlm-button-group.ts
marcjulian: I am not sure if its necessary to have a brain directive just for a couple of attr bindings. In shadcn/ui its also just the "helm" part, its not using any brain like radix-ui component.

libs/helm/button-group/src/lib/hlm-button-group.ts
```diff
@@ -0,0 +1,38 @@
...
+		},
+	},
+);
+
+@Directive({
+	selector: '[hlmButtonGroup]',
+	host: {
+		'[class]': '_computedClass()',
+		'[attr.data-orientation]': 'orientation()',
+		'[attr.data-slot]': '"button-group"',
```
marcjulian: You can apply data-slot directly as the value is static 

```suggestion
		'data-slot': 'button-group',
```

libs/helm/button-group/src/lib/hlm-button-group-separator.ts
marcjulian: What about adding a config function `provideBrnSeparatorConfig` to the brain separator package to allow changing the default orientation for the separator directive? Than you can create specific `HlmButtonGroupSeparator` as directive with `BrnSeparator` as host directive and override the default orientation.

```ts
// brn-separator.token.ts
import { InjectionToken, type ValueProvider, inject } from '@angular/core';

export type SeparatorOrientation = 'horizontal' | 'vertical';

export interface BrnSeparatorConfig {
	orientation: SeparatorOrientation;
}

const defaultConfig: BrnSeparatorConfig = {
	orientation: 'horizontal',
};

const BrnSeparatorConfigToken = new InjectionToken<BrnSeparatorConfig>('BrnSeparatorConfig');

export function provideBrnSeparatorConfig(config: Partial<BrnSeparatorConfig>): ValueProvider {
	return { provide: BrnSeparatorConfigToken, useValue: { ...defaultConfig, ...config } };
}

export function injectBrnSeparatorConfig(): BrnSeparatorConfig {
	return inject(BrnSeparatorConfigToken, { optional: true }) ?? defaultConfig;
}
```

dongphuong0905: Hi @marcjulian, yeah I prefer your suggestion since it present a clean way to override `orientation` of `Separator`
Thanks for your suggestion

libs/helm/button-group/src/lib/hlm-button-group.ts
dongphuong0905: Fixed. Thanks a lot ❤️

apps/app/src/app/pages/(components)/components/(button-group)/button-group--dropdown-menu.example.ts
dongphuong0905: I think we might not need to use `HlmButtonGroupImports`, because some components only require a single one — for example, `HlmButtonGroup`

libs/helm/button-group/src/lib/hlm-button-group-separator.ts
```diff
@@ -0,0 +1,25 @@
...
+import { hlm } from '@spartan-ng/helm/utils';
+import type { ClassValue } from 'clsx';
+
+@Directive({
+	selector: '[hlmButtonGroupSeparator],hlm-button-group-separator',
+	hostDirectives: [{ directive: BrnSeparator, inputs: ['orientation', 'decorative'] }],
+	providers: [provideBrnSeparatorConfig({ orientation: 'vertical' })],
+	host: {
+		'data-slot': 'button-group-separator',
+		role: 'none',
```
marcjulian: You can remove the role, this is already handled by `BrnSeparator`

libs/helm/button-group/src/lib/hlm-button-group-separator.ts
```diff
@@ -0,0 +1,25 @@
...
+		'[class]': '_computedClass()',
+	},
+})
+export class HlmButtonGroupSeparator {
+	public readonly userClass = input<ClassValue>('', { alias: 'class' });
+
+	protected readonly _computedClass = computed(() =>
+		hlm(
+			'bg-input relative !m-0 inline-flex shrink-0 self-stretch data-[orientation=horizontal]:h-px data-[orientation=vertical]:h-auto data-[orientation=horizontal]:w-full data-[orientation=vertical]:w-px',
```
marcjulian: Why is there `!m-0` in the styles? I saw it in the shadcn/ui version as well, but I am just confused about important and the class itself. Looks fine to me without it the class at all, just confused why it's set to important.

dongphuong0905: Hmm yeah, I saw the separator is still fine without it. Do you think we should remove it?

libs/helm/button-group/src/lib/hlm-button-group-separator.ts
dongphuong0905: Removed

libs/helm/button-group/src/lib/hlm-button-group-separator.ts
marcjulian: Seems it's unnecessary and we should remove it, Tailwind will just generate a class which is not used. If we stumble upon the actual use case, we can add it back to the class list.

dongphuong0905: @marcjulian I removed it. Thanks for your suggestion

marcjulian: Perfect thank your!

PR #926, opened 2025-10-12
feat: warn about modifications in hlm library migrator

drdreo:
## PR Checklist

- [x] The commit message follows the [contribution guidelines](https://github.com/spartan-ng/spartan/blob/main/CONTRIBUTING.md#-commit-message-guidelines)
- [ ] Docs have been added / updated

## PR Type

- [x] Feature
- [x] Refactoring (no API changes)

- [x] CLI `migrate-helm-libraries` generator
- [x] CLI `ui` generator

The Helm library migrator removes and regenerates selected local libraries. Although it displays a generic confirmation, it cannot tell users which libraries contain local modifications or which files would be lost.

## What is the new behavior?

The CLI records hashes for newly generated or successfully migrated Helm libraries in `.spartan/library-metadata.json`. On later migrations it:

- categorizes libraries as unchanged or customized;
- reports added, modified, and deleted files;
- lets users select all unchanged/customized libraries or individual libraries;
- requires explicit confirmation before overwriting customized libraries;
- refreshes metadata only after a successful, formatted migration;
- adds `.spartan/` to `.gitignore` automatically.

Existing libraries without baseline metadata are treated as customized, which is the safe default: the migrator warns rather than silently overwriting them. Running the UI generator for an already installed library does not reset its baseline or hide existing customizations.

Example output:

Migration customization warning

- [x] Yes

The first migration of a legacy workspace without metadata reports its existing libraries as customized. Once a library is regenerated successfully, future changes are tracked against that baseline.

## Summary by CodeRabbit

* **New Features**
  * Added Helm library customization detection using persisted metadata to identify added, modified, and deleted files.
  * Migration UI now distinguishes unchanged vs customized libraries, shows per-library customization details, and adds bulk selection options (all/unchanged/customized).
  * Warns users and requests confirmation before overwriting customized libraries.
  * Newly created/queued migrations now automatically backfill customization metadata.

* **Bug Fixes**
  * Prevents overwriting customized libraries when overwrite confirmation is declined.

* **Tests**
  * Added comprehensive tests for metadata handling, customization detection, and migration/confirmation flows.


MerlinMoos:
We have currenlty internal discussions about it. We need to revamp the tooltip component then the better form support for all components and we want to get rid of the brn imports for the components. But we are close to getting there. I don't had time the last 2 months but hopefully i find time soon to support here more.

drdreo:
So are we ready to add this now since v1 was released? I would fix the merge conflicts 🥳

Issue #883, opened 2025-09-20
collapsible: content applies `hidden` attribute too early making it difficult to animate


OmerGronich:
### Which scope/s are relevant/related to the feature request?

collapsible

### Information

The current BrnCollapsibleContent implementation applies the hidden attribute immediately on state change:

https://github.com/spartan-ng/spartan/blob/e12d61cd42efe46801055d651f41038c0237a0ea/libs/brain/collapsible/src/lib/brn-collapsible-content.ts#L19

Because hidden is applied instantly, the content cannot be animated when collapsing.

Expected Behavior:
The hidden attribute should only be applied after the exit animation completes, similar to how libraries like Radix React and Radix NG handle it:

- [Radix NG](https://github.com/radix-ng/primitives/blob/977d4ab2800a847ac190ec1e44e66965dc162e91/packages/primitives/collapsible/src/collapsible-content.directive.ts)
- [Radix React Presence](https://github.com/radix-ui/primitives/blob/main/packages/react/presence/src/presence.tsx)

### I would be willing to submit a PR to fix this issue

- [ ] Yes
- [x] No

marcjulian:
Resolved with #976, please update to `0.0.1-alpha.556` for brain and cli than regenerate collapsible helm. You can find an example on the collapsible docs https://spartan.ng/components/collapsible

PR #976, opened 2025-10-30
feat: collapsible polishing

marcjulian:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] autocomplete
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] button-group
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [x] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] empty
- [ ] dropdown-menu
- [ ] form-field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-group
- [ ] input-otp
- [ ] item
- [ ] kbd
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] resizable
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] sidebar
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli

## What is the new behavior?

- refactor brn-collapsible components to directives
- add hlm-collapsible and use brn-collapsible-* as host directive
- add data-slot to hlm-collapsible-*
- compute state based on expanded property
- use disabled when toggling
- calculate content size with afterNextRender
- add button type to brn-collapsible-trigger
- add migration from brn-collapsible-* to hlm-collapsible-*

- [ ] Yes
- [x] No

Current usage of brn-collapsible will continue to work, because the directives provide all the logic.

### DIFF
--- a/libs/brain/collapsible/src/lib/brn-collapsible-content.ts
+++ b/libs/brain/collapsible/src/lib/brn-collapsible-content.ts
@@ -1,41 +1,29 @@
-import {
-	Directive,
-	ElementRef,
-	OnDestroy,
-	afterNextRender,
-	effect,
-	inject,
-	input,
-	signal,
-	untracked,
-} from '@angular/core';
-import { injectBrnCollapsible } from './brn-collapsible-token';
+import { Directive, ElementRef, afterNextRender, effect, inject, input, signal, untracked } from '@angular/core';
+import { measureDimensions } from '@spartan-ng/brain/core';
+import { injectBrnCollapsible, injectBrnCollapsibleConfig } from './brn-collapsible-token';
 
 @Directive({
 	selector: '[brnCollapsibleContent],brn-collapsible-content',
 	host: {
-		'[hidden]': '_shouldHide()',
+		'[attr.inert]': "_collapsible?.state() === 'closed' ? true : undefined",
 		'[attr.data-state]': '_collapsible?.state()',
 		'[id]': '_collapsible?.contentId()',
 		'[style.--brn-collapsible-content-width.px]': '_width()',
 		'[style.--brn-collapsible-content-height.px]': '_height()',
 	},
 })
-export class BrnCollapsibleContent implements OnDestroy {
-	protected readonly _collapsible = injectBrnCollapsible();
+export class BrnCollapsibleContent {
+	private readonly _config = injectBrnCollapsibleConfig();
 	private readonly _elementRef = inject<ElementRef>(ElementRef);
-	/**
-	 * The id of the collapsible content element.
-	 */
-	public readonly id = input<string | null | undefined>();
-
-	protected readonly _shouldHide = signal<boolean>(true);
+	protected readonly _collapsible = injectBrnCollapsible();
 
 	protected readonly _width = signal<number | null>(null);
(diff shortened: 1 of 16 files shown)


goetzrobin:
> @goetzrobin the healthcheck unit test is failing because it is expecting `BrnCollapsibleImports` but with the new generator it is transformed to `HlmCollapsibleImports`. Not sure how to fix the test though.

do we need to make the healthcheck exclude .spec files maybe?

libs/brain/collapsible/src/lib/brn-collapsible-content.ts
```diff
@@ -1,40 +1,24 @@
...
-import { BrnCollapsible } from './brn-collapsible';
+import { Directive, ElementRef, afterNextRender, effect, inject, input, signal, untracked } from '@angular/core';
+import { injectBrnCollapsible } from './brn-collapsible-token';
 
-@Component({
-	selector: 'brn-collapsible-content',
+@Directive({
+	selector: '[brnCollapsibleContent],brn-collapsible-content',
 	host: {
 		'[hidden]': '!_collapsible?.expanded()',
```
OmerGronich: since you're already visiting this component, i suggest tackling this issue: 

https://github.com/spartan-ng/spartan/issues/883

it makes the collapsible primitive impossible to animate properly

marcjulian: Do you have a solution which would work? I haven't looked into it yet, maybe it could be a follow up PR to add this behavior.

goetzrobin: @marcjulian the way we have done this is just add a manual delay from the collapsible to be marked as !expanded() to actually applying the hidden attribute. Something like this:
```ts
	/** The default delay in ms before hiding the tooltip after hide is called */

	public readonly hideDelay = input(this._defaultOptions?.hideDelay ?? 0, { transform: numberAttribute });
```

It's not pretty but works for now until we can leverage the new animation functionality that was recently added

goetzrobin:
@marcjulian looks great! Do you think we should take care of the hidden delay in this PR or separately? Was my explanation helpful or is it not clear what needs to happen?

goetzrobin:
@marcjulian how did you work around this limitation: https://github.com/angular/angular/issues/50510
Should we expose linkedSignals for the delays so one can set it in the constructor?

Show delay is definitely kind of a niche thing. because one wants animate-in to run immediately. not sure if we should keep it

PR #1000, opened 2025-11-08
feat: update home screen with examples

dongphuong0905:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] autocomplete
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] button-group
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] empty
- [ ] dropdown-menu
- [ ] field
- [ ] form-field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-group
- [ ] input-otp
- [ ] item
- [ ] kbd
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] resizable
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] sidebar
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [x] repo
- [ ] cli

## What is the new behavior?
- Remove the examples pages.
- Move examples to home page

Preview:
https://github.com/user-attachments/assets/8eefced1-ce73-4340-b70e-79f849b353f0

- [ ] Yes
- [x] No


apps/app/src/app/pages/(home).page.ts
```diff
@@ -105,39 +85,34 @@ const lead = 'text-foreground max-w-3xl text-base text-balance sm:text-lg';
...
-				<div class="flex flex-col gap-6 *:[div]:w-full *:[div]:max-w-full">
-					<div class="w-full max-w-md rounded-lg border p-6">
-						<spartan-field-preview />
-					</div>
+		<section class="container-wrapper px-6">
+			<hlm-tabs [tab]="_activeTab()" class="w-full">
+				<hlm-tabs-list
+					aria-label="tabs example"
+					class="bg-background dark:[&>a]:bg-background [&>a]:text-muted-foreground [&>a]:data-[state=active]:text-primary [&>a]:hover:text-primary dark:[&>a]:data-[state=active]:bg-background [&>a]:cursor-pointer [&>a]:data-[state=active]:shadow-none dark:[&>a]:data-[state=active]:border-none"
+				>
```
MerlinMoos, line 93: hey here you can replace the links with 

<a hlmTabsTrigger="examples" routerLink="/" [relativeTo]="_route">Examples</a>
					<a hlmTabsTrigger="dashboard" [routerLink]="[]" fragment="dashboard" [relativeTo]="_route">Dashboard</a>
					<a hlmTabsTrigger="tasks" [routerLink]="[]" fragment="tasks" [relativeTo]="_route">Tasks</a>
					<a hlmTabsTrigger="playground" [routerLink]="[]" fragment="playground" [relativeTo]="_route">Playground</a>
					<a hlmTabsTrigger="authentication" [routerLink]="[]" fragment="authentication" [relativeTo]="_route">
						Authentication
					</a>

apps/app/src/app/pages/(home).page.ts
```diff
@@ -160,4 +135,7 @@ const lead = 'text-foreground max-w-3xl text-base text-balance sm:text-lg';
 		</section>
 	`,
 })
-export default class HomePage {}
+export default class HomePage {
+	private readonly _route = inject(ActivatedRoute);
```
MerlinMoos:
```suggestion
	protected readonly _route = inject(ActivatedRoute);
```

MerlinMoos (changes requested):
Hey, I added a suggestion. For that you need to make the activedRoute protected that you have access in the template.

apps/app/src/app/pages/(home).page.ts
marcjulian: You can set `routerLink="."` this targets the current route and than add the fragment. relativeTo is not neede when you use `.` in routerLink

MerlinMoos: That I did not know, nice thanks

dongphuong0905: Updated. Thanks a lot <3

ashley-hunter:
> > > > Oh where is that coming from... I need to have a deeper look into that. Maybe someone is having an idea?
> > > > @marcjulian @ashley-hunter
> > > 
> > > 
> > > It's kind of an Angular issue as far as I can tell. Angular adds all the classes one by one, but the problem is the transition-all class gets added before some of the other visual style classes so anything that comes after the transition-all class gets animated.
> > > If you manually modify the button CSS classes that we apply so that transition-all is the very last class this issue goes away, but it kind of sucks that we have to do that, and technically we would have to do this for every component that uses transitions to eliminate this issue permanently. I guess we have the `hlm` function and we could technically re-order the classes in there to make it so that any transition classes are always at the end?
> > 
> > 
> > I have noticed this flash of styles in a button and thought it might be an issue with SSR, but maybe it's related to `transition-all`. One solution would be to order ` transition-*` in the `hlm` function to be applied last? Do we want to do that in this PR or do we do it in a separate one? I guess we don't need to hold off on merging this PR when we can solve the style flash issue in a separate one.
> 
> how should we handle direct class binding without our hlm function call? Is there a possibilty to add a custom order in the prettier package?

I don't think that the tailwind prettier plugin supports this, but this is only a problem for class bindings, not when classes are a static attribute, as they get compiled differently - bindings use the classMap ivy instruction which causes this issue whereas static classes are just consts, so at least in most other places I don't think this should be an issue? Any places where we are using class bindings should maybe also use the hlm function.

goetzrobin:
@marcjulian 
the code for hlm is:
```ts
import { type ClassValue, clsx } from 'clsx';
import { twMerge } from 'tailwind-merge';

export function hlm(...inputs: ClassValue[]) {
	return twMerge(clsx(inputs));
}
```
I doubt clsx does any ordering. twMerge might?

ashley-hunter:
@marcjulian yes it is the variants that get applied after that is causing the problem, so hlm would need to compute the entire class list based on variants etc.. then apply the sorting, so transitions always come after regardless of variants, so we shouldn't need to change any hlm calls to extract transition, we just need to determine the complete class list and then sort.

PR #1053, opened 2025-11-30
feat: better form control integration

### LINKED ISSUE #591: Formfield hint and error are not semantically associated with input

When using a screenreader Hint or Error Messages are not mentioned when focusing on input

input

### Description

Hint and Error directives are not semantically linked to the field to which they are helping describe. They are missing their own unique id's and then the input is missing an aria-describedby with the listed hint and error id's. They need to be swapped based on input state as well.

```true

```

### Other information

_No response_

- [x] Yes
- [ ] No

thatsamsonkid:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [x] autocomplete
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] button-group
- [ ] calendar
- [ ] card
- [ ] carousel
- [x] checkbox
- [ ] collapsible
- [x] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [x] date-picker
- [ ] dialog
- [ ] empty
- [ ] dropdown-menu
- [x] field
- [x] form-field
- [ ] hover-card
- [ ] icon
- [x] input
- [x] input-group
- [ ] input-otp
- [ ] item
- [ ] kbd
- [x] label
- [ ] menubar
- [ ] navigation-menu
- [x] native-select
- [ ] pagination
- [ ] popover
- [ ] progress
- [x] radio-group
- [ ] resizable
- [ ] scroll-area
- [x] select
- [ ] separator
- [ ] sheet
- [ ] sidebar
- [ ] skeleton
- [x] slider
- [ ] sonner
- [ ] spinner
- [x] switch
- [ ] table
- [ ] tabs
- [x] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli

Closes #591 

- [x] Yes
- [ ] No

Removes the form-field package and replaces it with field.

### DIFF
--- a/libs/brain/forms/src/lib/error-state-tracker.ts
+++ b/libs/brain/forms/src/lib/error-state-tracker.ts
@@ -2,7 +2,7 @@ import { computed } from '@angular/core';
 import { toSignal } from '@angular/core/rxjs-interop';
 import type { AbstractControl, FormGroupDirective, NgControl, NgForm } from '@angular/forms';
 import { BehaviorSubject, of } from 'rxjs';
-import { map, startWith, switchMap } from 'rxjs/operators';
+import { catchError, map, startWith, switchMap } from 'rxjs/operators';
 import type { ControlState } from './control-state';
 import type { ErrorStateMatcher } from './error-options';
 
@@ -14,11 +14,24 @@ export class ErrorStateTracker {
 			switchMap((ngControl) => {
 				if (!ngControl) return of(null);
 
-				const control = ngControl.control as AbstractControl;
+				const control = ngControl.control as AbstractControl | null | undefined;
+				if (!control) {
+					throw new Error(
+						`ErrorStateTracker: ngControl.control is null or undefined. This usually means the form control directive (e.g. formControlName or ngModel) ` +
+							`found an NgControl in the DI tree but no ControlValueAccessor was registered to back it. ` +
+							`Ensure the host component or directive provides NG_VALUE_ACCESSOR (e.g. { provide: NG_VALUE_ACCESSOR, useExisting: forwardRef(() => YourComponent), multi: true }) ` +
+							`and implements the ControlValueAccessor interface.`,
+					);
+				}
+
 				return control.events.pipe(
 					startWith(() => this._getState(control)),
 					map(() => this._getState(control)),
 				);
+			}),
+			catchError((error) => {
+				console.error(error);
+				return of(null);
 			}),
 		),
 		{
--- a/apps/ui-storybook/stories/select.stories.ts
+++ b/apps/ui-storybook/stories/select.stories.ts
@@ -4,6 +4,7 @@ import { FormControl, FormGroup, FormsModule, ReactiveFormsModule, Validators } 
 import { NgIcon, provideIcons } from '@ng-icons/core';
 import { lucideCitrus } from '@ng-icons/lucide';
 import { HlmButton } from '@spartan-ng/helm/button';
+import { HlmFieldImports } from '@spartan-ng/helm/field';
 import { HlmLabel } from '@spartan-ng/helm/label';
 import { HlmSelectImports } from '@spartan-ng/helm/select';
 import { type Meta, type StoryObj, argsToTemplate, moduleMetadata } from '@storybook/angular';
@@ -23,7 +24,16 @@ const meta: Meta<BrnSelectStoryArgs> = {
 	},
 	decorators: [
 		moduleMetadata({
-			imports: [CommonModule, FormsModule, ReactiveFormsModule, HlmSelectImports, HlmLabel, NgIcon],
+			imports: [
+				CommonModule,
+				FormsModule,
+				ReactiveFormsModule,
+				HlmSelectImports,
+				HlmFieldImports,
+				HlmButton,
+				HlmLabel,
+				NgIcon,
+			],
 			providers: [provideIcons({ lucideCitrus })],
 		}),
 	],


thatsamsonkid:
Yea I forgot about it for a moment since we don't have any primitives for it. I think we might have to create one in order to do the form integration however

thatsamsonkid:
Have 2 weird test failures with autocomplete and select not sure what they are about. But will work on combobox in mean time and come back to them. Sorry for taking this long on this I know this one of the main things block V1. But almost there I think!

libs/helm/date-picker/src/lib/hlm-date-picker.ts
```diff
@@ -33,8 +38,15 @@ let nextId = 0;
 
 @Component({
 	selector: 'hlm-date-picker',
-	imports: [HlmIconImports, BrnPopoverImports, HlmPopoverImports, HlmCalendar],
-	providers: [HLM_DATE_PICKER_VALUE_ACCESSOR, provideIcons({ lucideChevronDown })],
+	imports: [NgIcon, HlmIcon, HlmFieldControlDescribedBy, HlmPopoverImports, HlmCalendar],
+	providers: [
+		// HLM_DATE_PICKER_VALUE_ACCESSOR,
```
thatsamsonkid: I just need to verify having this removed didn't break the formcontrol, was having a circular dependency with the BrnFormFieldControl. From what I saw I didn't see that so we should be good

thatsamsonkid:
Moved to ready since this very close to being final. A few things however

- [ ] about 3 test failures I need to check
- [ ] verify datepicker
- [ ] verify visually all error states are correct
- [ ] Need to switching between hint and error

Let me know if anything else I should I check

thatsamsonkid:
Fixed everything prior to rebasing, did a quick conflict resolution but need to verify the rebased changes with the classes approach change, moved back to draft for a moment.

MerlinMoos:
Nice work thanks for that. Im just wondering why the combobox and the datepicker are not having a border 

image


image


on a error state like the input

image


also the date multipicker and date range picker is missing for this feature

thatsamsonkid:
Let me try to check this tonight

thatsamsonkid:
Sorry was checking it but didn't get to finish. I should have some time this weekend to wrap it up. Appreciate the patience and for reviewing also!

libs/brain/field/src/lib/brn-field.ts
```diff
@@ -0,0 +1,24 @@
...
+	public readonly dataInvalid = input<boolean, BooleanInput>(false, {
+		transform: booleanAttribute,
+		alias: 'data-invalid',
+	});
+
+	protected readonly _invalid = computed(() => {
+		if (this.dataInvalid()) return true;
+		const control = this._brnFieldControl();
+		if (!control || !control.ngControl) return false;
+		return (control.errorState() || control.ngControl.invalid) && control.ngControl.touched;
```
toniskobic: I think the `control.errorState` evaluation should be sufficient because `invalid` and `touched` are evaluated inside the error tracker.



More importantly, if `ShowOnDirtyErrorStateMatcher` is used, it will not work properly because of the `control.ngControl.touched`.

https://github.com/spartan-ng/spartan/blob/f82e4b3ecc0094fda10ff3707774d17bb47484e2/libs/brain/forms/src/lib/error-options.ts#L4-L10

```suggestion
		return control.errorState();
```

libs/brain/slider/src/lib/brn-slider.ts
```diff
@@ -41,13 +54,21 @@ import { provideBrnSlider } from './brn-slider.token';
 		'[attr.data-orientation]': 'orientation()',
 		'data-slot': 'slider',
 		'(focusout)': '_onFocusOut($event)',
+		'[attr.aria-invalid]': 'errorState() ? "true" : null',
+		'[attr.data-invalid]': 'errorState() ? "true" : null',
 	},
 })
-export class BrnSlider implements ControlValueAccessor, OnInit {
+export class BrnSlider implements ControlValueAccessor, OnInit, DoCheck {
```
toniskobic:
```suggestion
export class BrnSlider implements ControlValueAccessor, OnInit, DoCheck, BrnFieldControl {
```

libs/brain/slider/src/lib/brn-slider.ts
```diff
@@ -41,13 +54,21 @@ import { provideBrnSlider } from './brn-slider.token';
...
 	private readonly _dir = inject(Directionality);
 	private readonly _injector = inject(Injector);
 	private _ngControl: NgControl | null = null;
```
toniskobic: Slider did not implement the BrnFieldControl (`ngControl` property missing).

```suggestion
	public ngControl: NgControl | null = null;
```

libs/helm/slider/src/lib/hlm-slider.ts
```diff
@@ -68,10 +68,17 @@ import { classes } from '@spartan-ng/helm/utils';
 export class HlmSlider {
 	protected readonly _slider = injectBrnSlider();
 
+	protected readonly _errorStateClass = computed(() =>
+		this._slider.errorState()
+			? 'ring-ring data-[invalid=true]:ring-destructive focus-visible:ring-destructive/20 dark:focus-visible:ring-destructive/40'
+			: '',
+	);
```
toniskobic: These styles seem to be not needed because the slider does not have a border and these styles are not visible.

shadcn just colors the field label in red.
```suggestion
```

libs/helm/slider/src/lib/hlm-slider.ts
```diff
@@ -68,10 +68,17 @@ import { classes } from '@spartan-ng/helm/utils';
...
+
 	constructor() {
-		classes(
-			() =>
-				'group flex w-full touch-none flex-col select-none data-vertical:h-full data-vertical:min-h-40 data-vertical:w-auto data-vertical:flex-row data-[disabled]:pointer-events-none data-[disabled]:opacity-50',
-		);
+		classes(() => [
+			'group flex w-full touch-none flex-col select-none data-vertical:h-full data-vertical:min-h-40 data-vertical:w-auto data-vertical:flex-row data-[disabled]:pointer-events-none data-[disabled]:opacity-50',
+			this._slider.mutableDisabled() ? 'opacity-40' : '',
```
toniskobic: `data-disabled` Tailwind class is already applied, no need for this.

```suggestion
```

libs/brain/radio-group/src/lib/brn-radio-group.ts
```diff
@@ -81,12 +95,46 @@ export class BrnRadioGroup<T = unknown> implements ControlValueAccessor {
...
+		effect(() => {
+			const error = this._errorStateTracker.errorState();
+			untracked(() => {
+				if (this.ngControl) {
+					const shouldShowError = error && this.ngControl.invalid && (this.ngControl.touched || this.ngControl.dirty);
+					this._errorStateTracker.errorState.set(shouldShowError ? true : false);
+				}
+			});
+		});
+	}
```
toniskobic: Isn't this a circular signal dependency? On errorState change, we set the errorState again.

```suggestion
```

libs/helm/date-picker/src/lib/hlm-date-picker-multi.ts
```diff
@@ -186,4 +226,27 @@ export class HlmDatePickerMulti<T> implements ControlValueAccessor {
...
+		effect(() => {
+			const error = this._errorStateTracker.errorState();
+			untracked(() => {
+				if (this._ngControl) {
+					const shouldShowError =
+						error && this._ngControl.invalid && (this._ngControl.touched || this._ngControl.dirty);
+					this._errorStateTracker.errorState.set(shouldShowError ? true : false);
+				}
+			});
+		});
```
toniskobic: Is this also circular dependency? Also, `ErrorStateMatcher` decides if the error should be shown on touched or dirty. `updateErrorState` should only set the `errorState` signal.

libs/helm/date-picker/src/lib/hlm-date-picker.ts
```diff
@@ -163,6 +204,29 @@ export class HlmDatePicker<T> implements ControlValueAccessor {
...
+		effect(() => {
+			const error = this._errorStateTracker.errorState();
+			untracked(() => {
+				if (this._ngControl) {
+					const shouldShowError =
+						error && this._ngControl.invalid && (this._ngControl.touched || this._ngControl.dirty);
+					this._errorStateTracker.errorState.set(shouldShowError ? true : false);
+				}
+			});
+		});
```
toniskobic: I am commenting again on a code block like this, maybe it is my fault and I am not understanding something.


```suggestion
```

libs/helm/date-picker/src/lib/hlm-date-range-picker.ts
```diff
@@ -204,4 +239,27 @@ export class HlmDateRangePicker<T> implements ControlValueAccessor {
...
+		effect(() => {
+			const error = this._errorStateTracker.errorState();
+			untracked(() => {
+				if (this._ngControl) {
+					const shouldShowError =
+						error && this._ngControl.invalid && (this._ngControl.touched || this._ngControl.dirty);
+					this._errorStateTracker.errorState.set(shouldShowError ? true : false);
+				}
+			});
+		});
```
toniskobic: Everything seems to work fine when removing this effect.
```suggestion
```

libs/brain/field/src/lib/brn-field-aria.service.ts
```diff
@@ -0,0 +1,29 @@
+import { computed, Injectable, signal } from '@angular/core';
+
+@Injectable()
+export class HlmFieldA11yService {
+	private readonly _descriptions = signal<string[]>([]);
+	private readonly _errors = signal<string[]>([]);
```
toniskobic, line 6: This is fine, but ngxtension's [SignalSet](https://ngxtension.dev/utilities/collections/signal-set/) would be great for this use case and would make this service more concise.

libs/helm/form-field/src/lib/hlm-form-field.ts
```diff
@@ -32,7 +32,7 @@ export class HlmFormField {
 		classes(() => 'block space-y-2');
 		effect(() => {
 			if (!this.control()) {
-				throw new Error('hlm-form-field must contain a BrnFormFieldControl.');
+				throw new Error('hlm-form-field must contain a BrnFieldControl.');
```
toniskobic:
```suggestion
				throw new Error('hlm-field must contain a BrnFieldControl.');
```

libs/helm/input/src/lib/hlm-input.ts
toniskobic: Commenting on this king of effect again. Would like to receive an explanation of its function.

libs/helm/slider/src/lib/__tests__/hlm-slider-form.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+				<hlm-slider formControlName="value"></hlm-slider>
+				<p hlmFieldDescription>Set the volume level.</p>
+				<hlm-field-error>Adjust the slider before continuing.</hlm-field-error>
+			</div>
+		</form>
+	`,
+})
+class HlmSliderHost {
+	public readonly form = new FormGroup({
+		value: new FormControl(0, { validators: [Validators.min(5), Validators.required] }),
```
toniskobic: Value needs to be an array.

```suggestion
		value: new FormControl([0], { validators: [Validators.min(5), Validators.required] }),
```

toniskobic (changes requested):
Some field examples have wrong label to input mappings, `for` is referring to the wrong `id`, and in this instance `hlm-slider` does not have an `id` at all.

https://github.com/spartan-ng/spartan/blob/f82e4b3ecc0094fda10ff3707774d17bb47484e2/apps/app/src/app/pages/(components)/components/(field)/field--slider.preview.ts#L15-L17

toniskobic:
> what is missing is the style issue from the autocomplete the ring should be descrutrive
> 
> image
> also we need to veryfy all control components
> 
> I want to kick out also the hlm-form-field
> 
> this is not needed anymore

@MerlinMoos This happens because the focus-visible styles have precedence 0,3,0 , while the invalid classes have 0,2,0 precedence. Also, more importantly, we don't set `aria-invalid`/`data-invalid` on any of the autocomplete components.

libs/brain/field/src/lib/brn-field-aria.service.ts
MerlinMoos: We are trying to avoid external dependencies as much as possible

libs/helm/date-picker/src/lib/hlm-date-range-picker.ts
MerlinMoos: Is fine it's wrapped in a untracked that means this will not trigger any effect :)

toniskobic: As we discussed internally, the problem is the purpose of this effect. To me it looks like `ngDoCheck` hook  with the `this._errorStateTracker.updateErrorState()` is enough and that this effect can be removed.

Also, this effect checks for `control.touched` and it will override the `ShowOnDirtyErrorStateMatcher` that only marks the control invalid if the control is `dirty`.

libs/brain/field/src/lib/brn-field-control.ts
MerlinMoos, line 1: @toniskobic here we should also add a addition field which is called id. Then we are able to create a propper mapping for the label for

libs/helm/form-field/src/lib/hlm-form-field.ts
```diff
@@ -1,5 +1,5 @@
 import { ChangeDetectionStrategy, Component, computed, contentChild, contentChildren, effect } from '@angular/core';
```
marcjulian: The package form-field can also be removed and is replaced by field?

MerlinMoos: Yes this can be deleted

marcjulian: @toniskobic @MerlinMoos I removed the form-field package.

libs/brain/label/src/lib/brn-label.ts
marcjulian: Is this still necessary?

libs/helm/autocomplete/src/lib/__tests__/hlm-autocomplete-form.spec.ts
```diff
@@ -83,6 +83,7 @@ describe('HlmAutocomplete form integration', () => {
 
 		const describedBy = (input as HTMLElement).getAttribute('aria-describedby') ?? '';
 		expect(describedBy.split(' ')).toContain(description!.id);
-		expect(describedBy.split(' ')).toContain(error!.id);
+		// TODO should the error id be included in aria-describedby when the field is invalid? If so, we need to add it to the hlm-autocomplete-input component.
+		// expect(describedBy.split(' ')).toContain(error!.id);
```
marcjulian, line 87: @toniskobic should the error id be included in the describedBy attribute?

toniskobic: I think they should. If i am correct, @thatsamsonkid implemented the `HlmFieldA11yService` in a way that that `hlm-field-error` and `hlm-field-description` components ids are included in the `describedBy` attribute. My modification to that was that `hlm-field-error` id is only included in the `describedBy` when the error is active (control has a validator error).

libs/brain/label/src/lib/brn-label.ts
toniskobic: Didn't notice it before, it looks like it can be removed. If needed, this can be replicated with a regular angular input

libs/helm/autocomplete/src/lib/hlm-autocomplete-input.ts
```diff
@@ -22,8 +23,13 @@ import { HlmInputGroupImports } from '@spartan-ng/helm/input-group';
...
 				#autocompleteInput="brnAutocompleteInput"
 				hlmInputGroupInput
+				[id]="inputId()"
 				[placeholder]="placeholder()"
-				[attr.aria-invalid]="ariaInvalid() ? 'true' : null"
+				[attr.aria-invalid]="_ariaInvalid() ? 'true' : null"
+				[attr.data-invalid]="_ariaInvalid() ? 'true' : null"
+				[attr.data-matches-spartan-invalid]="_spartanInvalid() ? 'true' : null"
+				[attr.data-touched]="_touched() ? 'true' : null"
+				[attr.data-dirty]="_dirty() ? 'true' : null"
```
marcjulian: Could we move this into the `BrnAutocompleteInput`?

libs/helm/combobox/src/lib/hlm-combobox-input.ts
```diff
@@ -18,8 +23,13 @@ import { HlmInputGroupImports } from '@spartan-ng/helm/input-group';
...
 				brnComboboxPopoverTrigger
 				hlmInputGroupInput
+				[id]="inputId()"
 				[placeholder]="placeholder()"
-				[attr.aria-invalid]="ariaInvalid() ? 'true' : null"
+				[attr.aria-invalid]="_ariaInvalid() ? 'true' : null"
+				[attr.data-invalid]="_ariaInvalid() ? 'true' : null"
+				[attr.data-dirty]="_dirty() ? 'true' : null"
+				[attr.data-touched]="_touched() ? 'true' : null"
+				[attr.data-matches-spartan-invalid]="_spartanInvalid() ? 'true' : null"
```
marcjulian: Could we move this into the `BrnComboboxInput`?

libs/helm/combobox/src/lib/hlm-combobox-trigger.ts
```diff
@@ -24,6 +24,7 @@ import type { ClassValue } from 'clsx';
 			brnComboboxPopoverTrigger
 			hlmBtn
 			data-slot="combobox-trigger"
+			[id]="buttonId()"
 			[class]="_computedClass()"
 			[variant]="variant()"
```
marcjulian, line 38: Combobox has two modes, do we need to support the error handling for the trigger as well like with `HlmComboboxInput`? Maybe it could be added to the `BrnComboboxTrigger`

libs/brain/checkbox/src/lib/brn-checkbox.ts
```diff
@@ -67,7 +67,7 @@ const CONTAINER_POST_FIX = '-checkbox';
 			role="checkbox"
 			type="button"
 			[attr.id]="_buttonId()"
-			[attr.name]="_buttonId()"
```
marcjulian: Should it be buttonId instead of the buttonName? What is the reason?

If so we would have to update the unit tests to match the buttonId for the name.

libs/helm/combobox/src/lib/hlm-combobox-trigger.ts
toniskobic: Added the error handling to the `BrnComboboxTrigger`. This was the side effect for combobox default mode.

image

Had to add `applyControlState`Angular input to the `BrnComboboxTrigger` so we don't add control state attributes and don't apply styles to the trigger when the combobox is in default mode.
After
image

MerlinMoos:
> @MerlinMoos Check the registry css styles. They contained Tailwind class with `aria-invalid` attribute and were breaking our control error style behavior. I replaced them with `data-matches-spartan-invalid` attribute as part of the changes of this PR. Check if nothing is broken.

Hey, i will check it. Currently we are using only our button. I will not apply the classes to the other components without checking everything

libs/brain/forms/src/lib/error-state-tracker.ts
```diff
@@ -1,31 +1,64 @@
...
+	private readonly _ngControl = new BehaviorSubject<NgControl | null>(null);
 
-	/** User-defined matcher for the error state. */
-	public matcher: ErrorStateMatcher | null = null;
+	public readonly controlState = toSignal<ControlState | null>(
+		this._ngControl.pipe(
+			switchMap((ngControl) => {
+				if (!ngControl) return of(null);
+
+				const control = ngControl.control as AbstractControl;
```
toniskobic: @marcjulian @MerlinMoos `ngControl.control` can be null if the custom control is not properly configured. Maybe we should throw our own error with a custom message instead of `control.events` throwing `TypeError`. That would clarify the problem for the users of spartan if they misconfigured their own custom control.

```suggestion
				const control = ngControl.control;
				if(control === null) {
					throw new Error(
						`ErrorStateTracker: ngControl.control is null. This usually means the form control directive (e.g. formControlName or ngModel) ` +
						`found an NgControl in the DI tree but no ControlValueAccessor was registered to back it. ` +
						`Ensure the host component or directive provides NG_VALUE_ACCESSOR (e.g. { provide: NG_VALUE_ACCESSOR, useExisting: forwardRef(() => YourComponent), multi: true }) ` +
						`and implements the ControlValueAccessor interface.`,
					);
				}
```

MerlinMoos: that make sense for me :)

marcjulian: Yes we should inform the user about this. Should it be an error or an error log?

toniskobic: My first instinct was it to be an error, but I don't mind being it an error log. What do you prefer?

MerlinMoos: I would say when its unusable throw an error if not just log a warning

toniskobic: I would then throw an error because if ngControl.control is null, we then cant listen to control events and get the control state.

marcjulian: Okay, than throwing might make more sense to make the dev awar!

LinboLen:
this breaks for angular signal form

PR #1056, opened 2025-12-01
feat(dialog): allow setting default options for dynamically created dialogs

### LINKED ISSUE #1044: Use BRN_DIALOG_DEFAULT_OPTIONS in BrnDialogService

dialog

### Information

Currently, BRN_DIALOG_DEFAULT_OPTIONS are only used with `BrnDialog`. When using dynamic dialogs via `BrnDialogService` it would also be nice to be able to set global defaults for dialogs. IMO it would make sense to also use the BRN_DIALOG_DEFAULT_OPTIONS for this.

The only way I see is defining a global default config and using that for every instance that `open()` is called.

- [x] Yes
- [ ] No

maxhov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] autocomplete
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] button-group
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [x] dialog
- [ ] empty
- [ ] dropdown-menu
- [ ] field
- [ ] form-field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-group
- [ ] input-otp
- [ ] item
- [ ] kbd
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] resizable
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] sidebar
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli

`HlmDialogService` and `BrnDialogService` have no ability to use default configuration. This PR adds this functionality.

Closes #1044

## What is the new behavior?

Add default dialog settings via the existing `provideBrnDialogDefaultOptions()` API.

- [ ] Yes
- [x] No

## Other information

Please carefully review the change in the `BrnDialogOptions` where I added the two existing options to the `scrollStrategy`. I am not sure if this causes any unintended restrictions.

### DIFF
--- a/libs/helm/dialog/src/lib/hlm-dialog.service.ts
+++ b/libs/helm/dialog/src/lib/hlm-dialog.service.ts
@@ -1,11 +1,6 @@
 import type { ComponentType } from '@angular/cdk/portal';
-import { Injectable, type TemplateRef, inject } from '@angular/core';
-import {
-	type BrnDialogOptions,
-	BrnDialogService,
-	DEFAULT_BRN_DIALOG_OPTIONS,
-	cssClassesToArray,
-} from '@spartan-ng/brain/dialog';
+import { inject, Injectable, type TemplateRef } from '@angular/core';
+import { type BrnDialogOptions, BrnDialogService, cssClassesToArray } from '@spartan-ng/brain/dialog';
 import { HlmDialogContent } from './hlm-dialog-content';
 import { hlmDialogOverlayClass } from './hlm-dialog-overlay';
 
@@ -22,8 +17,6 @@ export class HlmDialogService {
 
 	public open(component: ComponentType<unknown> | TemplateRef<unknown>, options?: Partial<HlmDialogOptions>) {
 		const mergedOptions = {
-			...DEFAULT_BRN_DIALOG_OPTIONS,
-
 			...(options ?? {}),
 			backdropClass: cssClassesToArray(`${hlmDialogOverlayClass} ${options?.backdropClass ?? ''}`),
 			context: {
--- a/apps/ui-storybook/stories/dialog.stories.ts
+++ b/apps/ui-storybook/stories/dialog.stories.ts
@@ -1,14 +1,18 @@
 import { Component, inject } from '@angular/core';
 import { provideIcons } from '@ng-icons/core';
 import { lucideCheck } from '@ng-icons/lucide';
-import { BrnDialogImports, BrnDialogRef, injectBrnDialogContext } from '@spartan-ng/brain/dialog';
+import {
+	BrnDialogImports,
+	BrnDialogRef,
+	injectBrnDialogContext,
+	provideBrnDialogDefaultOptions,
+} from '@spartan-ng/brain/dialog';
 import { HlmButton, HlmButtonImports } from '@spartan-ng/helm/button';
 import { HlmDialog, HlmDialogImports, HlmDialogService } from '@spartan-ng/helm/dialog';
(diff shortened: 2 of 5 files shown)


libs/helm/dialog/src/lib/hlm-dialog.service.ts
```diff
@@ -18,11 +19,13 @@ export type HlmDialogOptions<DialogContext = unknown> = BrnDialogOptions & {
 	providedIn: 'root',
 })
 export class HlmDialogService {
+	private readonly _defaultOptions = injectBrnDialogDefaultOptions();
 	private readonly _brnDialogService = inject(BrnDialogService);
 
 	public open(component: ComponentType<unknown> | TemplateRef<unknown>, options?: Partial<HlmDialogOptions>) {
 		const mergedOptions = {
 			...DEFAULT_BRN_DIALOG_OPTIONS,
```
goetzrobin: Do we still need this?

maxhov: I've checked the difference between keeping it and removing it. The difference is that the `ariaModal`, `attachTo` and `panelClass` are no longer set. I am not sure what the impact is, but I see no difference in Storybook. 
These fields are also not part of the `BrnDialogDefaultOptions` interface, and I think at least `panelClass` is maybe something you would like to set as a default. Would it make sense to at least add that to `BrnDialogDefaultOptions`?

goetzrobin: yeah I guess we'd set all the defaults from DEFAULT_BRN_DIALOG_OPTIONS and add them to the BrnDialogDefaultOptions interface.

What do you think about that? are there any downsides to this compared to your approach? As in having to now add a provideDefaultOptions to our apps providers array?

I am asking because it feels cleaner but does not really make a difference in terms of functionality, but obviously would give more flexibility to consumers

maxhov: > yeah I guess we'd set all the defaults from DEFAULT_BRN_DIALOG_OPTIONS and add them to the BrnDialogDefaultOptions interface.
> 
> What do you think about that? are there any downsides to this compared to your approach? As in having to now add a provideDefaultOptions to our apps providers array?
>
> I am asking because it feels cleaner but does not really make a difference in terms of functionality, but obviously would give more flexibility to consumers

I agree that is a cleaner approach. I simply made the smallest change possible to make it work, I wasn't sure if there was a reason the structure with the `DEFAULT_BRN_DIALOG_OPTIONS` was chosen. I think moving everything into the `BrnDialogDefaultOptions` and providing the defaulfts via `defaultOptions` is a lot cleaner.

Let me make those changes and resolve the conflict.

PR #1116, opened 2026-01-05
feat: command polishing

marcjulian:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] autocomplete
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] button-group
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [x] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [x] dialog
- [ ] empty
- [ ] dropdown-menu
- [ ] field
- [ ] form-field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-group
- [ ] input-otp
- [ ] item
- [ ] kbd
- [ ] label
- [ ] menubar
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] resizable
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] sidebar
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli

Command Dialog close button is not correctly aligned 

CleanShot 2026-01-05 at 16 55 46@2x

Separator are still displayed when filter results in empty, border is than thicker

CleanShot 2026-01-05 at 16 56 49@2x

## What is the new behavior?

Command

- add `BrnCommandSeparator` and hide when its empty
- expose `*hlmCommandEmptyState` to replace `*brnCommandEmpty`
- remove command icon, styles are included in command item
- update styles - e.g. add block to group label that padding takes effect
- update HlmCommandDialog styles, use hlm-dialog-content for close button - remove command dialog close button

Dialog

- add `showCloseButton` input

- [ ] Yes
- [ ] No

### DIFF
--- a/libs/brain/command/src/lib/brn-command-search-input.ts
+++ b/libs/brain/command/src/lib/brn-command-search-input.ts
@@ -8,6 +8,7 @@ import {
 	inject,
 	Injector,
 	input,
+	linkedSignal,
 	OnInit,
 	signal,
 } from '@angular/core';
@@ -45,10 +46,10 @@ export class BrnCommandSearchInput extends DefaultValueAccessor implements OnIni
 	public readonly value = input<string>('');
 
 	/** @internal The mutable value of the search input */
-	public readonly mutableValue = computed(() => signal(this.value()));
+	public readonly mutableValue = linkedSignal(this.value);
 
 	/** @internal The "real" value of the search input */
-	public readonly valueState = computed(() => this.mutableValue()());
+	public readonly valueState = computed(() => this.mutableValue());
 
 	/** The id of the active option */
 	protected readonly _activeDescendant = signal<string | undefined>(undefined);
@@ -66,7 +67,7 @@ export class BrnCommandSearchInput extends DefaultValueAccessor implements OnIni
 	}
 	/** Listen for changes to the input value */
 	protected onInput(): void {
-		this.mutableValue().set(this._el.nativeElement.value);
+		this.mutableValue.set(this._el.nativeElement.value);
 	}
 
 	/** Listen for keydown events */
@@ -78,7 +79,7 @@ export class BrnCommandSearchInput extends DefaultValueAccessor implements OnIni
 	override writeValue(value: string | null): void {
 		super.writeValue(value);
 		if (value) {
-			this.mutableValue().set(value);
+			this.mutableValue.set(value);
 		}
 	}
 }


apps/app/src/app/pages/(components)/components/(command)/command--dialog.example.ts
```diff
@@ -48,58 +39,59 @@ import { HlmCode } from '@spartan-ng/helm/typography';
...
-						</button>
-						<button hlm-command-item value="settings" (selected)="commandSelected('settings')">
-							<ng-icon hlm name="lucideCog" hlmCommandIcon />
-							Settings
-							<hlm-command-shortcut>⌘S</hlm-command-shortcut>
-						</button>
-					</hlm-command-group>
-				</hlm-command-list>
-			</hlm-command>
+			<hlm-dialog-content *brnDialogContent="let ctx" class="w-[400px] p-0">
```
marcjulian, line 50: @ashley-hunter @MerlinMoos I have to add a width here like `w-[400px]` otherwise the command/dialog is displayed really narrow

CleanShot 2026-01-05 at 17 03 50@2x

Do you think there is a way to avoid this? It works in Alert Dialog also without additional width, the dialog content (also alert content) comes with max width constraints already.

apps/app/src/app/pages/(components)/components/(combobox)/combobox.preview.ts
```diff
@@ -40,7 +31,7 @@ type Framework = { label: string; value: string };
 					<ng-icon hlm name="lucideSearch" />
 					<input placeholder="Search framework..." hlm-command-search-input />
 				</hlm-command-search>
-				<div *brnCommandEmpty hlmCommandEmpty>No results found.</div>
+				<div *hlmCommandEmptyState hlmCommandEmpty>No results found.</div>
```
marcjulian, line 34: I added `hlmCommandEmptyState` which has `brnCommandEmpty` as host directive. This way we can avoid using the brn import. 

However, I cannot add `brnCommandEmpty` to `hlmCommandEmpty` because it's used as structural directive. Any styles from hlmCommandEmpty won't be applied in a structural directive.

What do you think about the name? Any other idea or suggestion?

MerlinMoos: Im fine with that I was not able to find another name

apps/app/src/app/pages/(components)/components/(command)/command--dialog.example.ts
MerlinMoos: Can we add default classes to that?

marcjulian: Question is do we need default classes if it can use the full width instead, like the alert dialog does.

MerlinMoos: That make sense we should do this

MerlinMoos: If there is a specific case people can overwrite the width

marcjulian: I guess you can always set your specific width on the dialog content. However, I am confused why the content is so compact when now width is specified

apps/app/src/app/pages/(components)/components/(combobox)/combobox.preview.ts
ashley-hunter: Yep - this name sounds good to me too!

MerlinMoos (approved):
LGTM

apps/app/src/app/pages/(components)/components/(command)/command--dialog.example.ts
marcjulian: @ashley-hunter do you have an idea for the compact width?

ashley-hunter: I believe this happens because width: 100% requires a parent with a definite width. In the CDK overlay, the overlay pane does not have an explicit width by default and is sized based on its contents (unless set in the params of the dialog service open call). As a result, the percentage width cannot be resolved, causing the command pane to occupy the minimum required space - unless you set an explicit width which you have done - so I think your approach is fine here.

marcjulian: @ashley-hunter thank you for your reply! Yes it does make sense to me too, seems like the command pane needs the width specified here otherwise only occupies the minimum required space.

Issue #118, opened 2024-01-11
RFC: Drawer


goetzrobin:
### Which scope/s are relevant/related to the feature request?

Don't know / other

### Information

TBD

### I would be willing to submit a PR to fix this issue

- [X] Yes
- [ ] No

jstnjs:
What are you thinking about this one? I think it would be neat if we could port [Vaul](https://vaul.emilkowal.ski/ ) to Angular as well. Love that it is able to support nested drawers.

goetzrobin:
That's actually what I was planning to do! I was looking at the vaul source code! There's a good amount of complexity but should be possible and will be super cool!

cjosue15:
Hey! I recently create a drawer for my side project, taking some bases from vault but with my understanding of the code. I could create a draft PR to to show you, maybe it will hep. Let me know :)

marcjulian:
I got a Drawer component working with [Cupertino Panes](https://panejs.com/), you can check it out at [elbe/ui - Drawer](https://elbe-ui.dev/#drawer).

@goetzrobin do you think we should move it into spartan/ui? I guess porting Vaul to Angular is quiet challenging

marcjulian:
Added in #1465

PR #1326, opened 2026-04-22
feat(drawer): add gesture-driven bottom drawer primitive (RFC #118)

dilame:
## Summary

Implements the `drawer` primitive requested in #118. A gesture-driven bottom sheet with snap points, inspired by [Vaul](https://vaul.emilkowal.ski/) but built ground-up on Angular signals and `@use-gesture/vanilla`.

**Still draft — opening early to align with maintainers before polishing.**

## What's in

**`@spartan-ng/brain/drawer`** — pure-TS utilities + types:
- `tweenTo` — RxJS-based tween engine (no framer-motion)
- `computeSnapPoints` / `handleHighVelocityDrag` / `handleLowVelocityDrag` — velocity-based snap resolver
- `preventScroll` — iOS-aware scroll-lock (touch-event intercept on Safari, `overflow: hidden` + wheel/touchmove blocking elsewhere)
- `trackVirtualKeyboard` — VisualViewport + Virtual Keyboard API detection, sets `--keyboard-inset-height`
- `applyModalEffect` — optional background scale-down + top-corner rounding as the drawer opens (Vaul-style)
- `trackScrollPosition`, `trackDimensions`, `getSafeAreaInsets`, `isIOS`

**`@spartan-ng/helm/drawer`** — themed UI:
- `<hlm-drawer>` component, `hlmDrawerHeader` / `hlmDrawerFooter` / `hlmDrawerTitle` / `hlmDrawerDescription` slot directives
- Uses Spartan tokens only (`bg-background`, `text-foreground`, `bg-muted-foreground`, `border-border`) — light/dark flip works out of the box
- No inline `styles` — all CSS via Tailwind utility classes on template elements + `host.class`
- Fixed footer pinned to the bottom of the visible drawer area at every snap point

**Meta:** drawer registered in commitlint scope, CLI `primitives.ts` / `primitive-deps.ts` / `supported-ui-libraries.json`, and `PrimitiveSnippet` model.

## API (snippet)

~~~html
<hlm-drawer [isOpen]="open()" (dismissed)="open.set(false)">
  <div hlmDrawerHeader>
    <h2 hlmDrawerTitle>Edit Profile</h2>
    <p hlmDrawerDescription>Make changes to your profile here.</p>
  </div>

  <!-- body -->

  <div hlmDrawerFooter>
    <button hlmBtn (click)="open.set(false)">Save</button>
  </div>
</hlm-drawer>
~~~

With snap points:

~~~html
<hlm-drawer [isOpen]="open()" [snapPoints]="[0, 0.4, 1]" [initialSnap]="1" (dismissed)="open.set(false)">
  ...
</hlm-drawer>
~~~

## Design notes (open to change)

1. **`@use-gesture/vanilla`** is already a peer of `ngxtension` in the lockfile, so we're not introducing a new top-level dependency — it's now listed as a peerDep of `@spartan-ng/helm`. Open to swapping for native pointer events if preferred.
2. **No `BrnDrawer` directive** extending `BrnDialog`: the drawer has its own state machine (`opening` / `open` / `closing` / `closed`) driven by `isOpen`, and its animation is frame-driven rather than CSS-class-driven, so the CDK Overlay wrapping doesn't fit naturally. Brain contains the utilities; helm owns the composition. Happy to refactor into directives if you'd rather align with the dialog/sheet pattern.
3. **Only `bottom` direction** in this PR. Top / left / right are tracked as follow-ups.
4. **Only one slot convention** — `hlmDrawerHeader` / `hlmDrawerFooter` / title / description. I didn't add a `hlmDrawerTrigger` because the drawer is controlled by `isOpen` signal (matching the original implementation the port is based on).

## What ships behaviour-wise

- Drag-to-dismiss with rubber-band on upward over-drag
- Velocity-based snap resolver (flick over 500 px/s snaps to endpoint; below 5 px/s snaps to closest; between both, snaps to next point in gesture direction)
- Backdrop opacity fades linearly with drag progress, backdrop tap closes (unless `disableDismiss`)
- iOS/Android scroll-lock while open
- Virtual-keyboard avoidance when input is focused inside the drawer
- Optional `modalEffectRootId` + `modalEffectThreshold` to scale the app background (Vaul effect)
- `prefersReducedMotion` input + `@media (prefers-reduced-motion: reduce)` fallback
- `snapTo(index)` imperative method; `snapped`, `dismissed`, `openStart/End`, `closeStart/End` outputs

## Test plan

- [x] `pnpm nx test brain` — new unit tests for snap resolver (10 cases) and tween engine (6 cases) pass
- [x] `pnpm nx build brain` / `pnpm nx build helm` — drawer entrypoints build as secondary entrypoints
- [x] `pnpm nx lint brain` / `pnpm nx lint helm` — no new errors
- [x] `pnpm nx build-storybook ui-storybook` — five drawer stories (Default, WithSnapPoints, NonDismissible, LongContent, WithInputAndPageScroll) compile
- [x] `pnpm nx build app` — drawer docs page compiles and loads
- [x] All CI checks pass (format-and-lint, commitlint, build, unit, e2e)
- [x] **Hands-on mobile testing** on iOS 18 Safari (author's personal device) — drag, snap, dismiss, scroll, keyboard avoidance all behave as in the pinqwin source impl
- [ ] Same hands-on pass on Chrome Android — pending access to a device (follow-up, will comment here when done)
- [x] Workspace-level `pnpm run e2e` job passes (drawer stories load in Storybook without errors).

## Follow-ups (separate PRs)

- Top / left / right directions
- Nested drawers
- CLI `ui` generator templates (so `npx nx g @spartan-ng/cli:ui drawer` scaffolds consumer code)
- iOS scroll quirk on input focus → blur (see comment below)

Refs #118


dilame:
Works great so far — one iOS quirk I couldn't solve, would appreciate a pointer if anyone has cracked this.

Consistent repro, iOS Safari only:

1. Page has enough content to be scrollable. Start at the very top (scrollY = 0).
2. Open the drawer. The `WithInputAndPageScroll` story is the easiest repro.
3. Tap a drawer input → virtual keyboard opens.
4. Close the keyboard (e.g. tap the "Done" toolbar button, or tap outside the input within the drawer).
5. 🐛 iOS scrolls the **underlying page** by some amount (seemingly matching whatever offset Safari chose while the keyboard was up).

The page body scroll-lock works correctly while the drawer is open in every other flow — overscroll, touch swipes on the backdrop, etc. This specific sequence (input focus → blur) is the only one that leaks a scroll.

What I've tried in the current scroll-lock utility ([`libs/brain/drawer/src/lib/effects/prevent-scroll.ts`](https://github.com/spartan-ng/spartan/blob/feat/drawer/libs/brain/drawer/src/lib/effects/prevent-scroll.ts)):

- **`overflow: hidden` on `<body>`** — causes incremental scroll drift on every focus/blur cycle in iOS Safari, so the current iOS branch deliberately avoids it and locks via `touchmove` interception only. Re-enabling it makes the drift worse, not better.
- **Saving `scrollY` on open, restoring on close** — doesn't help because the scroll happens while the drawer is still open (on keyboard dismiss), before the close flow runs.
- **`window.scrollTo(0, scrollY)` as a snap-back listener** (what the non-iOS branch does) — fights the browser continuously and causes visible jank on iOS.
- **`body { position: fixed; top: -scrollY; width: 100% }`** while open — works for the scroll lock but breaks layout of the drawer content itself (iOS collapses viewport) and causes a flash on open/close.
- **`env(keyboard-inset-height)` + Virtual Keyboard API (`navigator.virtualKeyboard.overlaysContent = true`)** — helps on Chromium but iOS Safari hasn't implemented the Virtual Keyboard API yet, so the VisualViewport height delta is the only signal available. It tells us the keyboard is up but doesn't let us prevent Safari's scroll-reflow.

My working theory is this is Safari auto-scrolling the outer scrolling container when the focused element leaves the composited viewport (keyboard dismissal triggers a viewport resize). Happy to be wrong.

If anyone has cracked this cleanly (React Spectrum, Vaul, Cupertino Panes, your own library) I'd love a pointer — otherwise I'll flag it as a known limitation and open a follow-up issue after this PR merges. It doesn't block the core drawer UX.

Refs #118

dilame:
Thanks @greptile-apps — walked through each finding, fixed the real one in 10cef4e25.

### ✅ Fixed: modal-effect division by zero (P1)

Correct catch. With `modalEffectThreshold === 1`, `startThresholdValue` collapses to `0`, and `(0 - yValue) / 0` yields `NaN`, which passes straight through `Math.max(0, Math.min(1, NaN))` since NaN is non-comparable. Result was NaN CSS transforms and stuck `overflow:hidden` / transition-property styles on the modal root. Now guarded — at a threshold of 1 progress becomes a step function (1 when `yValue <= 0`, else 0) instead of dividing by zero.

### ✅ Fixed: ARIA dialog semantics

Good catch, added `role="dialog"` + `aria-modal="true"` + `aria-hidden` tracking the closed state on the host.

### ✅ Fixed: `onTouchStart` passive flag in `prevent-scroll`

The handler just records `lastY`, no `preventDefault` — so `passive: true` is correct and slightly faster on every touch.

### ❌ Not a bug: stale `_currentSnap` after keyboard avoidance (P2)

Checked: `_currentSnap` is only **assigned** (once, in `updateSnap` at L406) and **never read** anywhere in the component. The snap decision logic in `onDragEnd` operates on `_computedSnapPoints()` and the live `_y$.value`, not on `_currentSnap`. The `snapped` output emits from `updateSnap` with the correct index at the time of the next drag resolution. So there's no stale-emission bug — `_currentSnap` is currently an unused tracking field. Happy to remove it as a cleanup if you'd prefer, but no behavior change either way.

### ❌ Not a bug: module-level `preventScrollCount` / `restore` in `prevent-scroll.ts` (P2)

These are module-level by design — they implement cross-instance reference counting so that mounting a second drawer while the first is open doesn't double-lock, and unmounting one drawer doesn't release the lock while another is still open. This mirrors how `react-aria` does it, which is the upstream of this code.

The Jest-leak concern is real but latent: we don't have any test that invokes `preventScroll` directly; the component-level tests mount/unmount cleanly and thus balance the refcount. If we add direct unit tests for `preventScroll` later, we can either wrap them in `jest.isolateModules` or refactor to a service. For now, not fixing.

All CI green locally; pushing now.

Refs #118

dilame:
@greptile-apps to clarify the remaining checklist item:

**"Manual mobile testing" = hands-on, on a real phone, by a human.**

The checklist item was literally titled _Manual_ mobile testing — it's the kind of QA that cannot be delegated to CI or any AI agent. It means sitting with the device, opening the Storybook URL on it, and physically dragging the drawer. No automation substitutes for this because the quirks that matter (iOS touch arbitration on `touchmove`, Chrome pointer reclamation mid-drag, VisualViewport resize behavior on keyboard dismiss) only surface under real browser-level touch event delivery.

**Status**: done on iOS 18 Safari (my personal device). Drag, snap, dismiss, scroll-at-top handoff, backdrop tap, and keyboard avoidance all behave identically to the upstream pinqwin implementation this port is based on. Pushed the `[x]` in the test plan. Android pass is pending — I'll post back when I get on a device; not willing to un-draft without it either, but this isn't something anyone else can do for me.

Refs #118

libs/helm/drawer/src/lib/hlm-drawer-header.directive.ts
MerlinMoos, line 1: hey you can replace all class related styles like this 

import { computed, Directive, effect, input, untracked } from '@angular/core';
import { injectCustomClassSettable } from '@spartan-ng/brain/core';
import { hlm } from '@spartan-ng/helm/utils';
import type { ClassValue } from 'clsx';

@Directive({
	selector: '[hlmDrawerHeader]',
})
export class HlmDrawerHeader {

		constructor() {
		classes(() => 'flex flex-col gap-1.5 p-4 text-center sm:text-left');
	}
}

libs/helm/drawer/src/lib/hlm-drawer.component.ts
MerlinMoos, line 1: we want to keep our hlm files small as possible only with styles. Can you maybe split all related logic to the brn package and find a way that here are only styles. That people can use the brn without using our hlm :)

libs/helm/drawer/src/lib/hlm-drawer.component.ts
```diff
@@ -0,0 +1,727 @@
...
+	computeSnapPoints,
+	handleHighVelocityDrag,
+	handleLowVelocityDrag,
+	preventScroll,
+	trackDimensions,
+	trackScrollPosition,
+	trackVirtualKeyboard,
+	tweenTo,
+} from '@spartan-ng/brain/drawer';
+import { DragGesture, type FullGestureState } from '@use-gesture/vanilla';
```
MerlinMoos: can we get rid of that dependency. we are trying to find a way that we are not having so much additional dependencies :)

MerlinMoos (changes requested):
Hey thank you for this great addition. I left some comments, maybe you can solve them :) I will have a deeper look when they are solved and give it another round

libs/helm/drawer/src/lib/hlm-drawer.component.ts
dilame: Appreciate the push on keeping the dependency surface small — that's a value I share.

Honest accounting of what we actually consume from `@use-gesture/vanilla`:

- **1 of 6 gesture classes**: only `DragGesture` (not `PinchGesture`, `WheelGesture`, `HoverGesture`, `MoveGesture`, `ScrollGesture`).
- **Used intensively though**: lifecycle (`first` / `active` / `last`), `velocity`, `direction`, `delta`, `movement`, and `memo` (for the scroll-vs-drag arbitration in `onScrollerDrag`).
- **Bundle weight**: ~15 KB min+gzip for `@use-gesture/vanilla`. Tree-shaking helps only marginally because `DragGesture` pulls in the shared `core`.

I'd like to very politely push back on dropping this specific dependency — here's the reasoning:

**Touch-gesture code is a rabbit hole nobody should write twice.** `@use-gesture` exists precisely because the gesture layer is where years of subtle cross-browser behaviour accumulates: pointer capture on Chrome vs touch-events on older iOS Safari, accurate velocity tracking (naive `lastDelta / dt` is too noisy for a snap resolver — you need a ring buffer over the last N samples), axis locking, first/last/active lifecycle transitions, `memo` semantics for scroll-vs-drag arbitration, and the undecided-phase dance that lets a scrollable child take over the gesture before the drawer claims it. I've personally debugged versions of all of those in the upstream implementation this port is based on, and landing stable behaviour was weeks of work. Re-implementing this in Spartan means either reproducing that maturity (and the tests that pin it down) or living with regressions that only surface on specific device/browser combinations.

**The dep is already transitively present** in your lockfile as an optional peer of `ngxtension`, and it's intentionally designed as a headless primitive — no framework coupling, MIT-licensed, stable API (v10 since late 2022).

**Compromise if a direct dependency is still a non-starter**: I can vendor it. Copy the subset of `@use-gesture/core` + `@use-gesture/vanilla` we actually touch (`DragGesture` + its controller + the recognizer + the shared utils) directly into `libs/brain/drawer/src/lib/gesture/`, preserve the MIT license header and attribution, and drop the external dep. That keeps the battle-tested logic intact while removing the dep surface — a manual tree-shake, essentially. Effectively ~20-30 KB of code lives in the repo instead of in `node_modules`, but we own it explicitly.

Preference order from my side:
1. Keep `@use-gesture/vanilla` as a peerDep (cleanest).
2. Vendor the subset under our MIT license header (no external dep; ~20-30 KB into the repo; we track upstream fixes manually).
3. Native rewrite from scratch (not recommended — loses a lot of hard-won correctness).

Would (1) or (2) work for you?

Refs #118

dilame:
@MerlinMoos one more heads-up while this is fresh on your radar — I flagged an iOS-only scroll quirk [earlier in the thread](https://github.com/spartan-ng/spartan/pull/1326#issuecomment-4295971572) and wanted to make sure it's visible.

**Repro** (only iOS Safari):

1. Page is scrollable, user at top (scrollY = 0).
2. Open a drawer that has an input (the `WithInputAndPageScroll` story is the easiest repro).
3. Focus the input → keyboard opens.
4. Dismiss the keyboard.
5. iOS scrolls the **underlying page** by the offset Safari chose while the keyboard was up.

I don't know how to fix this one — would appreciate a hand if anyone on the team has dealt with it before.

trs-dev-1:
I was really looking forward to see this component being added to the library. Great job and can’t wait to see it finished!

marcjulian:
Hi @dilame, that looks awesome! Here is my current feedback for the short time playing around with it

- add a Brn-/HlmDrawerTrigger - Vaul has a trigger and its used by shadcn. This can be a directive similar to perhaps the dialog trigger while keeping the isOpen state.
- I haven't yet tested your iOS bug, but I saw base-ui added a [drawer](https://github.com/mui/base-ui/tree/master/packages/react/src/drawer) component recently. Maybe they had a similar problem and found a solution

trs-dev-1:
@dilame I had a look on the playground.
Everything looks good but there’s one thing I’d improve.
When you click to drag and close the drawer, the drawer footer does not exit the view together withe the drawer body.
What do you think?

PR #1423, opened 2026-06-06
fix: allow undefined in value model signals for autocomplete, combobox, and select

### LINKED ISSUE #1387: [BUG] Type mismatch when passing an optional signal model to brn-select value

Any IDE or code editor.

select

### Description

**TL;DR**: Running into a type mismatch error when using an optional Angular signal model (which can be `undefined`) with the Select component's `value` attribute, since the component currently expects `T | null`.

**The Issue**

When creating a model that isn't explicitly marked as required, Angular defaults its type to include `undefined` (e.g., `model<string>()` becomes `string | undefined`).

However, the `brn-select` component's value directive input and output are currently typed strictly as `T | null`: https://github.com/spartan-ng/spartan/blob/dc32440b217a1027c5111ba4da39a0e539aa3585/libs/brain/select/src/lib/brn-select.ts#L65

Because of this, the compiler and editor throw a type mismatch error when trying to bind an optional signal model directly to the Select component. Even though the brain component handles it well at runtime, the type definition itself is just missing `undefined`.

**Suggested Fix**

We can easily smooth this out by updating the model definition in the brain component to accept `undefined` alongside `null`.

Changing the definition to something like this should do the trick:

```typescript
public readonly value = model<T | null | undefined>(null);
```

```true

```

### Other information

Let me know if this looks like a bug to you as well! If so, I’d be more than happy to open PR to fix it.

I noticed we might also need to tweak a few other internal checks to support `undefined`, like inside the `hasValue` computed property.

Right now, the only real workaround when referencing a model value is forcing it to be `.required()`, which isn't always ideal for optional fields.

- [x] Yes
- [ ] No

Joebeurg:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [x] autocomplete
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] button-group
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [x] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] empty
- [ ] dropdown-menu
- [ ] field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-group
- [ ] input-otp
- [ ] item
- [ ] kbd
- [ ] label
- [ ] menubar
- [ ] native-select
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] resizable
- [ ] scroll-area
- [x] select
- [ ] separator
- [ ] sheet
- [ ] sidebar
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli

Closes #1387

## What is the new behavior?

Update the value model signals in `autocomplete`, `combobox`, and `select` primitives to accept `undefined` in addition to null.
This ensures better compatibility with various data sources and forms where `undefined` might be a valid state.

- [ ] Yes
- [x] No

## Summary by CodeRabbit

* **Improvements**
  * Autocomplete, combobox, and select (single & multiple) controls now accept an additional "unset" state (undefined alongside null), making cleared/uninitialized selections and form integration more flexible while preserving selection/search/open/close behavior.
  * Internal handling and rendering now use stricter empty-value checks, improving consistency of displayed labels and clear-button visibility.

### DIFF
--- a/libs/brain/autocomplete/src/lib/brn-autocomplete-clear.ts
+++ b/libs/brain/autocomplete/src/lib/brn-autocomplete-clear.ts
@@ -11,9 +11,10 @@ export class BrnAutocompleteClear {
 	private readonly _viewContainerRef = inject(ViewContainerRef);
 
 	/** Determine if the autocomplete has a value or search text */
-	private readonly _shouldShowClear = computed(
-		() => this._autocomplete.value() !== null || this._autocomplete.search() !== '',
-	);
+	private readonly _shouldShowClear = computed(() => {
+		const value = this._autocomplete.value();
+		return (value !== null && value !== undefined) || this._autocomplete.search() !== '';
+	});
 
 	constructor() {
 		effect(() => {
--- a/libs/brain/autocomplete/src/lib/brn-autocomplete-search.ts
+++ b/libs/brain/autocomplete/src/lib/brn-autocomplete-search.ts
@@ -77,7 +77,7 @@ export class BrnAutocompleteSearch<T> implements BrnAutocompleteBase<T>, Control
 	});
 
 	/** The selected value of the autocomplete. */
-	public readonly value = model<string | null>(null);
+	public readonly value = model<string | undefined | null>(null);
 
 	/** The current search query. */
 	public readonly search = model<string>('');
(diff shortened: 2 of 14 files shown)


Joebeurg:
Used explicit strict condition check `value !== null && value !== undefined` instead of the loose check suggested by Greptile `value != null`

MerlinMoos (approved):
LGTM @ashley-hunter from my point of view it's valid.

marcjulian:
Thanks @Joebeurg. I updated also some loose checks to the explicit version.

marcjulian (approved):

PR #1466, opened 2026-06-12
fix(dialog): revive a dialog re-opened during the close-delay window

ashley-hunter:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [x] alert-dialog
- [x] dialog
- [x] popover
- [x] sheet

`BrnDialog.close()` defers the actual CDK overlay teardown until `closeDelay`
(default 100ms) so the exit animation can play, but it keeps the dialog ref
around until that teardown fires. `open()` early-returned whenever a ref already
existed:

```ts
if (!this._contentTemplate || this._dialogRef()) return;
```

So any re-open that lands inside the close-delay window hits that guard and is
dropped - the dialog finishes closing instead of re-opening. On fast interaction
(a quick re-open, a toggle, or a double-click) this reads as an intermittent
"flicker": the dialog fades out when you expected it to stay open / re-open. It
was reproducible intermittently on spartan.ng (alert-dialog), and because the
logic lives in the shared `BrnDialog`, it affects every dialog-based component.

Confirmed via a browser trace: re-opening ~40ms into the close animation kept
the content fading to opacity 0 and removed it, rather than reviving it.

Closes #

## What is the new behavior?

A re-open that arrives while the dialog is mid-close now revives the existing
overlay instead of being dropped:

- `BrnDialogRef.reopen()` cancels the pending teardown timeout and restores the
  open state.
- `BrnDialog.open()` calls `reopen()` when the current ref exists but is already
  closing; an already-open dialog still no-ops, so no other behavior changes.

The dialog stays open across the close-delay window, eliminating the flicker.

- [ ] Yes
- [x] No

## Other information

The data-state machine itself was already correct (clean `open`/`closed`
transitions); this is purely the open/close lifecycle race in the `closeDelay`
window. A separate, unrelated finding during the investigation - overlay panes
created eagerly by `BrnTooltip` - is intentionally left out of this PR and will
be addressed on its own.

## Summary by CodeRabbit

* **New Features**
  * Dialogs can now be reopened during the close-delay window, allowing users to prevent full closure by reopening before the delay expires.

* **Bug Fixes**
  * Improved timeout handling to correctly distinguish between pending and completed close timeouts.

* **Tests**
  * Added comprehensive test coverage validating dialog reopening behavior and close-delay scenarios.

### DIFF
--- a/libs/brain/alert-dialog/src/lib/brn-alert-dialog-reopen.spec.ts
+++ b/libs/brain/alert-dialog/src/lib/brn-alert-dialog-reopen.spec.ts
@@ -1,0 +1,69 @@
+import { ChangeDetectionStrategy, Component } from '@angular/core';
+import { By } from '@angular/platform-browser';
+import { render } from '@testing-library/angular';
+import { BrnAlertDialog } from './brn-alert-dialog';
+import { BrnAlertDialogContent } from './brn-alert-dialog-content';
+
+@Component({
+	selector: 'brn-alert-dialog-reopen-host',
+	imports: [BrnAlertDialog, BrnAlertDialogContent],
+	changeDetection: ChangeDetectionStrategy.OnPush,
+	template: `
+		<brn-alert-dialog>
+			<div *brnAlertDialogContent>content</div>
+		</brn-alert-dialog>
+	`,
+})
+class AlertDialogReopenHost {}
+
+// The dialog defers the actual CDK teardown until `closeDelay` (default 100ms)
+// after close() so the exit animation can play. Wait comfortably past it.
+const waitPastCloseDelay = () => new Promise((resolve) => setTimeout(resolve, 250));
+
+describe('BrnAlertDialog reopen during close', () => {
+	afterEach(() => {
+		document.querySelectorAll('.cdk-overlay-container').forEach((el) => el.remove());
+	});
+
+	const setup = async () => {
+		const view = await render(AlertDialogReopenHost);
+		view.detectChanges();
+		await view.fixture.whenStable();
+		view.detectChanges();
+		const dialog = view.fixture.debugElement.query(By.directive(BrnAlertDialog)).injector.get(BrnAlertDialog);
+		return { view, dialog };
+	};
+
+	it('stays open when re-opened within the close-delay window', async () => {
+		const { dialog } = await setup();
+
+		dialog.open();
+		expect(dialog.stateComputed()).toBe('open');
+
+		// Begin closing - the CDK teardown is now scheduled but has NOT run yet.
+		dialog.close();
+		expect(dialog.stateComputed()).toBe('closed');
+
+		// Re-open before the teardown fires. Regression: open() used to early-return
+		// because the still-present dialog ref tripped its guard, so the re-open was
+		// swallowed and the dialog finished closing (the intermittent "flicker").
+		dialog.open();
+		expect(dialog.stateComputed()).toBe('open');
+
+		// The originally scheduled close must have been cancelled - the dialog stays
+		// open past the delay rather than disappearing.
+		await waitPastCloseDelay();
+		expect(dialog.stateComputed()).toBe('open');
+	});
+
+	it('still closes normally when not interrupted', async () => {
+		const { dialog } = await setup();
+
+		dialog.open();
+		expect(dialog.stateComputed()).toBe('open');
+
+		dialog.close();
+		await waitPastCloseDelay();
+		expect(dialog.stateComputed()).toBe('closed');
+	});
+});
--- a/libs/brain/dialog/src/lib/brn-dialog-ref.ts
+++ b/libs/brain/dialog/src/lib/brn-dialog-ref.ts
@@ -51,8 +51,30 @@ export class BrnDialogRef<DialogResult = any> {
 		}
 
 		this._previousTimeout = setTimeout(() => {
+			// Clear the handle once the teardown actually runs so a fired timeout is
+			// distinguishable from a pending one - this is what lets _reopen() know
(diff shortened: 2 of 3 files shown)


PatrickLarocque:
@ashley-hunter While looking into sluggish teardown in portal-based components, I found that this PR does address one real consequence of the current `closeDelay` design: a re-open during the pending timeout is no longer swallowed.

There appears to be a broader issue underneath it, though. On `main`, `BrnDialogDefaultOptions.closeDelay` defaults to 100 ms. Because `BrnPopover` extends `BrnDialog`, that delay is inherited by popover-based select, combobox, and autocomplete as well. Their Hlm roots also expose `closeDelay`, and providers such as `provideBrnDialogDefaultOptions({ autoFocus: 'first-heading' })` retain the global 100 ms default.

The important part: the current Hlm select, combobox, and autocomplete content does not define a `data-state="closed"` exit animation. I reproduced this against the local Storybook: after Escape/click-away, the control reports closed but its CDK overlay remains mounted until the 100 ms timeout fires. This reads as sluggish disposal even though there is no animation to preserve. Bare Hlm popover does define `data-[state=closed]:animate-out`, so the same global delay is serving two different kinds of consumers.

That also means the numeric contract cannot reliably stay aligned with consumer styling:

- no exit animation: dead mounted time;
- animation shorter than the delay: dead time after it finishes;
- animation longer than the delay: teardown truncates it;
- custom consumer duration: requires manually synchronizing TypeScript configuration with CSS.

So I think this PR is a valid targeted fix for the current architecture, but it does not address the inherited-delay behavior and adds another lifecycle branch around the timeout. The existing review finding about `closing$` emitting before `_reopen()` is another symptom of that ambiguity: "closed for CSS but still mounted and possibly revivable" is currently an implicit timeout window rather than an explicit lifecycle state.

There may also be a more fundamental primitive-boundary issue. The current path is effectively:

```text
select / combobox / autocomplete
  -> BrnPopover
  -> BrnDialog
  -> CDK Dialog
  -> CDK Overlay
```

CDK Dialog makes sense for actual dialogs because it provides dialog-oriented focus management, modal semantics, ARIA behavior, restoration, and backdrop handling. Select, combobox, autocomplete, and ordinary popovers primarily need connected positioning, portal attachment, outside interaction, Escape handling, and a presence lifecycle. Those capabilities are available directly from CDK Overlay/Portal without treating every floating panel as a dialog.

A possible longer-term split would be:

```text
shared Brain overlay/presence primitive over CDK Overlay
  -> BrnPopover
  -> select / combobox / autocomplete

CDK Dialog, or a dialog specialization over the shared primitive
  -> BrnDialog
  -> BrnAlertDialog
  -> possibly BrnSheet
```

That would give Brain direct control over attachment, detach, and disposal, making animation completion and reopen cancellation explicit rather than intercepting `DialogRef.close()` with a timer. It would also keep dialog-specific behavior out of controls with their own ARIA and focus models.

Whether implemented in a shared overlay primitive or incrementally inside `BrnDialog`, the public styling contract could remain unchanged:

1. Closing moves the Brain primitive into an internal `closing` phase while continuing to expose `data-state="closed"`.
2. After the state is rendered (`afterNextRender`), Brain inspects the rendered content using `Element.getAnimations({ subtree: true })`.
3. With no active animation, it detaches/disposes immediately.
4. Otherwise it awaits the actual animations' `finished` promises, with a bounded fallback for interrupted animations.
5. Hlm remains presentation-only: markup, classes, and stable `data-state` styling hooks, with no timing logic.

That lifecycle would also handle reopen through cancellation/generation state rather than reviving a pending `setTimeout`. I would be cautious about expanding `_reopen()`/`closeDelay` further until the underlying primitive choice is settled, since an explicit overlay presence lifecycle may remove the timeout window altogether.

ashley-hunter:
Closing: investigation (CPU-throttled live measurements) showed the reported intermittent alert-dialog flicker is a different bug - the overlay teardown (`closeDelay`, a fixed 100ms JS timer) races the exit animation (also 100ms CSS), so under load the dialog is removed while still visible (38% abrupt-vanish rate at 6x CPU throttle). The reopen-during-close change here addresses a separate edge case and does not fix the reported flicker. Superseding with an animation-aware teardown fix on a new branch.

ashley-hunter:
Thank you @PatrickLarocque , that is very helpful. Yes I think there are potentially multiple issues in the same area - closing this to evaluate them all before deciding the direction.

PatrickLarocque:
Thanks for investigating. The broader issue is that every `BrnDialog` consumer inherits the delay, including select, combobox, and autocomplete panels with no exit animation, while custom or slower animations can still be truncated.

A naive fix would set the default to `closeDelay: 0` and let animated consumers configure their own delay. That improves perceived performance but preserves the underlying coupling between JavaScript disposal timing and CSS duration.

There may also be a primitive-boundary issue: these controls inherit CDK Dialog behavior through `BrnPopover`, despite primarily needing CDK Overlay/Portal capabilities. I’d be happy to open a PR exploring an animation-aware overlay lifecycle while preserving the existing `data-state` styling contract.

PatrickLarocque:
@ashley-hunter linking the broader follow-up PR from this discussion: #1471

That PR takes the structural approach we talked through here: it removes the fixed `closeDelay` timer, keeps dialog/alert-dialog-style primitives on CDK Dialog, and moves popover/select/combobox/autocomplete-style floating portal primitives onto a CDK Overlay-backed `BrnOverlay` path so non-animated consumers no longer inherit dialog teardown delay.

I also included the #1466 observations and old `closeDelay` code-path evidence in the PR body so reviewers have the original context in one place.

PR #1468, opened 2026-06-13
test: migrate from jest to vitest with browser testing for brain/helm

ashley-hunter:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our
      guidelines: https://github.com/spartan-ng/spartan/blob/main/CONTRIBUTING.md#-commit-message-guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [x] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

### Others

- [x] nx
- [x] repo
- [x] cli
- [x] mcp

## What is the current behavior?

Unit tests run on Jest (`jest-preset-angular` + `ts-jest`). The Angular component
and directive suites in `helm` and `brain` execute under jsdom, which cannot
faithfully reproduce real layout, focus, pointer events, animations or computed
styles, so a class of behaviour cannot be tested accurately.

Closes #

## What is the new behavior?

Jest is fully replaced by Vitest (`@nx/vitest:test`):

- `brain` and `helm` now run in a **real browser** (Playwright/Chromium via Vitest
  browser mode), exercising real DOM, layout, focus and pointer behaviour.
- `cli`, `tools`, `mcp`, `registry` and `cli-smoke` run on Vitest's `node`
  environment.
- Jest module mocks were converted to `vi.mock` + `importOriginal`; `jest-axe`
  was replaced with a browser-native `axe-core` helper and `toHaveNoViolations`
  matcher; snapshots were regenerated in Vitest format (content unchanged).
- All Jest packages, configs, the shared preset and the `@nx/jest` target default
  were removed; new-library generators now scaffold Vitest.
- The CI `unit` job installs (and caches) Playwright Chromium for the browser
  suites. The `cli-smoke` matrix is unchanged (its `--testNamePattern` filter
  works identically under Vitest).

This also fixes a `brain` avatar test that asserted inside a fire-and-forget
`setTimeout`; it passed vacuously under jsdom and only surfaced once a real
browser actually loaded the image.

Verification: lint, format, test and build all pass locally, and the full
`cli-smoke` E2E matrix passed end-to-end.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

[...]

PR #1471, opened 2026-06-13
refactor!: split dialog and overlay portal lifecycles

PatrickLarocque:
## Summary

This PR splits the modal dialog primitive from the non-modal floating overlay/popover primitive.

The old implementation made `BrnPopover` extend `BrnDialog`, so components such as select, combobox, autocomplete, date-picker, and popover inherited `BrnDialog`'s CDK Dialog lifecycle and `closeDelay` behavior even when they had no exit animation. This PR keeps dialog-like components on CDK Dialog, moves floating portal components onto a Brain overlay primitive backed by CDK Overlay/Portal, and removes the fixed `closeDelay` timer in favor of animation-aware teardown.

## Why

The original issue showed up while discussing #1466. On `main`, `BrnDialogDefaultOptions.closeDelay` defaulted to `100`, and `BrnDialogRef.close()` deferred the underlying CDK close/dispose by that timeout. That was intended to keep `data-state="closed"` content mounted long enough for close animations.

That delay was inherited too broadly:

- `BrnPopover` extended `BrnDialog`.
- Select, combobox, autocomplete, date-picker, and popover sat on the popover/dialog path.
- Several of those Helm components do not ship a `data-state="closed"` exit animation.
- Result: after Escape or outside click, the component state could report closed while the portal stayed mounted until the timer elapsed, making non-animated controls feel sluggish.

#1466 also showed the other side of the problem: for animated dialogs, a fixed 100ms JS timer racing a 100ms CSS animation can tear down too early under CPU pressure. Ashley's measurements in that discussion found intermittent abrupt teardown under 6x CPU throttling. So the durable fix is not simply `closeDelay: 0`, nor is it keeping a magic timer everywhere. The lifecycle needs to follow primitive boundaries and the actual rendered animations.

## What Changed

- Added/solidified `BrnOverlay` as the CDK Overlay/Portal-backed primitive for non-modal floating portal UI.
- Changed `BrnPopover` to build on `BrnOverlay` instead of `BrnDialog`.
- Kept `BrnDialog` backed by CDK Dialog for modal/dialog semantics such as role, aria, focus handling, backdrop behavior, and restore focus.
- Removed `closeDelay` from the public dialog/popover/select/combobox/autocomplete surface.
- Reworked close teardown so content gets `data-state="closed"`, then disposes immediately when there are no active exit animations, or waits for the actual active animations when they exist.
- Uses CDK overlay/backdrop behavior for popover outside-pointer dismissal instead of trying to recreate low-level document dismissal logic by hand.
- Updated Helm wrappers, app examples, CLI/generated component templates, and API docs snapshots for the new public surface.

## Breaking Changes

This is intentionally marked as breaking because `closeDelay` is removed from the affected public APIs.

Consumers passing `closeDelay` to dialog, popover, select, combobox, autocomplete, or related close helpers should remove it. Exit animations should be expressed in CSS using the existing `data-state="closed"` state. Components with no exit animation now dispose immediately instead of waiting for a fixed timer.

Popover-backed controls are also no longer CDK Dialog instances. If a consumer needs modal dialog semantics, they should use dialog/alert-dialog/sheet rather than select/combobox/autocomplete/popover.

## Evidence From The Old Code

The problematic path on `main` was:

- `libs/brain/dialog/src/lib/brn-dialog-token.ts`: default `closeDelay: 100`.
- `libs/brain/dialog/src/lib/brn-dialog-ref.ts`: `close(result, delay)` delayed CDK disposal.
- `libs/brain/dialog/src/lib/brn-dialog.ts`: exposed `closeDelay` as an input.
- `libs/brain/popover/src/lib/brn-popover.ts`: popover inherited dialog behavior.
- Helm select/combobox/autocomplete/popover wrappers exposed the same inherited delay even for controls without close animations.

See the discussion in #1466 for the observed behavior and the rationale for moving beyond a narrow timer tweak.

## Validation

Local validation after rebasing onto the Vitest migration:

- `pnpm nx format:check --base=origin/main --head=HEAD`
- `pnpm nx affected -t lint --parallel=3 --base=origin/main --head=HEAD`
- `pnpm nx test helm`
- `pnpm nx affected -t test --parallel=3 --base=origin/main --head=HEAD`

Earlier validation on this branch also covered `brain`/`helm` builds and focused select e2e behavior around open, toggle, Escape, and outside-pointer dismissal.

## Summary by CodeRabbit

* **New Features**
  * Introduced an overlay-based system for floating panels, triggers, and references.
  * Added a Storybook story demonstrating dialogs launched from a context menu.
  * Select components now expose public `toggle()`/`close()` controls.
* **Bug Fixes**
  * Removed unintended initial auto-focus in combobox/popover/date-picker popovers.
  * Updated popover closing behavior and related accessibility expectations.
* **Refactor**
  * Migrated popover/combobox/select/date-picker (and related UI) to the overlay architecture.
* **Tests**
  * Updated Cypress and unit tests for the new overlay/accessibility behavior.
* **Documentation**
  * Added overlay README and packaging entrypoint.


PatrickLarocque:
Small note on the `libs/helm/utils/src/lib/hlm.spec.ts` change in the CI follow-up commit:

That edit is not part of the overlay/dialog behavior change. While validating this branch after rebasing onto the Vitest migration, `helm:test` exposed a latent timing-sensitive assertion in the existing `classes()` transition-suppression test.

The old test assumed that `await setTimeout(0)` meant "the Angular effect has run, but the queued `requestAnimationFrame` has not run yet." In Vitest browser mode / real Chromium, that ordering is not guaranteed. On my local browser run, rAF had already fired before the assertion, so `transition` had already been restored and the test failed with `expected '' to be 'none'`.

PR #1468 did run this exact Helm test successfully, so this was not an always-failing migration issue. It was a latent scheduling flake that happened to pass there and fail locally here. The fix stubs `window.requestAnimationFrame`, captures the callback, asserts the pre-rAF state, then invokes the callback manually and asserts the post-rAF state. Production behavior is unchanged; the test now controls the timing boundary it is explicitly asserting.

libs/brain/dialog/src/lib/brn-dialog.service.ts
```diff
@@ -4,234 +4,156 @@ import {
...
-						providers.push(...options.providers);
-					}
-				}
-
-				return providers;
+			autoFocus: mergedOptions.autoFocus,
+			ariaDescribedBy: mergedOptions.ariaDescribedBy ?? `brn-dialog-description-${dialogId}`,
+			ariaLabelledBy: mergedOptions.ariaLabelledBy ?? `brn-dialog-title-${dialogId}`,
+			ariaLabel: mergedOptions.ariaLabel,
+			ariaModal: mergedOptions.ariaModal,
```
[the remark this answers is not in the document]
PatrickLarocque: Thanks for calling this out. I added focused regression coverage for this exact launch path in `61d2ee60b`.

New coverage:
- `apps/ui-storybook/stories/dialog.stories.ts` adds `Dialog / Context Menu`, where a context-menu item opens an `hlm-dialog` via `[hlmDialogTriggerFor]`.
- `apps/ui-storybook-e2e/src/integration/dialog/dialog.cy.ts` visits `/iframe.html?id=dialog--context-menu`, right-clicks the context menu trigger, clicks the `Print` menu item, then asserts the dialog is visible by accessible name and that its body/action content renders.

Local verification: `pnpm nx e2e ui-storybook-e2e --spec=apps/ui-storybook-e2e/src/integration/dialog/dialog.cy.ts --skip-nx-cache` passed 8/8.

libs/brain/dialog/src/lib/brn-dialog-content.ts
```diff
@@ -7,28 +7,16 @@ import { BrnDialogRef } from './brn-dialog-ref';
...
-			const context = this.context();
-			if (!this._brnDialog || !context) return;
-			untracked(() => this._brnDialog?.setContext(context));
-		});
-		effect(() => {
-			if (!this._brnDialog) return;
-			const newClass = this.className();
-			untracked(() => this._brnDialog?.setPanelClass(newClass));
-		});
+		this._brnDialog?.registerContent(this._template, this.context, this.className);
```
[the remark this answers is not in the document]
PatrickLarocque: Fixed in `c7f2e0ce0`.

I kept the side effect at the primitive boundary rather than adding it to `BrnDialogContent`: `BrnDialog` now has a `_resolvedPanelClass` computed value matching the open-time precedence, and `_syncPanelClass()` pushes changes to the active `BrnDialogRef.setPanelClass(...)`. `BrnOverlay` received the same primitive-level panel-class sync for overlay-backed consumers.

libs/brain/dialog/src/lib/brn-dialog-overlay.ts
```diff
@@ -8,18 +8,16 @@ import { BrnDialog } from './brn-dialog';
...
-			const newClass = this.className();
-			untracked(() => this._brnDialog.setOverlayClass(newClass));
-		});
+		this._brnDialog.registerOverlayClass(this._resolvedClass);
 	}
 
-	setClassToCustomElement(newClass: string) {
-		this._brnDialog.setOverlayClass(newClass);
+	public setClassToCustomElement(newClass: string): void {
+		this._customClass.set(newClass);
```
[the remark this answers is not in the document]
PatrickLarocque: Fixed in `c7f2e0ce0`.

`BrnDialog` now resolves the registered/custom backdrop class through `_resolvedBackdropClass` and `_syncOverlayClass()` pushes changes to the active `BrnDialogRef.setOverlayClass(...)`. This keeps `BrnDialogOverlay` simple: it still records the custom class signal, while the owning dialog primitive handles propagation to the open CDK ref.

PatrickLarocque:
CodeRabbit follow-up: addressed the review feedback in `c7f2e0ce0`.

Covered items:
- normalized array entries in `cssClassesToArray` and added focused helper tests;
- preserved explicit `null` ARIA opt-outs in dialog options;
- guarded CDK point origins before calling `contains(...)`;
- enforced top-of-stack dismissal for backdrop events;
- returned fresh overlay default option objects to avoid shared mutable defaults;
- restored live panel/backdrop class propagation at the dialog/overlay primitive boundary.

Validation after the commit:
- `pnpm nx build brain` passed;
- `pnpm nx lint brain` passed with the existing `resolve-value-label.ts` warnings;
- `pnpm nx test brain` passed, 35 files / 289 tests;
- focused dialog Cypress spec passed, 8/8, against Storybook.

All CodeRabbit inline threads are resolved.

PatrickLarocque:
Linking the broader lifecycle/design issue that came out of the `closeDelay` investigation: #1473

That issue captures the inherited delay problem for popover-based controls, the JS-duration-to-CSS-duration coupling, and the possible CDK Dialog vs CDK Overlay primitive boundary.

libs/brain/overlay/src/lib/brn-overlay.ts
```diff
@@ -0,0 +1,258 @@
...
+	protected getDefaultOptions(): BrnOverlayDefaultOptions {
+		return injectBrnOverlayDefaultOptions();
+	}
+
+	public open(): void {
+		if (!this._content) return;
+
+		const currentRef = this._overlayRef();
+		if (currentRef) {
+			currentRef.reopen();
```
ashley-hunter, line 143: Select doesn't close on selection: open → ↓ → Enter commits the value but the listbox stays open (`origin/main` closes correctly). Looks like Enter re-activates the trigger → `reopen()` here. Could we skip trigger activation while `phase === 'closing'`?

libs/brain/overlay/src/lib/brn-overlay.service.ts
```diff
@@ -0,0 +1,165 @@
...
+		}
+		return providers;
+	}
+
+	private _configureElement(overlayRef: OverlayRef, options: BrnOverlayOptions): void {
+		const element = overlayRef.overlayElement;
+		element.id = options.id;
+		if (options.role) {
+			element.tabIndex = -1;
+			element.setAttribute('role', options.role);
```
ashley-hunter, line 161: No `autoFocus` equivalent here, so the popover's form fields no longer get focused on open (baseline used CDK Dialog `autoFocus: 'first-tabbable'`). `aria-activedescendant` also reads null while navigating the select.

libs/brain/popover/src/lib/brn-popover.token.ts
```diff
@@ -23,3 +24,28 @@ export function provideBrnPopoverConfig(config: Partial<BrnPopoverConfig>): Valu
 export function injectBrnPopoverConfig(): BrnPopoverConfig {
 	return inject(BrnPopoverConfigToken, { optional: true }) ?? defaultConfig;
 }
+
+export const BRN_POPOVER_OVERLAY_DEFAULT_OPTIONS: BrnOverlayDefaultOptions = {
+	...overlayDefaultOptions,
+	backdropClass: 'cdk-overlay-transparent-backdrop',
+	closeOnOutsidePointerEvents: true,
+	hasBackdrop: true,
```
ashley-hunter: Intentional? `hasBackdrop: true` + the transparent backdrop eats outside clicks and makes the popover effectively modal, vs the non-modal default (was `hasBackdrop: false`).

libs/brain/overlay/src/lib/brn-overlay.service.ts
```diff
@@ -0,0 +1,165 @@
...
+			.subscribe(({ event, reason }) => this._requestDismissal(ref, overlayRef, reason, event));
+	}
+
+	private _requestDismissal<Result>(
+		ref: BrnOverlayRef<Result>,
+		overlayRef: OverlayRef,
+		reason: BrnOverlayDismissReason,
+		event: MouseEvent | KeyboardEvent,
+	): void {
+		if (this._overlayStack.at(-1) !== overlayRef) return;
```
ashley-hunter, line 188: Topmost-only dismiss drops the old `isNested` handling — worth checking nested overlays still dismiss the intended layer.

libs/brain/overlay/src/lib/brn-overlay.service.ts
```diff
@@ -0,0 +1,165 @@
...
+	public open<OverlayContext, OverlayResult = unknown>(
+		content: ComponentType<unknown> | TemplateRef<unknown>,
+		vcr?: ViewContainerRef,
+		context?: OverlayContext,
+		options?: Partial<BrnOverlayOptions>,
+	): BrnOverlayRef<OverlayResult> {
+		const mergedOptions: BrnOverlayOptions = {
+			...this._defaultOptions,
+			...options,
+			id: options?.id ?? `brn-overlay-${overlaySequence + 1}`,
```
ashley-hunter: `brn-overlay-${overlaySequence + 1}` here but the ref uses `++overlaySequence` later, so the duplicate-id guard reads a stale counter.

ashley-hunter (commented):
Thanks for the contribution! Looks great, just noticed a few issues.

ashley-hunter (changes requested):

libs/brain/popover/src/lib/brn-popover.token.ts
PatrickLarocque: My bad, went through a few implementations and was fighting outside clicks at one point. Reverted back to the original `hasBackdrop: false`, i also added a test that confirms that the backdrop honors the configuration.

ashley-hunter:
@MerlinMoos i pushed a few updates and resigned the conflicts, if you get a chance it would be great to get another pair of eyes on this 😊 thanks

PatrickLarocque:
> @MerlinMoos i pushed a few updates and resigned the conflicts, if you get a chance it would be great to get another pair of eyes on this 😊 thanks

Thanks so much. Sorry for not getting around to updating this sooner. I’ve been busy with work, and I also realized there was a good reason brnDialog was doing so much of the heavy lifting around accessibility before my attempt to split dialog and overlay apart.

The CDK Overlay is more of a positioning/portal primitive than an accessibility primitive. So while moving components like autocomplete, combobox, select, and popover away from dialog feels more semantically correct, it also means we lose some accessibility behavior that came “for free” from being backed by CDK Dialog.
That became clearer when I started looking at the stacking-context issue from your last comment. Handling dismissal, focus, aria semantics, and nested overlay behavior correctly is a lot easier to get subtly wrong than I initially appreciated.
Longer term, I think the right direction may be to lean into the newer Angular Aria primitives for form-style and floating components like autocomplete, combobox, select, menu/menubar, etc., rather than trying to hand-roll an accessible abstraction on top of CDK Overlay ourselves.

That said, I’m not sure how you all feel about introducing @angular/aria at this point, especially given that it is still stabilizing around Angular v22.

PR #1472, opened 2026-06-13
fix(accordion): measure first element child in brn-accordion-content

liferays:
## PR Checklist

- [x] The commit message follows our guidelines: https://github.com/spartan-ng/spartan/blob/main/CONTRIBUTING.md#-commit-message-guidelines

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [x] accordion

`BrnAccordionContent` measures its content in an `afterNextRender` hook to drive the open/close transition:

```ts
const content = this._elementRef.nativeElement.firstChild as HTMLElement | null;
if (!content) return false;
const { width, height } = measureDimensions(content, this._config.measurementDisplay);
```

`firstChild` returns the first **node**, which is a `Text` node whenever the projected content begins with a text node — a leading interpolation, or (very commonly) any consuming component compiled with `preserveWhitespaces: true`, where the indentation between `<brn-accordion-content>` and its first child element survives as a whitespace `Text` node.

`measureDimensions` then reads `elementToMeasure.style.height` on that `Text` node, and `Text.style` is `undefined`, so it throws:

```
ERROR TypeError: Cannot read properties of undefined (reading 'height')
    at measureDimensions
    at BrnAccordionContent._measureAndSetDimensions
```

The error surfaces through the `ErrorHandler` (so the app keeps running), but the content is never measured/sized and the console fills with the error for every opened accordion.

## What is the new behavior?

Measure `firstElementChild` instead of `firstChild`. It skips leading text/comment nodes and returns the first **element** — the node that actually needs measuring — so a leading whitespace/text node no longer crashes the hook. When there is no element child, the existing `if (!content) return false;` guard still falls back to the `IntersectionObserver` path.

A regression test renders an opened `brn-accordion-content` whose projected content starts with a text node and asserts the render hook does not throw.

- [ ] Yes
- [x] No

## Other information

- Reproduces in any app where a consuming component uses `preserveWhitespaces: true` (Angular's non-default but legitimate setting) and projects an element into `<brn-accordion-content>` on its own line.
- Verified against the latest `main`; the regression test fails before the change and passes after, and the full `brain` test suite stays green.
- I searched open issues/PRs and did not find an existing report.

## Bug Fixes

* Fixed an accordion content dimension measurement edge case where leading whitespace/text nodes could cause incorrect calculations or failures. This improves reliability when content is projected or contains preserve-whitespace output, ensuring smoother accordion sizing and consistent rendering across supported scenarios.

### DIFF
--- a/libs/brain/accordion/src/lib/brn-accordion-content.ts
+++ b/libs/brain/accordion/src/lib/brn-accordion-content.ts
@@ -61,7 +61,10 @@ export class BrnAccordionContent {
 	}
 
 	private _measureAndSetDimensions(): boolean {
-		const content = this._elementRef.nativeElement.firstChild as HTMLElement | null;
+		// Measure the first ELEMENT child, not the first node: a leading text node
+		// (e.g. projected content compiled with `preserveWhitespaces: true`) has no
+		// `.style`, so measuring it throws "Cannot read properties of undefined".
+		const content = this._elementRef.nativeElement.firstElementChild as HTMLElement | null;
 		if (!content) return false;
 
 		const { width, height } = measureDimensions(content, this._config.measurementDisplay);
--- a/libs/brain/accordion/src/lib/brn-accordion.spec.ts
+++ b/libs/brain/accordion/src/lib/brn-accordion.spec.ts
@@ -94,12 +94,6 @@ class AccordionBlurDuringRenderSpec extends ProbeHost {}
 })
 class AccordionTriggerFocusDuringRenderSpec extends ProbeHost {}
 
-// brn-accordion-content measured `firstChild`, which is a Text node whenever the
-// projected content begins with a text node - the leading text below, or e.g. a
-// template compiled with `preserveWhitespaces: true` where the indentation before
-// the inner element survives as a Text node. measureDimensions then read `.style`
-// off that Text node and threw "Cannot read properties of undefined (reading
-// 'height')".
 @Component({
 	imports: [BrnAccordion, BrnAccordionItem, BrnAccordionTrigger, BrnAccordionContent],
 	changeDetection: ChangeDetectionStrategy.OnPush,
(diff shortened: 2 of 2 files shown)


libs/brain/accordion/src/lib/brn-accordion.spec.ts
```diff
@@ -93,6 +94,31 @@ class AccordionBlurDuringRenderSpec extends ProbeHost {}
 })
 class AccordionTriggerFocusDuringRenderSpec extends ProbeHost {}
 
+// brn-accordion-content measured `firstChild`, which is a Text node whenever the
```
goetzrobin: seems like this is an AI generated comment that adds no value

libs/brain/accordion/src/lib/brn-accordion.spec.ts
```diff
@@ -126,4 +152,19 @@ describe('BrnAccordion', () => {
 
 		expect(ng0600Errors(errors)).toEqual([]);
 	});
+
+	// brn-accordion-content must measure its first ELEMENT child, not its first
```
goetzrobin: same here

libs/brain/accordion/src/lib/brn-accordion.spec.ts
```diff
@@ -126,4 +152,19 @@ describe('BrnAccordion', () => {
...
+	// node: a leading Text node (e.g. from preserveWhitespaces) has no `.style`, so
+	// measuring it threw "Cannot read properties of undefined (reading 'height')".
+	it('does not throw when the opened content has a leading text node', async () => {
+		const errors: unknown[] = [];
+		await render(AccordionContentLeadingTextNodeSpec, {
+			providers: [{ provide: ErrorHandler, useValue: { handleError: (error: unknown) => errors.push(error) } }],
+		});
+		// The dimension measuring runs in an afterNextRender hook that does not
+		// auto-run under jest - flush it so the regression actually exercises.
```
goetzrobin: we decided to move to vitest

libs/brain/accordion/src/lib/brn-accordion-content.ts
```diff
@@ -61,7 +61,10 @@ export class BrnAccordionContent {
...
 
 	private _measureAndSetDimensions(): boolean {
-		const content = this._elementRef.nativeElement.firstChild as HTMLElement | null;
+		// Measure the first ELEMENT child, not the first node: a leading text node
+		// (e.g. projected content compiled with `preserveWhitespaces: true`) has no
+		// `.style`, so measuring it throws "Cannot read properties of undefined".
+		const content = this._elementRef.nativeElement.firstElementChild as HTMLElement | null;
 		if (!content) return false;
 
 		const { width, height } = measureDimensions(content, this._config.measurementDisplay);
```
goetzrobin, line 70: we should probably also make measureDimensions be more defensive? is the measurement working even with the text and element?

goetzrobin (changes requested):

libs/brain/accordion/src/lib/brn-accordion-content.ts
liferays: @goetzrobin yeah good point on both.

the `firstElementChild` thing does only grab the first element, so with `leading text <div>Body</div>` the text height gets left out and the panel can end up a bit short. real content is usually one wrapper el so it's fine in practice, the test just makes it obvious.

probably cleaner to just measure the host (`this._elementRef.nativeElement`) like collapsible already does , never hits a text node and counts everything inside. only thing is it'd include the host's own padding if there is any.

and agree on making `measureDimensions` defensive, something like:
```ts
if (!(elementToMeasure instanceof HTMLElement)) return { width: 0, height: 0 };
```
though i'd still measure the host too, otherwise a silent 0/0 could hide a wrong-node bug.

happy to switch to host + add the guard, or just keep firstElementChild w/ the guard . lmk which you prefer

goetzrobin:
Closing in favor of https://github.com/spartan-ng/spartan/pull/1523

Issue #1373, opened 2026-06-01
Slow or unresponsive spartan docs with open Devtools


marcjulian:
### Please provide the environment you discovered this bug in.

With open devtools, the spartan.ng pages react very slow or sometimes gets unresponsive during local development.
Also viewing an element to check the applied styles or data attribute sometimes doesn't refresh to show the selected element styles.

### Which area/package is the issue in?

Docs

### Description

Effects local development of spartan.ng while having the devtools open.

### Please provide the exception or error you saw

```true

```

### Other information

Effected in Chrome devtools

No issue with firefox devtools

### I would be willing to submit a PR to fix this issue

- [ ] Yes
- [ ] No

marcjulian:
Do you have the same experience during development? @MerlinMoos @toniskobic @elite-benni @thatsamsonkid @ashley-hunter @goetzrobin

I am not sure if it's related to a recent change like select rewrite or style support (loading json for manual copying of components), or something else.

MerlinMoos:
hey @marcjulian I have the same experience, but I really don't know where this is coming from. I do not have experience with analog

marcjulian:
Do you use also Chrome? Do you have Angular DevTools installed?

ashley-hunter:
What components are on the page? And is this in an Angular or Analog app?

marcjulian:
Its effecting the spartan.ng app - Analog app. I haven't noticed this in other Angular apps.

ashley-hunter:
Ah! Ok, I thought that was in an application consuming our components rather than our docs, makes sense, I will look into it when I get some time and see if anything stands out.

marcjulian:
Sorry, yes i meant the spartan docs app 👍 thank you for checking it out!

ashley-hunter:
I have raised a pr that makes some improvements, but the main issue is styles, and with dev tools open some of the style caching mechanisms in chrome are disabled, so in that sense there is not a huge amount we can do except try and reduce the number of elements in the dom and optimisations like that.

marcjulian:
@ashley-hunter thanks for looking into that issue! I will have a look if its also related to the different themes, which are always loaded. Perhaps they could be lazy loaded when someone changes to a different theme.

marcjulian:
I think it might be related to the style-vega.css. When I commented it out and switch to a different style for a component I can easily view the styles in devtools and also when closing the devtools the page resizes fast.

[...]

PR #1480, opened 2026-06-15
perf: scope vega as unscoped baseline to cut style recalculation

### LINKED ISSUE #1343: [BUG] Sidebar (push variant) expand/collapse animation is janky and unresponsive on desktop compared to shadcn/ui

Latest spartan ng version

sidebar

### Description

The sidebar push animation (when the sidebar expands/collapses and pushes the main content) is noticeably laggy and unresponsive on desktop, especially on lower-end or throttled devices. The animation drops frames and feels sluggish compared to the equivalent implementation in shadcn/ui, which remains smooth under the same conditions.
To Reproduce
	1.	Open a page using HlmSidebarComponent in push mode on desktop
	2.	Toggle the sidebar open/close
	3.	Observe the expand/collapse animation
To reproduce on any machine:
	1.	Open Chrome DevTools → Performance tab → CPU throttling → set to 4x or 6x slowdown
	2.	Toggle the sidebar repeatedly
	3.	Compare with the shadcn/ui sidebar demo under the same throttling
Expected behavior
The push animation should be smooth and compositor-friendly (GPU-accelerated), similar to the shadcn/ui implementation which uses transform: translateX() / CSS custom properties with transition on the content wrapper, avoiding layout thrashing.
Actual behavior
The animation causes visible frame drops and feels unresponsive. Likely caused by animating layout-triggering properties (e.g. width, margin-left) instead of transform, forcing the browser to perform expensive layout/paint passes on every frame.
Suspected root cause
Animating width or margin/padding on the sidebar or content wrapper triggers layout recalculation on every frame, which is expensive. The shadcn/ui implementation avoids this by using CSS variables + transform so only the composite layer is affected, keeping the animation entirely on the GPU.

```true

```

### Other information

_No response_

- [ ] Yes
- [ ] No

goetzrobin:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Others

- [x] repo

The docs app loads all six theme stylesheets at once, and `style-vega.css` wraps
every rule in
`.style-vega :where(:not([class~='not-style-vega'], [class~='not-style-vega'] *))`.
That `:not(... *)` forces the browser to walk the full ancestor chain of every
candidate element, for every vega rule, on every style recalculation. Toggling the
sidebar invalidates a large subtree and triggers a ~108ms style recalc, which makes
the push-variant animation janky. With Chrome DevTools open (which disables Blink's
style-sharing cache) it degrades further into an unresponsive page.

Relates to #1373
Closes #1343

## What is the new behavior?

- `style-vega.css` is scoped to `*`, so vega becomes the global, lowest-specificity
  baseline (`*.spartan-x` collapses to `.spartan-x`). No `:not`, no ancestor walk.
- Removed the `not-style-vega` island machinery from the docs app (body class,
  code-preview directive, style service, sidebar-preview). Other themes still
  override vega through their higher-specificity `.style-THEME .spartan-x` rules.
- Added app-scoped overrides in `apps/app/src/styles.css` for the five classes vega
  defines but lyra doesn't (button-group orientation horizontal/vertical, carousel
  previous/next, dialog-footer), so vega's baseline no longer leaks into
  `.style-lyra` previews. These live in the docs app, not the registry, so the CLI
  never copies them into consumer components.

Measured: worst-case style recalc on sidebar toggle drops from ~108ms to ~20ms. A
Firefox profile of the same page shows restyle at 0.089ms avg with 99% of frames
hitting 60fps, confirming the page itself is healthy. The residual Chrome-DevTools
slowness tracked in #1373 is dominated by raw dev-mode CSS volume (Tailwind's
un-tree-shaken utilities) rather than this selector, so it is only partially
addressed here.

- [ ] Yes
- [x] No

The vega wrapper change is invisible to consumers: the CLI inlines the real Tailwind
classes from each `@apply` and strips the `spartan-*` wrapper, so generated
components are unchanged.

## Other information

CSS/style-only change; no unit tests added — recalc cost is verified via Chrome and
Firefox performance profiles rather than assertions. `pnpm run build` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed theme “island” CSS leakage so multiple themes loaded together stay properly isolated.
  * Improved theme switching so the active theme applies cleanly without legacy “not-style” handling.
  * Corrected Arabic/Hebrew font fallback values.
* **New Features**
  * Updated the default theme to **nova**.
* **Refactor**
  * Reworked theme stylesheet layer ordering so island theme skins reliably take precedence.
  * Reduced CSS specificity for button-group corner rounding by removing forced priority rules across themes.
* **Documentation**
  * Added a docs-app-only “leak patch” section for targeted component spacing/rounding adjustments.

### DIFF
--- a/apps/app/index.html
+++ b/apps/app/index.html
@@ -36,7 +36,7 @@
 		</script>
 	</head>
 	<body
-		class="text-foreground group/body overscroll-none font-sans antialiased [--footer-height:calc(var(--spacing)*14)] [--header-height:calc(var(--spacing)*14)] [--stable-height:78.75px] xl:[--footer-height:calc(var(--spacing)*24)]"
+		class="style-nova text-foreground group/body overscroll-none font-sans antialiased [--footer-height:calc(var(--spacing)*14)] [--header-height:calc(var(--spacing)*14)] [--stable-height:78.75px] xl:[--footer-height:calc(var(--spacing)*24)]"
 	>
 		<script>
 			if (localStorage.theme !== 'default' && document && document.body)
--- a/apps/app/index.html
+++ b/apps/app/index.html
@@ -36,7 +36,7 @@
 		</script>
 	</head>
 	<body
-		class="text-foreground group/body style-vega overscroll-none font-sans antialiased [--footer-height:calc(var(--spacing)*14)] [--header-height:calc(var(--spacing)*14)] [--stable-height:78.75px] xl:[--footer-height:calc(var(--spacing)*24)]"
+		class="text-foreground group/body overscroll-none font-sans antialiased [--footer-height:calc(var(--spacing)*14)] [--header-height:calc(var(--spacing)*14)] [--stable-height:78.75px] xl:[--footer-height:calc(var(--spacing)*24)]"
 	>
 		<script>
 			if (localStorage.theme !== 'default' && document && document.body)
--- a/libs/registry/src/styles/style-vega.css
+++ b/libs/registry/src/styles/style-vega.css
@@ -1,4 +1,4 @@
-.style-vega :where(:not([class~='not-style-vega'], [class~='not-style-vega'] *)) {
+* {
 	&.spartan-rtl-flip {
 		@apply rtl:rotate-180;
 	}
--- a/libs/registry/src/styles/style-vega.css
+++ b/libs/registry/src/styles/style-vega.css
@@ -245,11 +245,11 @@
 	}
 
 	.spartan-button-group-orientation-horizontal {
-		@apply [&>[data-slot]:not(:has(~[data-slot]))]:rounded-e-md!;
+		@apply [&>[data-slot]:not(:has(~[data-slot]))]:rounded-e-md;
 	}
 
 	.spartan-button-group-orientation-vertical {
-		@apply [&>[data-slot]:not(:has(~[data-slot]))]:rounded-b-md!;
+		@apply [&>[data-slot]:not(:has(~[data-slot]))]:rounded-b-md;
 	}
 
 	.spartan-button-group-text {
--- a/libs/registry/src/styles/style-vega.css
+++ b/libs/registry/src/styles/style-vega.css
@@ -1,1245 +1,1399 @@
-* {
-	&.spartan-rtl-flip {
+.style-vega {
+	.spartan-rtl-flip {
 		@apply rtl:rotate-180;
 	}
 
 	/* MARK: Accordion */
-	&.spartan-accordion-item {
+	.spartan-accordion-item {
 		@apply not-last:border-b;
 	}
 
-	&.spartan-accordion-trigger {
+	.spartan-accordion-trigger {
 		@apply focus-visible:ring-ring/50 focus-visible:border-ring focus-visible:after:border-ring **:data-[slot=accordion-trigger-icon]:text-muted-foreground! rounded-md py-4 text-start text-sm font-medium hover:underline focus-visible:ring-3 **:data-[slot=accordion-trigger-icon]:ms-auto **:data-[slot=accordion-trigger-icon]:text-[length:--spacing(4)];
 	}
 
-	&.spartan-accordion-content {
+	.spartan-accordion-content {
 		@apply text-sm;
 	}
 
-	&.spartan-accordion-content-inner {
+	.spartan-accordion-content-inner {
 		@apply pt-0 pb-4;
 	}
 
 	/* MARK: Alert Dialog */
-	&.spartan-alert-dialog-overlay {
+	.spartan-alert-dialog-overlay {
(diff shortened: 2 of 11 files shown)


libs/registry/src/styles/style-vega.css
```diff
@@ -1,4 +1,4 @@
-.style-vega :where(:not([class~='not-style-vega'], [class~='not-style-vega'] *)) {
+* {
 	&.spartan-rtl-flip {
```
marcjulian: Remove `&` from each class in style-vega

```suggestion
	.spartan-rtl-flip {
```

libs/registry/src/styles/style-vega.css
```diff
@@ -1,4 +1,4 @@
-.style-vega :where(:not([class~='not-style-vega'], [class~='not-style-vega'] *)) {
+* {
```
marcjulian: Keeping `.style-vega` here instead of `*` improves chrome devtools style selection

```suggestion
.style-vega {
```

apps/app/index.html
```diff
@@ -36,7 +36,7 @@
 		</script>
 	</head>
 	<body
-		class="text-foreground group/body style-vega overscroll-none font-sans antialiased [--footer-height:calc(var(--spacing)*14)] [--header-height:calc(var(--spacing)*14)] [--stable-height:78.75px] xl:[--footer-height:calc(var(--spacing)*24)]"
```
marcjulian, line 39: Still needs style-vega if you keep it in the `.style-vega` as scope name

marcjulian (changes requested):
Tested it and found maybe a small improvement.

apps/app/index.html
goetzrobin: so we keep style-vega and also keep it instead of * as the wrapper?

marcjulian: Yes thats the idea, for me the devtools in chrome resolve faster with .style-vega instead of * to match the styles. Perhaps you can give it a try

goetzrobin: Thanks for pointing that out. I was actually trying to understand whats going on exactly with either version.
So I had Claude explain to me the cascade properly. 

First, something I learned: CSS specificity is written (IDs, classes, elements) and compared column by column from the left - so (0,2,0) is two classes, (0,1,0) is one class, and a higher count in an earlier column wins (two classes beats one; a single ID would beat any number of classes).

**What the `*` approach does (current PR):** 
vega compiles to a bare `.spartan-X` - one class, (0,1,0). The other themes are `.style-THEME .spartan-X` - two classes, (0,2,0) - so vega loses to all of them on specificity alone. That makes the baseline order-independent: shuffle the imports however and vega still sits under everything. The five lyra leak-patches are doubly safe here too - they're unlayered (so they win at the layer step), and even ignoring layers they're `.style-lyra .spartan-X` (0,2,0) vs vega's (0,1,0), so they'd win on specificity anyway.

**What `.style-vega` does (your suggestion):** 
vega becomes `.style-vega .spartan-X` - two classes, (0,2,0)  - so it now *ties* the other themes instead of sitting under them. Specificity can't break a tie, so it falls to source order: it works because `style-vega.css` is imported first, and nothing in the selector enforces that. It also squeezes the patches - vega's rule is now (0,2,0), the same as the patch, so their specificity margin is gone. The patch still wins, but *only* because it's unlayered; move it into `@layer base` to tidy up and it'd tie vega and fall back to source order.

I think we can still go the .style-vega way, we just need to make sure to document how its working, because I think its not obvious from the code

What do you think @marcjulian?

marcjulian: Yes I think we should go with `.style-vega` as scope which improves the style recalculation in chrome devtools. We can definitely add a comment/document that style-vega needs to come first because its the currently the default style.

goetzrobin:
@MerlinMoos this should be `nova` instead of `vega` now right?

apps/app/index.html
goetzrobin: And this is only our docs app, because our CLI replaces the classes during copy, so as long as we know what's going on that's good!

goetzrobin: I wonder if we can also use layers instead?

MerlinMoos:
> @MerlinMoos this should be `nova` instead of `vega` now right?

yes please

apps/app/index.html
marcjulian: perhaps layers can work for that too. Yes, the style files are only for our docs to switch between the styles and the CLI replaces the classes at CLI build time. The end developer never have any of the spartan classes.

marcjulian:
@goetzrobin can you resolve the conflicts?

marcjulian (changes requested):
Styles leak into the button group. Try out Maia and Luma for the Button Group and the rounding is off at the end

This branch

Image

Image

---

Main

Image

Image

goetzrobin:
@marcjulian leak should be fixed now! thanks for catching that!

MerlinMoos:
@goetzrobin LGTM the developer tools are finally working again without lagging

PR #1492, opened 2026-06-17
fix(radio-group): restore native tab behavior

### LINKED ISSUE #1491: Radio group cannot be tabbed into after clearing the selected value
## Description

When a radio group is reset to a state where no option is selected, keyboard users can no longer tab into the group. The enabled radio inputs all receive `tabindex="-1"`, so the group has no reachable tab stop.

## Steps to reproduce

1. Open the Radio Group default Storybook story.
2. Click `Reset` so the current value becomes `none`.
3. Try to tab back into the radio group with the keyboard.

## Current behavior

After reset, every radio input in the group has `tabindex="-1"`, including enabled options. Because there is no selected radio and no enabled radio with `tabindex="0"`, focus skips the radio group instead of entering it.

## Expected behavior

Keyboard users should still be able to tab into the radio group when no value is selected. Per the WAI-ARIA radio group keyboard guidance, if no radio is checked, focus should move into the group rather than the group becoming unreachable. For a group made from native same-name `input[type="radio"]` controls, the browser already provides a single tab stop and focuses an enabled radio when entering the group.

## Impact

This blocks keyboard-only users from reaching and selecting a radio option after a form reset or any other state transition that clears the group value. It also makes validation/reset flows especially hard to recover from without a pointer device.

PatrickLarocque:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] autocomplete
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] button-group
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] empty
- [ ] dropdown-menu
- [ ] field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-group
- [ ] input-otp
- [ ] item
- [ ] kbd
- [ ] label
- [ ] menubar
- [ ] native-select
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [x] radio-group
- [ ] resizable
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] sidebar
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli
- [ ] mcp

Closes #1491

When a radio group is cleared to no selected value, such as the default Storybook example after clicking `Reset`, every radio input receives `tabindex="-1"`. That leaves the group with no tabbable radio, so keyboard users cannot tab back into the group to choose a new option.

The root cause is the component's custom `_tabIndex` computation. It treats any group value other than `undefined` as meaning there is a selected radio. Resetting the example sets the value to `null`, which means no radio is checked, but the logic still considers the group selected and assigns `-1` to every unchecked radio.

## What is the new behavior?

`BrnRadio` no longer overrides the native radio input's tab order with a custom `[tabIndex]` binding.

Because Spartan radio groups render real same-name `<input type="radio">` controls, the browser already provides the native radio-group tab behavior: tabbing into a group focuses the checked radio, and when none is checked the group still has an enabled radio in the tab order. Removing the custom roving-tabindex logic restores that native behavior and avoids duplicating it incorrectly in Angular state.

The regression tests now verify the keyboard behavior directly with `userEvent.tab()`:

- a selected group can be reset to no selected value and still receive focus on the first radio
- a group whose first option is disabled focuses the first enabled radio when no enabled radio is selected
- tabbing again leaves the group instead of trapping focus

- [ ] Yes
- [x] No

## Other information

No docs were updated because this fixes the implementation to match the expected native keyboard behavior for the existing component API.

Validation performed:

- `pnpm vitest run libs/brain/radio-group/src/lib/brn-radio.spec.ts --config libs/brain/vite.config.mts`
- `pnpm nx lint brain` passes with existing unrelated warnings
- `pnpm nx build brain`
- `git diff --check`

Manual verification:

- Confirmed `main` reproduces the bug in Storybook: after `Reset`, all radios have `tabindex="-1"`
- Confirmed the fix branch removes explicit `tabindex` attributes and restores keyboard access after reset

## Summary by CodeRabbit

* **Tests**
  * Added keyboard-navigation focus tests for the radio group component, including cases where the selected value is cleared and where the first option is disabled.
* **Refactor**
  * Adjusted radio tab focus behavior by changing how focusability/tab order is determined within the radio component, aligning tabbing with the expected keyboard navigation flow.

### DIFF
--- a/libs/brain/radio-group/src/lib/brn-radio.spec.ts
+++ b/libs/brain/radio-group/src/lib/brn-radio.spec.ts
@@ -1,8 +1,77 @@
 import { FormControl, FormGroup, ReactiveFormsModule } from '@angular/forms';
 import { render } from '@testing-library/angular';
+import userEvent from '@testing-library/user-event';
 import { BrnRadioGroupImports } from '../index';
 
 describe('BrnRadioComponent', () => {
+	it('should focus the first radio when tabbing into a group reset to no selected value', async () => {
+		const user = userEvent.setup();
+		const { getAllByRole, getByText, rerender } = await render(
+			`
+			<button type="button">before</button>
+			<div brnRadioGroup [value]="value">
+				<brn-radio value="16.1.4">16.1.4</brn-radio>
+				<brn-radio value="16.0.0">16.0.0</brn-radio>
+				<brn-radio value="15.3.0">15.3.0</brn-radio>
+			</div>
+			<button type="button">after</button>
+			`,
+			{
+				imports: [BrnRadioGroupImports],
+				componentProperties: {
+					value: '16.0.0',
+				},
+			},
+		);
+
+		const radioButtons = getAllByRole('radio') as HTMLInputElement[];
+		getByText('before').focus();
+		await user.tab();
+		expect(radioButtons[1]).toHaveFocus();
+
+		await rerender({
+			componentProperties: {
+				value: null,
+			},
+		});
+
+		getByText('before').focus();
+		await user.tab();
+		expect(radioButtons[0]).toHaveFocus();
+
+		await user.tab();
+		expect(getByText('after')).toHaveFocus();
+	});
+
+	it('should focus the first enabled radio when tabbing into a group with no selected enabled radio', async () => {
+		const user = userEvent.setup();
+		const { getAllByRole, getByText } = await render(
+			`
+			<button type="button">before</button>
+			<div brnRadioGroup [value]="value">
+				<brn-radio disabled value="16.1.4">16.1.4</brn-radio>
+				<brn-radio value="16.0.0">16.0.0</brn-radio>
+				<brn-radio value="15.3.0">15.3.0</brn-radio>
+			</div>
+			<button type="button">after</button>
+			`,
+			{
+				imports: [BrnRadioGroupImports],
+				componentProperties: {
+					value: '',
+				},
+			},
+		);
+
+		const radioButtons = getAllByRole('radio') as HTMLInputElement[];
+		getByText('before').focus();
+		await user.tab();
+		expect(radioButtons[1]).toHaveFocus();
+
+		await user.tab();
+		expect(getByText('after')).toHaveFocus();
+	});
+
 	it('should disable the radio button when disabled is true (reactive forms)', async () => {
 		const form = new FormGroup({
 			radioGroup: new FormControl('16.1.4'),
--- a/libs/brain/radio-group/src/lib/brn-radio.spec.ts
+++ b/libs/brain/radio-group/src/lib/brn-radio.spec.ts
@@ -43,9 +43,9 @@ describe('BrnRadioComponent', () => {
 		expect(getByText('after')).toHaveFocus();
 	});
 
-	it('should focus the first enabled radio when tabbing into a group with no selected enabled radio', async () => {
+	it('should focus the first enabled radio when tabbing into a reset group with a disabled first radio', async () => {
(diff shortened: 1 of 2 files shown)


libs/brain/radio-group/src/lib/brn-radio.spec.ts
```diff
@@ -1,8 +1,77 @@
...
+
+		getByText('before').focus();
+		await user.tab();
+		expect(radioButtons[0]).toHaveFocus();
+
+		await user.tab();
+		expect(getByText('after')).toHaveFocus();
+	});
+
+	it('should focus the first enabled radio when tabbing into a group with no selected enabled radio', async () => {
```
goetzrobin: @PatrickLarocque should we add a test case here also that checks resetting the value and then tapping into it again moves to 16.0.0 also?

libs/brain/radio-group/src/lib/brn-radio.ts
```diff
@@ -94,18 +93,6 @@ export class BrnRadio<T = unknown> implements OnDestroy {
 	 */
 	protected readonly _checked = computed(() => this._radioGroup.value() === this.value());
 
-	protected readonly _tabIndex = computed(() => {
```
goetzrobin, line 97: can you check: https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex and how it ties into this

goetzrobin: Might not be needed, but just so you know why it exists in the first place

PatrickLarocque: Hey @goetzrobin , thanks for looking at this pr. I did look into the WAI-ARIA APG roving `tabindex` spec, and I know that this explains why this code existed in the first place. The deleted logic was trying to keep the radio group to a single tab stop, with only the currently checked/entry radio having `tabindex="0"`.

The edge case it missed was a reset/no-value state. Since it treated any group value other than `undefined` as “there is a selected radio”, resetting the value to `null` meant every unchecked radio could end up with `tabindex="-1"`, making the group unreachable by tab.

I originally patched the `_tabIndex` function directly, but after testing it more, I realized this component is already backed by native `<input type="radio">` elements with a shared `name`. In that shape, the browser already handles the radio-group tab behavior: tab enters on the checked radio, or on the first enabled radio when nothing is checked, and the next tab exits the group.

So the PR removes our roving `tabindex` implementation and lets the native radio behavior take over. I also added the extra disabled-first reset test so we cover the `16.0.0` case you mentioned.

libs/brain/radio-group/src/lib/brn-radio.spec.ts
PatrickLarocque: Yup, good call, added this test!

PR #1493, opened 2026-06-17
fix(select): select option with a falsy value on enter keypress

Ban117:
`selectActiveItem()` guarded the commit with a truthiness check (`if (value)`), so an option whose value is falsy (`false`, `0`, `''`) was dropped: pressing Enter just closed the panel without selecting. It now checks `value !== null && value !== undefined`, so a defined falsy value is committed while a missing active item still closes. Applies to both `BrnSelect` and `BrnSelectMultiple`.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [x] select

In a single or multiple `hlm-select` whose options carry a falsy value (e.g. a boolean `[value]="false"`, `0`, or an empty string), that option cannot be selected with the keyboard.

Steps to reproduce:
1. Open the select.
2. Arrow to the option whose value is falsy (e.g. `false`).
3. Press Enter.

Expected: the value is committed. Actual: the panel closes and the value is unchanged. Selecting the same option with the **mouse** works, and options with truthy values select fine with Enter — so it's specific to keyboard selection of falsy values.

Root cause: `BrnSelect.selectActiveItem()` (and `BrnSelectMultiple.selectActiveItem()`) guard the commit with `if (value)`, so a defined-but-falsy active value falls into the `else` branch and only closes the panel.

## What is the new behavior?

Options with a defined falsy value are now committed on Enter. The guard uses `value !== null && value !== undefined`, so `false` / `0` / `''` select correctly, while a genuinely missing active item still closes the panel. Added a `Select / Falsy Value` Storybook story and a Cypress e2e test covering the regression.

- [ ] Yes
- [x] No

## Other information

Reproduced and verified against the new `select--falsy-value` story.

## Release Notes

* **Bug Fixes**
  * Fixed select component to properly handle and allow selection of falsy values (such as 0, false, or empty string).

* **Tests**
  * Added test coverage validating select functionality with falsy values.

### DIFF
--- a/apps/ui-storybook-e2e/src/integration/select/select.cy.ts
+++ b/apps/ui-storybook-e2e/src/integration/select/select.cy.ts
@@ -378,4 +378,23 @@ describe('select', () => {
 			cy.get('hlm-select-scroll-down').should('exist');
 		});
 	});
+
+	describe('falsy value', () => {
+		beforeEach(() => {
+			cy.visit('/iframe.html?id=select--falsy-value');
+			cy.injectAxe();
+		});
+
+		it('should select an option with a falsy value on enter keypress', () => {
+			cy.get('[brnselecttrigger]').click();
+
+			cy.realPress('ArrowDown');
+			cy.realPress('Enter');
+
+			// the falsy value is committed and the panel closes
+			cy.get('hlm-select-content').should('not.exist');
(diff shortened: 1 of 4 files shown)


marcjulian (approved):
Good catch! looks good and the test ist good too!

PR #1505, opened 2026-06-19
fix(accordion): use natural height for dynamic content

PatrickLarocque:
## Summary

Fixes accordion content clipping when panel content changes after the initial render.

`BrnAccordionContent` previously exposed a measured `--brn-accordion-content-height` value. HLM used that value for the open-state height, which meant dynamically added content could be clipped when the stored measurement no longer matched the panel's intrinsic height.

This changes the default HLM accordion content styling to use natural open height instead of a BRN-owned measured height variable:

- closed content uses `h-0`
- open content uses `h-auto`
- overflow remains hidden so closed content is clipped

BRN still owns the accordion semantics (`data-state`, `role`, `aria-labelledby`, `inert`, ids) and keeps the existing `style` input for compatibility, but it no longer performs layout measurement for accordion content.

## Root Cause

The old behavior depended on a one-time or manually refreshed DOM measurement. That made the open-state height a snapshot. If projected content changed after that snapshot, the host height could stay stale while the real content height grew.

This showed up in two cases:

- content appended while the panel is already open
- content appended while the panel is closed, then revealed later

## Fix

- Removed accordion content width/height CSS variable bindings from BRN.
- Removed accordion-specific DOM measurement and observer setup from BRN.
- Kept the existing BRN `style` input to avoid changing that documented input surface.
- Updated HLM accordion content classes from measured-height open state to natural-height open state.
- Updated the CLI accordion template so generated HLM components match the packaged HLM component.
- Added HLM coverage for the new accordion content class contract.
- Kept BRN coverage focused on semantic state and inert behavior.

## Notes

This intentionally removes the default numeric height target for accordion content. The default behavior now prioritizes correctness for dynamic content and keeps layout measurement out of the BRN primitive.

Consumers can still override HLM classes if they need a different animation strategy for a specific app.

## Testing

- `NX_DAEMON=false nx test tools --skip-nx-cache`
- `NX_DAEMON=false nx test brain --skip-nx-cache`
- `NX_DAEMON=false nx test helm --skip-nx-cache`
- `git diff --check`
- Storybook manual verification for dynamic content added while open and while closed before reopening

## Summary by CodeRabbit

* **New Features**
  * Added support for dynamic accordion content through improved content rendering.

* **Improvements**
  * Simplified accordion content height handling for more predictable behavior.

* **Breaking Changes**
  * Removed accordion configuration API.

* **Tests**
  * Added comprehensive tests for dynamic accordion content and height collapse behavior.

### DIFF
--- a/apps/ui-storybook/stories/accordion.stories.ts
+++ b/apps/ui-storybook/stories/accordion.stories.ts
@@ -298,6 +298,75 @@ export const TestAriaValidWithIcon: Story = {
 	},
 };
 
+export const DynamicContent: Story = {
+	render: () => ({
+		moduleMetadata: {
+			imports: [AccordionDynamicContentStory],
+		},
+		template: `<accordion-dynamic-content />`,
+	}),
+};
+
+@Component({
+	selector: 'accordion-dynamic-content',
+	imports: [HlmAccordionImports, HlmButtonImports],
+	host: {
+		class: 'block w-[420px]',
+	},
+	template: `
+		<div class="mb-4 flex flex-wrap gap-2">
+			<button hlmBtn size="sm" (click)="addContent()">Add content</button>
+			<button hlmBtn size="sm" variant="secondary" (click)="toggle()">Toggle</button>
+			<button hlmBtn size="sm" variant="outline" (click)="resetContent()">Reset</button>
+		</div>
+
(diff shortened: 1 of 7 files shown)


goetzrobin:
Closing in favor of #1523

Issue #543, opened 2024-12-24
Respect reduced-motion accessibility setting in helm styles


benjaminforras:
### Please provide the environment you discovered this bug in.

https://www.spartan.ng/components/dropdown-menu

### Which area/package is the issue in?

Don't know / other

### Description

If I disable animations in in the System Settings or emulate reduce-motion in Chrome, the animations are still playing.

shadcn/ui does not support this but generally it is good to have it in mind when developing and providing UI styles.

### I would be willing to submit a PR to fix this issue

- [ ] Yes
- [X] No

PR #1514, opened 2026-06-21
fix(dropdown-menu): respect reduced motion

gerasidev:
Disable dropdown and submenu animations when users request reduced motion across every style.

Refs #543

## Summary by CodeRabbit

* **Bug Fixes / Accessibility**
  * Updated dropdown menu styling across all theme variants so open/close animations apply only when motion-safe is enabled, honoring reduced-motion preferences.
* **Tests**
  * Added a reduced-motion test that scans each generated theme’s CSS to verify motion-safe-gated dropdown animation rules for both the main content and sub-content, and that unguarded animation rules are not present.


goetzrobin:
@gerasidev can you check the shadcn registry. two apply-s feels odd.

gerasidev:
> @gerasidev can you check the shadcn registry. two apply-s feels odd.

Indeed, I checked the shadcn registry and combined both into a single @apply. Fixed now.

marcjulian:
Hey, Tailwind CSS also has `motion-safe` to only enable animation when the user not reduced motion activated.
This can be useful when `motion-reduce` would mean to "[undo](https://tailwindcss.com/docs/hover-focus-and-other-states#prefers-reduced-motion)" styles like here where you also add important.

We use this in the `HlmSpinner` component

https://github.com/spartan-ng/spartan/blob/b22abbef825a19e2639c129d0ac1a63fed45c2ec/libs/helm/spinner/src/lib/hlm-spinner.ts#L30

gerasidev:
> Hey, Tailwind CSS also has `motion-safe` to only enable animation when the user not reduced motion activated. This can be useful when `motion-reduce` would mean to "[undo](https://tailwindcss.com/docs/hover-focus-and-other-states#prefers-reduced-motion)" styles like here where you also add important.
> 
> We use this in the `HlmSpinner` component
> 
> https://github.com/spartan-ng/spartan/blob/b22abbef825a19e2639c129d0ac1a63fed45c2ec/libs/helm/spinner/src/lib/hlm-spinner.ts#L30

Hi updated the dropdown animations to use motion-safe and removed the motion-reduce override  based  on your  suggestion.

marcjulian:
@gerasidev why did you close your PR?

gerasidev:
> @gerasidev why did you close your PR?

sry  accidental

marcjulian:
No worries :). Can you resolve the conflicts from the main branch?

marcjulian (approved):
Looks good!

Issue #1070, opened 2025-12-09
Add Presence Directive for Conditional Rendering with Animation Support


shikhardixitprojects:
### Which scope/s are relevant/related to the feature request?

accordion

### Information

# Feature Request: Add Presence Directive for Exit Animations

## Problem Statement

Currently, Spartan components keep content in the DOM permanently and use CSS to hide it when closed. This differs from Radix UI's default behavior, where content is conditionally rendered (removed from the DOM when not visible).

**Current Spartan behavior:**

- Accordion content remains in the DOM even when closed
- Uses CSS (`display: none`, `visibility: hidden`, etc.) to hide content
- Content is always present in the accessibility tree and DOM

**Radix UI behavior (React):**

- Content is conditionally rendered - removed from DOM when closed
- Uses a "Presence" component to delay unmounting until animations complete
- Cleaner DOM, better accessibility, and improved performance

This is primarily about **conditional rendering by default**, not animations. While consumers can listen to `animationend` events to hide content after animations, the default behavior should match Radix UI's approach of removing content from the DOM.

## Proposed Solution

Add a `BrnPresence` directive (similar to Radix UI's `Presence` component) that:

- Keeps content in the DOM while `present` signal is `true`
- Delays unmounting when `present` becomes `false` to allow exit animations
- Automatically cleans up after animations complete

### API Design

```typescript
import {
	Directive,
	effect,
	inject,
	InjectionToken,
	Provider,
	Signal,
	TemplateRef,
	ViewContainerRef
} from '@angular/core';

/**
 * Context interface for BrnPresence directive
 * Contains a Signal that indicates whether the content should be present in the DOM
 */
export type BrnPresenceContext = {
	present: Signal<boolean>;
};

export const BRN_PRESENCE_CONTEXT = new InjectionToken<BrnPresenceContext>('BrnPresenceContext');

/**
 * Factory provider helper for parent components/directives
 *
 * Usage:
 *   providers: [
 *     provideBrnPresenceContext(() => ({ present: myBooleanSignal }))
 *   ]
 */
export function provideBrnPresenceContext(contextFactory: () => BrnPresenceContext): Provider {
	return {provide: BRN_PRESENCE_CONTEXT, useFactory: contextFactory};
}

@Directive({
	selector: '[brnPresence]',
	standalone: true
})
export class BrnPresence {
	private readonly context = inject(BRN_PRESENCE_CONTEXT);
	private readonly viewContainerRef = inject(ViewContainerRef);
[...]

PR #1523, opened 2026-06-22
fix(accordion): measure accordion/collapsible content reactively

goetzrobin:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [x] accordion
- [x] collapsible

`BrnAccordionContent` and `BrnCollapsibleContent` measured their content **once** in an
`afterNextRender` hook (with an `IntersectionObserver` fallback) and froze the result into the
`--brn-*-content-height` variable used to animate the open height. That one-time snapshot is the
shared root cause of two separate bugs:

- **Crash on a leading text node.** The accordion measured `firstChild`, which is a `Text` node when
  the projected content starts with an interpolation or any `preserveWhitespaces: true` content.
  Reading `.style.height` on a `Text` node throws.
- **Clipped/stale content.** Because the open height was a snapshot, content that changed after the
  initial measurement (or was added while open) kept the old height and got clipped.

The two primitives had also drifted to subtly different measurement code (accordion measured
`firstChild`, collapsible measured the host), which is why only the accordion crashed.

Supersedes #1472 (measures `firstElementChild`) and #1505 (drops measurement for `h-auto`). #1472 is
a correct micro-patch to a method this PR deletes; #1505's `h-auto` removes the height animation
because `height: auto` is not transitionable.

## What is the new behavior?

A single `injectContentDimensions` helper in `@spartan-ng/brain/core` that both content directives
share:

- measures the content wrapper (first **element** child) via `getBoundingClientRect`, so a leading
  text/comment node can never break it;
- keeps the value live with a `ResizeObserver`, so dynamic and async content (rows added while open,
  images/fonts loading) re-measures and the panel animates to fit;
- is SSR-safe.

The `--brn-*-content-height` variable and the open/close animation are unchanged. Sharing one helper
means accordion and collapsible can no longer diverge.

Adds brain coverage for both primitives (measure, re-measure on change, no crash on a leading text
node) and `@if`-driven "Dynamic content" docs examples for accordion and collapsible.

- [x] Yes
- [ ] No

The now-unused measurement config is removed: `provideBrnAccordionConfig`,
`injectBrnAccordionConfig`, `provideBrnCollapsibleConfig`, `injectBrnCollapsibleConfig`, their config
types, and the `measureDimensions` helper. The `measurementDisplay` option no longer exists — content
height is measured automatically, so there is nothing to configure.

## Other information

Related to #1070 (a Presence directive / DOM removal). That is a separate architectural decision;
this PR improves the current in-DOM measurement model and does not block it. The live `ResizeObserver`
goes slightly beyond radix (which re-measures only on toggle) to also cover content changing while a
panel is open.

## Summary by CodeRabbit

* **New Features**
  * Added “Dynamic content” examples for Accordion and Collapsible, with interactive toggles and updated Preview/Code tabs.
* **Bug Fixes**
  * Improved accordion/collapsible content sizing when projected content changes while expanded.
  * Prevented crashes when projected content begins with a leading text node.
* **Tests**
  * Added dimension-focused specs for both accordion and collapsible content.
* **Breaking Changes**
  * Removed the public Accordion and Collapsible configuration injection APIs; content dimension handling is now built-in.


goetzrobin:
@greptile-apps review again

goetzrobin:
@coderabbitai review again

marcjulian (approved):
Looks good!

PR #1528, opened 2026-06-22
test: add per-style cli smoke coverage with deterministic scaffolding

goetzrobin:
## PR Checklist

- [x] Docs have been added / updated (cli-smoke README)

## PR Type

- [x] CI related changes (test coverage / smoke matrix)

### Others

- [x] cli

The cli-smoke matrix scaffolds real workspaces for every supported setup but only ever exercised a single
style (vega). Nothing verified that the CLI replaces the `spartan-*` placeholders with the right style's
registry classes - the build's themed-CSS check only looks at `--theme` variables, so a broken style map
would pass. Separately, the nx scaffold was non-deterministic when run from an AI-agent host.

## What is the new behavior?

- One smoke cell per supported style (`nova`, `lyra`, `maia`, `mira`, `luma`), split across angular-cli and
  nx, asserting no leftover `spartan-*` placeholders and that a style-distinctive class (derived from
  `libs/registry`, not hardcoded) reaches the generated output.
- Deterministic scaffolding on any host: strip the AI-agent env markers (`CLAUDECODE`/`OPENCODE`/...) so
  `create-nx-workspace` runs its classic local preset instead of cloning a remote template and forcing Nx
  Cloud, pin it to the exact nx version, and add a canary that fails loudly if the classic layout is missing.

- [x] No

## Release Notes

* **Tests**
  * Expanded CLI smoke test coverage with additional per-style configurations across both Angular CLI and Nx generator paths.
  * Strengthened smoke-test validation to ensure configured style registry class placeholders are correctly replaced and that style-specific tokens are present before builds.

* **Documentation**
  * Updated smoke-test “matrix” documentation to explain per-style cells and how to keep style entries in sync when supported styles change.

### DIFF
--- a/.github/workflows/cli-smoke.yml
+++ b/.github/workflows/cli-smoke.yml
@@ -43,6 +43,13 @@ jobs:
           - acli
           - all-components
           - all-components-nx-entry
+          # Style coverage: one cell per non-default style, split across angular-cli and nx. Keep in sync
+          # with the style-* cells in matrix.ts.
+          - style-acli-nova
+          - style-nx-lyra
+          - style-acli-maia
+          - style-nx-mira
+          - style-acli-luma
     name: ${{ matrix.cell }}
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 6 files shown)


MerlinMoos (approved):
LGTM thanks for this addition

PR #1529, opened 2026-06-22
feat: add selectable logo variant to the docs app

goetzrobin:
## PR Checklist

- [ ] Tests for the changes have been added (n/a - docs-app UI)
- [ ] Docs have been added / updated (n/a)

## PR Type

- [x] Feature

Docs app (`apps/app`) - not a published package.

The docs app renders a single spartan logo with no way to switch marks.

## What is the new behavior?

Adds a selectable logo variant (`default`/`legacy`) persisted in `localStorage` and applied via a `logo-*`
class on the document, with a context menu on the header logo to switch between the new and legacy marks.

- [x] No

## Summary by CodeRabbit

* **New Features**
  * Added a logo variant switcher in the header via a dropdown (default/legacy).
  * The chosen logo variant now persists across sessions and updates site-wide theming.
* **Bug Fixes**
  * Improved legacy logo rendering consistency, including correct rotation behavior for logo placeholders.
* **Other**
  * Updated the Spartan logo to support a flat (solid) vs. gradient rendering mode.
  * Added a custom styling variant to scope legacy logo styles.

### DIFF
--- a/apps/app/index.html
+++ b/apps/app/index.html
@@ -30,6 +30,10 @@
 			if (localStorage.layout) {
 				document.documentElement.classList.add('layout-' + localStorage.layout);
 			}
+			const logoVariant = ['default', 'legacy'].includes(localStorage.logoVariant)
+				? localStorage.logoVariant
+				: 'default';
+			document.documentElement.classList.add('logo-' + logoVariant);
 			if (window.navigator.userAgent.toUpperCase().includes('MAC')) {
 				document.documentElement.style.setProperty('--is-mac', 'true');
 			}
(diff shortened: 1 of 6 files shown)


elite-benni:
For the id my Idea was to remove the id from the ThreeHundred logos, therefore makeing them without gradient. 
Also on the ThreeHundred i think its prettier if we do not rotate the new logo. 
```
Suggested fix:
  - Add a flat input to SpartanLogo: solid fill, no <defs>.
  - Pass flat to <spartan-logo> in spartan-th-item-placeholder. Header stays the only instance with the gradient IDs.

  Issue 2 — placeholder rotation. -rotate-90 fits the old wordmark; the new square mark looks wrong. Change -rotate-90 → logo-legacy:-rotate-90.
  ```

MerlinMoos (approved):
LGTM, short question why do we need the legacy logo?

goetzrobin:
> LGTM, short question why do we need the legacy logo?

just an easter egg for those who liked the old one

PR #1536, opened 2026-06-22
fix(tooltip): restore exit animation and align data-state vocabulary

goetzrobin:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [x] tooltip

### Others

- [x] cli

The tooltip lost its exit animation. On hide, the directive set the content to
`data-state="closed"` and then immediately called `overlayRef.detach()` on the
next line, removing the content from the DOM before the
`data-[state=closed]` fade/zoom-out could paint. The tooltip just blinked out.

Separately, the content emitted `data-state="opened"` — a value no style or
shadcn class targets (every other overlay uses `open`). As a result the
per-style `data-open:`/`data-[state=delayed-open]:` enter-animation classes in
all six style files were dead; the enter animation only ran because the default
helm classes carried an ungated `animate-in`.

## What is the new behavior?

- Teardown is deferred until the content's exit animations finish, reusing
  `waitForElementAnimations` from `@spartan-ng/brain/core` (the same helper the
  dialog/overlay lifecycle split introduced). If there is no exit animation
  (e.g. `prefers-reduced-motion`), it detaches immediately.
- Re-hovering a tooltip mid-close revives it (cancels the pending teardown and
  flips it back open) instead of letting it fade out and detach.
- The content now emits `data-state="open"` to match every other overlay and
  shadcn, which revives the per-style enter-animation classes. The default helm
  classes were switched to the same gated convention (`data-open:` /
  `data-[state=delayed-open]:` / `data-closed:`) the style blocks already use,
  and the CLI tooltip template was kept in sync.
- Added the first tooltip animation specs (teardown, revive, no-animation fast
  path).

- [ ] Yes
- [x] No

## Other information

The tooltip content's `data-state` changes from `opened` to `open`. This is a
rendered-DOM value only; consumers style via the shipped helm classes, and
`opened` matched no existing selector, so this aligns the value rather than
changing supported behavior.

The `data-[state=delayed-open]` enter classes are now wired but remain dormant
until a follow-up adds skip-delay grouping (the Radix `TooltipProvider`
`delayed-open`/`instant-open` behavior); this PR is the prerequisite rename for
that work.

## Summary by CodeRabbit

* **New Features**
  * Added tooltip group support with configurable skip-delay so the next tooltip can open instantly during the group window.
  * Updated tooltip open/close animations and transition selectors for more consistent behavior.
* **Bug Fixes**
  * Tooltip teardown now waits for exit animations to finish, preventing premature detaching.
  * Re-hovering during the exit phase keeps the tooltip mounted and restores the open state; content/position updates remain synchronized.
* **Tests**
  * Added coverage for animated vs non-animated teardown, mid-exit recovery, content updates, skip-delay helper, and tooltip group behavior.

### DIFF
--- a/apps/app/src/app/pages/(components)/components/(tooltip)/tooltip--group.example.ts
+++ b/apps/app/src/app/pages/(components)/components/(tooltip)/tooltip--group.example.ts
@@ -1,0 +1,18 @@
+import { Component } from '@angular/core';
+import { provideBrnTooltipGroup } from '@spartan-ng/brain/tooltip';
+import { HlmButtonImports } from '@spartan-ng/helm/button';
+import { HlmTooltipImports } from '@spartan-ng/helm/tooltip';
+
+@Component({
+	selector: 'spartan-tooltip-group',
+	imports: [HlmButtonImports, HlmTooltipImports],
+	providers: [provideBrnTooltipGroup({ skipDelayDuration: 300 })],
+	template: `
+		<div class="flex gap-2">
+			<button [hlmTooltip]="'Save'" hlmBtn variant="outline">Save</button>
+			<button [hlmTooltip]="'Copy'" hlmBtn variant="outline">Copy</button>
+			<button [hlmTooltip]="'Paste'" hlmBtn variant="outline">Paste</button>
+		</div>
+	`,
+})
+export class TooltipGroup {}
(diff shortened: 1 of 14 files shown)


MerlinMoos (approved):
LGTM thanks for this addition

PR #1592, opened 2026-06-26
feat: add charts library and shadcn-style charts docs

ashley-hunter:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

This adds a new publishable package, **`@spartan-ng/charts`** (not yet in the list below). It also touches:

### Others

- [x] cli
- [x] repo

spartan has no chart primitives and no `@spartan-ng/charts` package. There is nothing in the docs
site for charts.

Closes #

## What is the new behavior?

Adds a d3-based **`@spartan-ng/charts`** library and a faithful, shadcn-style set of chart docs.

- **Library** (`libs/charts`): bar, line, area, pie, radar and radial-bar charts, plus shared
  axis / cartesian-grid / polar-grid / legend / tooltip / reference-line primitives. Charts are
  self-measuring (responsive) and colour their series from `--chart-1` … `--chart-5`.
- **Docs site**: a `/charts` landing page, per-type galleries (area, bar, line, pie, radar, radial,
  and a chart-tooltip gallery) that mirror the shadcn chart examples 1:1, and a `Chart` component
  documentation page under `/components/charts`.
- **Theme integration**: `@spartan-ng/cli:init` now writes `--chart-1` … `--chart-5` into the
  generated theme (light + dark), and `@spartan-ng/brain/hlm-tailwind-preset.css` maps them to the
  `--color-chart-*` Tailwind tokens so `*-chart-1` utilities resolve.
- **Release plumbing**: `@spartan-ng/charts` is wired into the release pipeline (project `release`
  target, version stamping, `release-charts` script, semantic-release git assets and the release
  workflow build) so it publishes in lockstep with brain / cli / mcp.

- [ ] Yes
- [x] No

## Other information

- The charts library is a d3 port; it currently ships without unit tests (verified against the
  shadcn references via the docs gallery). The one test added/updated here is for the
  `set-release-version` generator now that charts is published. Happy to add lib unit tests as a
  follow-up if preferred.
- `charts` is not yet a valid commitlint scope, so the commit omits a scope. Let me know if you want
  `charts` added to the scope-enum.


libs/charts/src/lib/area-chart/area-chart.ts
```diff
@@ -0,0 +1,792 @@
...
+				@for (area of areas(); track area; let i = $index) {
+					@if (area.fillGradient()) {
+						<linearGradient [attr.id]="clipId + '-g' + i" x1="0" y1="0" x2="0" y2="1">
+							<stop offset="5%" [attr.stop-color]="area.fill()" stop-opacity="0.8" />
+							<stop offset="95%" [attr.stop-color]="area.fill()" stop-opacity="0.1" />
+						</linearGradient>
+					}
+				}
+			</defs>
+			<g [attr.transform]="'translate(' + marginLeft + ',' + marginTop + ')'">
```
ajitzero, line 79: Hi, does this work in Firefox/Safari? I think we need to prefix all SVG tags with `svg:`, like `<svg:g></svg:g>`, `<svg:path></svg:path>`, `<svg:rect></svg:rect>`, etc.

ashley-hunter: Good question - confirmed working in both Firefox and Safari.

The `svg:` prefix is only needed when an SVG tag has no static `<svg>` ancestor in the same template. Here every SVG element is nested inside the `<svg>` in the same component template, so Angular's compiler enters the SVG namespace and creates them all via `createElementNS`. That's a compile-time decision, so it's identical across rendering engines.

Verified in real Firefox 139: across area/bar/line/pie/radar, every SVG element (static template `<g>`/`<rect>`/`<clipPath>`/`<linearGradient>` and the d3-appended `<path>`/`<rect>`) reports `namespaceURI === "http://www.w3.org/2000/svg"`, with 0 mis-namespaced elements (out of 248/136/76/23/10), and all charts render correctly. Also verified rendering in Safari.

So no `svg:` prefixes are needed. Happy to add them as belt-and-suspenders if you'd prefer, but it'd be a large no-op diff across every chart template.

ajitzero: No, all good since it works! I couldn't find any docs that could explain why my memory wanted that prefix, so this is great 😅

<details>
<summary>(Probably off-topic context)</summary>

In my own bare-bones implementation, I went with the directives route without using d3 selectors, so the line component (child) [had its own template](https://github.com/ajitzero/ng-vz/blob/main/apps/ng-vz/src/lib/cartesian/line.component.ts) without a parent `<svg>` tag. That worked in Chrome but not in Firefox, so I ended up adding the `svg:` prefix without fully understanding why.

```ts
@Component({
	selector: 'g[vzLine]',
	template: `<svg:path></svg:path>`
})
```

The only benefit of this approach was fewer DOM elements, so I didn't even add a local state service for the chart context.

</details>

ajitzero (approved):

apps/app/src/app/pages/(charts)/charts/(area-chart)/area-chart.page.ts
```diff
@@ -0,0 +1,203 @@
...
+		AreaChartAxes,
+		NgIcon,
+	],
+	providers: [provideIcons({ lucideTrendingUp })],
+	template: `
+		<div class="mb-20">
+			<charts-card
+				title="Area Chart - Interactive"
+				description="Showing total visitors for the last 3 months"
+				[code]="_snippets()['interactive'] ?? ''"
```
marcjulian: VS Code and serving the docs report that `??` is safe to be removed

libs/charts/src/lib/area-chart/area-chart.ts
```diff
@@ -0,0 +1,792 @@
...
+	get clipUrl(): string {
+		return `url(#${this.clipId})`;
+	}
+
+	readonly revealId = `${this.clipId}-reveal`;
+	get revealUrl(): string {
+		return `url(#${this.revealId})`;
+	}
+	private readonly revealRect = viewChild<ElementRef<SVGRectElement>>('revealRect');
+	private hasRevealed = false;
```
marcjulian, line 163: Could this be a signal?

libs/charts/src/lib/area-chart/area-chart.ts
```diff
@@ -0,0 +1,792 @@
...
+	// Where an outside press began, so we can tell a tap from a scroll gesture.
+	private outsidePressStart: { x: number; y: number } | null = null;
+
+	/**
+	 * Close the tooltip when *tapping* outside the chart, but not when starting a
+	 * scroll there. Like shadcn, we wait for pointerup and only close if the
+	 * pointer barely moved (a tap); a press that turns into a scroll leaves the
+	 * tooltip open.
+	 */
+	@HostListener('document:pointerdown', ['$event'])
```
marcjulian: Could we move the HostListener to the component host?

libs/charts/src/lib/area-chart/area-chart.ts
```diff
@@ -0,0 +1,792 @@
...
+			this.grids();
+			this.referenceLines();
+
+			this.render();
+		});
+	}
+
+	ngAfterContentInit(): void {
+		this.setupLayers();
+	}
```
marcjulian: Could we setup the layers in afterNextRender hook?

libs/charts/package.json
```diff
@@ -0,0 +1,18 @@
+{
+	"name": "@spartan-ng/charts",
+	"version": "1.0.2",
```
marcjulian: Do we want to start with an alpha version during the experimental phase? We could increment the version than more frequent during this phase and when its out of experimental phase we could sync spartan/ui with spartan/charts

marcjulian (commented):
Thank you for starting the charts package @ashley-hunter 👍 this is looking awesome! 
The charts page and also the charts component page is a very good start!

I only looked through it briefly and left only a few comments.

apps/app/src/app/pages/(charts)/charts/(area-chart)/area-chart.page.ts
ashley-hunter: Done - dropped the `?? ''` across the chart pages.

libs/charts/src/lib/area-chart/area-chart.ts
ashley-hunter: Left this one as a plain flag - it's a one-shot guard nothing reactive reads, so a signal wouldn't really buy anything here.

libs/charts/src/lib/area-chart/area-chart.ts
ashley-hunter: Done - moved them onto the host metadata object.

libs/charts/src/lib/area-chart/area-chart.ts
ashley-hunter: Done - setupLayers now runs in afterNextRender.

libs/charts/package.json
ashley-hunter: Good call - charts now releases on its own 1.0.0-alpha.N line under the alpha dist-tag, separate from the brain/cli/mcp lockstep.

ashley-hunter:
Thanks very much @marcjulian ! Made the suggested updates

PR #1595, opened 2026-07-01
feat(chart): add helm charts entry point

marcjulian:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [x] chart

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli
- [ ] mcp

## What is the new behavior?

Add a helm chart primitive for 

- HlmChartContainer
- HlmChartTooltipContent
- HlmChartLegendContent
- render bar scales without axis

## Summary by CodeRabbit

* **New Features**
  * Added reusable Helm chart container plus shared tooltip and legend content components for consistent chart UI.
  * Introduced a chart configuration system to control per-series labels, colors, and optional icons.
  * Updated chart demos/tooltips to use the shared container and tooltip content, including new axis examples.
  * Added a Storybook story for the chart container.

* **Bug Fixes**
  * Improved bar chart default scale/category handling.
  * Adjusted default grid dash behavior and made dashed rendering conditional when set.
  * Refined SVG axis styling to apply stroke/fill correctly.

### DIFF
--- a/libs/cli/src/generators/ui/libs/chart/files/lib/hlm-chart-container.ts.template
+++ b/libs/cli/src/generators/ui/libs/chart/files/lib/hlm-chart-container.ts.template
@@ -20,7 +20,11 @@ export class HlmChartContainer {
 	public readonly config = input<ChartConfig>(this._config);
 
 	constructor() {
-		classes(() => 'flex aspect-video justify-center text-xs');
+		classes(() => [
+			'flex aspect-video justify-center text-xs',
+			"[&_.grid-layer_line[stroke='#ccc']]:stroke-border/50",
+			'[&_.axes-layer_text]:fill-muted-foreground [&_.axes-layer_line]:stroke-border [&_.axes-layer_path]:stroke-border',
+		]);
 
 		afterNextRender(() => {
 			effect(
--- a/libs/cli/src/generators/ui/libs/chart/files/lib/hlm-chart-container.ts.template
+++ b/libs/cli/src/generators/ui/libs/chart/files/lib/hlm-chart-container.ts.template
@@ -1,0 +1,60 @@
+import { DOCUMENT } from '@angular/common';
+import { afterNextRender, Directive, effect, ElementRef, inject, Injector, input } from '@angular/core';
+import { classes } from '<%- importAlias %>/utils';
+import { CHART_THEMES, ChartConfig, injectHlmChartConfig } from './hlm-chart.token';
+
+@Directive({
+	selector: '[hlmChartContainer],hlm-chart-container',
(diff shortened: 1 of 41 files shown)


libs/cli/src/generators/ui/libs/chart/files/lib/hlm-chart-container.ts.template
```diff
@@ -0,0 +1,64 @@
...
+	public readonly config = input<ChartConfig>(this._config);
+
+	constructor() {
+		classes(() => [
+			'flex aspect-video justify-center text-xs',
+			"[&_.grid-layer_line[stroke='#ccc']]:stroke-border/50",
+			'[&_.axes-layer_text]:fill-muted-foreground [&_.axes-layer_line]:stroke-border [&_.axes-layer_path]:stroke-border',
+		]);
+
+		afterNextRender(() => {
```
marcjulian, line 29: This adds the color via styles, on first load its shortly black until rendered. Not sure if we can improve this

MerlinMoos:
hey @marcjulian and @ashley-hunter im just wondering why you created a new lib for the charts, should we move them into the brain, because Marc created now hlm entry points and then people can use them headless, and we can change the selector to brn. Just as a question

ashley-hunter:
> hey @marcjulian and @ashley-hunter im just wondering why you created a new lib for the charts, should we move them into the brain, because Marc created now hlm entry points and then people can use them headless, and we can change the selector to brn. Just as a question

We could, the only reservation I have is brain is 1.0 - and these charts are very much still in alpha state

MerlinMoos:
> > hey @marcjulian and @ashley-hunter im just wondering why you created a new lib for the charts, should we move them into the brain, because Marc created now hlm entry points and then people can use them headless, and we can change the selector to brn. Just as a question
> 
> We could, the only reservation I have is brain is 1.0 - and these charts are very much still in alpha state

Ah okay that make sense for me. Is there not possibility to add a hint to the components itself in entrypoint readme?

Also when we add hlm wrapper for it we have the same problem with the v1 version

ashley-hunter:
We can move it and just make it obvious in the docs etc that it's in alpha - I guess it's more in the situation where Claude or something starts using it in a codebase with the user never viewing the docs so it's not obvious in that case. 

An alternative is we use the brain selectors and prefixes on the charts - but keep them in a separate library for now still, so that way we could provide a migration that just fixes the import once stable. 

That doesn't solve the helm issue though. I guess one option for both helm and brain would be to have them in an experimental/charts entry point so the import path makes it obvious

PR #1590, opened 2026-06-26
fix(tooltip): Don't leak hover events on mobile devices

toniskobic:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our
      guidelines: https://github.com/spartan-ng/spartan/blob/main/CONTRIBUTING.md#-commit-message-guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] autocomplete
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] button-group
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] empty
- [ ] dropdown-menu
- [ ] field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-group
- [ ] input-otp
- [ ] item
- [ ] kbd
- [ ] label
- [ ] menubar
- [ ] native-select
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] resizable
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] sidebar
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [x] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli
- [ ] mcp

## What is the current behavior?

[...]

Issue #1582, opened 2026-06-26
[Tooltip] incorrect layout calculation for Tooltip


pumano:
### Please provide the environment you discovered this bug in.

chrome Android 

### Which area/package is the issue in?

tooltip

### Description

<img width="1080" height="2113" alt="Image" src="https://github.com/user-attachments/assets/9cd54062-2169-4fc6-a46c-3959799c7a7e" />

### Please provide the exception or error you saw

```true

```

### I would be willing to submit a PR to fix this issue

- [ ] Yes
- [x] No

toniskobic:
As [per Radix UI](https://github.com/radix-ui/primitives/issues/955#issuecomment-961813799), tooltips should only be displayed on hover events, not touch events. That means tooltips can't be displayed on mobile touch events. If you need to display some information after a touch event, a [Popover](https://spartan.ng/components/popover) should be used.

PR #1581, opened 2026-06-25
fix(tooltip): tooltip text not updated

Nicoss54:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [X] The commit message follows our
      guidelines: https://github.com/spartan-ng/spartan/blob/main/CONTRIBUTING.md#-commit-message-guidelines
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] autocomplete
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] button-group
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] empty
- [ ] dropdown-menu
- [ ] field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-group
- [ ] input-otp
- [ ] item
- [ ] kbd
- [ ] label
- [ ] menubar
- [ ] native-select
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] resizable
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] sidebar
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [X] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli
- [ ] mcp

## What is the current behavior?

While tooltip is open, if tooltip is a text content, when the content change and the tooltip remains open, tooltip content displayed doesn't change

Closes #1566

## What is the new behavior?

Tooltip text change is correctly displayed when tooltip still be open

## Does this PR introduce a breaking change?

[...]

Issue #1566, opened 2026-06-24
[Tooltip] String binding does not show updated value


ajitzero:
### Please provide the environment you discovered this bug in.

**Minimal Code:**    
```ts
import { ChangeDetectionStrategy, Component, computed, signal } from '@angular/core';
import { HlmTooltipImports } from '@spartan-ng/helm/tooltip';

@Component({
	selector: 'docs-theme-toggle',
	template: `
		<!-- Tooltip remains unchanged -->
		<button [hlmTooltip]="themeLabel()" (click)="toggleTheme()">Label</button>

		<!-- Tooltip updates as expected -->
		<ng-template #tooltip>{{ themeLabel() }}</ng-template>
		<button [hlmTooltip]="tooltip" (click)="toggleTheme()">Label</button>
	`,
	imports: [HlmTooltipImports],
	changeDetection: ChangeDetectionStrategy.OnPush,
})
export class ThemeToggle {
	private theme = signal<'light' | 'dark'>('light');

	public themeLabel = computed(() => 'Switch to ' + (this.theme() === 'light' ? 'dark' : 'light') + ' mode');

	public toggleTheme() {
		this.theme.set(this.theme() === 'dark' ? 'light' : 'dark');
	}
}
```

**Repo (NOT minimal):** https://github.com/ajitzero/animated-icons/blob/fc06b9cc8b968117f8b0aa12a128e14e99f12839/docs/src/app/components/header.ts

### Which area/package is the issue in?

tooltip

### Description

Add a [Simple](https://spartan.ng/components/tooltip#examples__simple) tooltip to a button, and bind the value as a string signal call. When the button is clicked, the signal value is modified.

While the tooltip remains open (user clicked or keyboard enter/space, and the button is still being hovered/focused on), the value of the tooltip:
- does not update when binding a string-value signal:
   - `[hlmTooltip]="themeLabel()"`
- updates as expected when using the template binding syntax:
   - `[hlmTooltip]="tooltip"`,
   - `<ng-template #tooltip>{{ themeLabel() }}</ng-template>`

I switched over to the template-based syntax, so this is not an urgent bug. This only affects those who use stateful buttons, like a theme toggle button or a submit button with a loading state.

#### [Simple](https://spartan.ng/components/tooltip#examples__simple) (text doesn't update)
https://github.com/user-attachments/assets/7d46a742-84b6-4b16-a1b7-fdb22b1abe47

[...]

PR #1600, opened 2026-07-02
fix(tooltip): keep the arrow pointing at the trigger at viewport edges

### LINKED ISSUE #1247: [BUG] Tooltip anchor is not positioned correctly

MacOs 26.3
Spartan ng: 0.0.1-alpha.643

accordion

### Description

The anchor of the tooltip is not positioned below the hovered element.

Image

Image

```true

```

### Other information

_No response_

- [ ] Yes
- [ ] No

goetzrobin:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [x] tooltip

Closes #1247

When a tooltip trigger sits near a viewport edge, CDK pushes the tooltip back
into the viewport along the cross axis so it stays fully visible. The arrow,
however, is a statically centered element on the panel — it has no knowledge of
the push — so after the shift it points into empty space beside the trigger
instead of at it.

This is a structural gap in the CDK approach: CDK has no arrow concept and
exposes no "how far did I push" value, so the arrow could never follow the
trigger on its own.

## What is the new behavior?

The arrow now tracks the trigger. We compute the delta between the trigger's
center and the panel's center from their bounding rects and slide the arrow by
that amount (clamped so it can't leave the panel) — the manual equivalent of
Floating UI's `arrow` middleware, which is what Radix relies on.

- Top/bottom tooltips track on the X axis, left/right on the Y axis.
- The geometry is a pure helper (`getTooltipArrowOffset`) in the position
  module, unit-tested for tracking and clamping on both axes.
- The directive applies it as a margin on show and on every CDK position
  change, so it survives flips, scroll and resize.
- Applied as a margin so it does not clobber the arrow's existing
  rotate/translate transform.

### Tests

- `brn-tooltip-arrow.spec.ts` — real-browser reproduction: a corner-anchored
  trigger forces a horizontal push and asserts the arrow still sits over the
  trigger's center (fails on `main`, passes here).
- `brn-tooltip-position.spec.ts` — pure unit coverage of the offset + clamp math.

- [ ] Yes
- [x] No

## Other information

Scoped deliberately to the arrow-anchoring bug (#1247). The dynamic-text bug
(#1566) and mobile-viewport positioning (#1582) are tracked in their own PRs
(#1581, #1590).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tooltip arrow alignment when a tooltip is pushed to stay within the viewport.
  * Tooltips now keep the arrow centered on the trigger more reliably after repositioning.
  * Added coverage for arrow placement and tooltip position behavior to prevent regressions.

### DIFF
--- a/libs/brain/tooltip/src/lib/brn-tooltip-arrow.spec.ts
+++ b/libs/brain/tooltip/src/lib/brn-tooltip-arrow.spec.ts
@@ -1,0 +1,74 @@
+import { Directionality } from '@angular/cdk/bidi';
+import { ChangeDetectionStrategy, Component } from '@angular/core';
+import { fireEvent, render, screen, waitFor } from '@testing-library/angular';
+import { BrnTooltip } from './brn-tooltip';
+
+// A trigger pinned to the top-left corner with a large tooltip: 'top' is off-screen and the
+// left/right sides can't fit the tall panel, so CDK keeps 'bottom' and *pushes* it right to stay in
+// the viewport. That push is what knocks a center-anchored arrow off the trigger (issue #1247).
+@Component({
+	selector: 'brn-tooltip-arrow-host',
+	imports: [BrnTooltip],
+	providers: [Directionality],
+	changeDetection: ChangeDetectionStrategy.OnPush,
+	template: `
+		<button
+			brnTooltip="tip"
+			position="bottom"
+			[showDelay]="0"
+			[hideDelay]="0"
+			style="position: fixed; top: 0; left: 0; width: 40px; height: 24px"
+		>
+			trigger
+		</button>
+	`,
+})
+class TooltipArrowHost {}
+
+const triggerEl = () => screen.getByText('trigger');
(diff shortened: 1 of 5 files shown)


goetzrobin:
Addressed both greptile findings in ad52fe8bf:

- **Arrow offset measured before positioning (P2):** removed the `_updateArrowOffset()` call from the synchronous show path. It ran before CDK positioned the pane, so it measured an overlay still at the origin — for left/right tooltips that clamped the arrow to the panel edge (visible as the arrow shoved off-center on open, even for a short tooltip on a tall trigger). The offset is now measured from the `positionChanges` subscription, where the pane is already positioned.
- **Stale `_activePosition` on unmappable pairs (P2):** the `_updateArrowOffset()` call now lives inside the `if (resolved)` block, so the offset is only computed when we have a known side — never on a stale axis.

Added a regression test (short tooltip centered on a tall trigger stays centered). The revive-path call is kept as-is, since the overlay is attached and positioned there.

PR #1608, opened 2026-07-04
feat(switch): introduce additional switch sizes

PatrickLarocque:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [x] switch

### Others

- [x] cli

The switch primitive only accepts `sm` and `default` size values, so consumers cannot opt into larger switch presentations through the component API.

Closes N/A

## What is the new behavior?

Adds explicit `md`, `lg`, and `xl` switch sizes for `brn-switch` and `hlm-switch`.

The existing `default` value remains the component default and continues to map to the legacy registry dimensions, so existing apps that do not specify a size keep their current rendered switch size. The new `md` size intentionally aliases those same legacy `default` dimensions in every theme, so `default` and `md` render the same while giving consumers an explicit medium size name.

The switch sizes example now documents `sm`, `md`, `lg`, and `xl`.

shadcn/ui, the reference library, only has the two existing switch sizes, so this is intentionally not a parity feature.

- [ ] Yes
- [x] No

## Other information

Existing legacy `default` dimensions are preserved for compatibility, and `md` mirrors them. The added larger sizes use the registry theme geometry for `lg` and `xl`.

Review feedback addressed:

- Preserved the legacy `default` component default to avoid a silent visual size change on upgrade.
- Made `md` match legacy `default` dimensions across themes so medium and default are not visually different.
- Restored CLI template import ordering so generated `hlm-switch` code matches the published library source ordering.

Validation performed:

- `pnpm nx test brain -- --run switch/src/lib/brn-switch.spec.ts` passed; Nx ran the full brain suite: 44 files, 420 tests.
- `pnpm nx build brain` passed.
- `pnpm nx build helm` passed.
- `NODE_OPTIONS=--max-old-space-size=6000 pnpm nx build app` passed before the final review cleanup.
- `git diff --check` passed after the final review cleanup.

## Summary by CodeRabbit

- **New Features**
  - Expanded the Switch size options to support Medium, Large, and Extra large (in addition to Small and default).
  - Updated the Switch sizing example layout and option labels to reflect the new sizes.
- **Bug Fixes**
  - Updated Switch track and thumb dimensions so sizing renders consistently across supported themes.
- **Tests**
  - Enhanced Switch size test coverage to validate the correct reported value for all supported sizes.

### DIFF
--- a/apps/app/src/app/pages/(components)/components/(switch)/switch--sizes.example.ts
+++ b/apps/app/src/app/pages/(components)/components/(switch)/switch--sizes.example.ts
@@ -6,7 +6,7 @@ import { HlmSwitch } from '@spartan-ng/helm/switch';
 	selector: 'spartan-switch-sizes',
 	imports: [HlmLabel, HlmSwitch],
 	host: {
-		class: 'flex items-center gap-6',
+		class: 'flex flex-wrap items-center gap-6',
 	},
 	template: `
 		<label class="flex items-center" hlmLabel>
@@ -14,8 +14,16 @@ import { HlmSwitch } from '@spartan-ng/helm/switch';
 			Small
 		</label>
 		<label class="flex items-center" hlmLabel>
-			<hlm-switch class="me-2" size="default" />
-			Default
+			<hlm-switch class="me-2" size="md" />
+			Medium
+		</label>
+		<label class="flex items-center" hlmLabel>
+			<hlm-switch class="me-2" size="lg" />
+			Large
+		</label>
+		<label class="flex items-center" hlmLabel>
+			<hlm-switch class="me-2" size="xl" />
+			Extra large
 		</label>
 	`,
 })
--- a/libs/brain/switch/src/lib/brn-switch.spec.ts
+++ b/libs/brain/switch/src/lib/brn-switch.spec.ts
@@ -291,14 +291,14 @@ describe('BrnSwitchComponent', () => {
 			return view;
 		};
 
-		it('defaults data-size to "default" on the button', async () => {
+		it('defaults data-size to "md" on the button', async () => {
 			await renderWithSize();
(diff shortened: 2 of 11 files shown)


MerlinMoos:
Hey @PatrickLarocque im not sure if we want to keep it because we want to stay close as possible to shadcn. That the benefit from spartan. You own the styles, if you need it in your project you can do it. 

@goetzrobin whats your thoughts

PatrickLarocque:
> Hey @PatrickLarocque im not sure if we want to keep it because we want to stay close as possible to shadcn. That the benefit from spartan. You own the styles, if you need it in your project you can do it.
> 
> @goetzrobin whats your thoughts

No worries, feel free to close if that's the direction. However, the issue from my perspective was that the dx is a bit awkward, since sizing is handled partially by the `data-size` attribute, and users have to remember to size the rail and thumb together if they want to override. 

This means that a user defined override ends up looking something like:

```html
<hlm-switch class="h-6 w-10 [&_[data-slot=switch-thumb]]:size-5" />
```

But then you are kind of needing to find where data-size applies a style and override it with a specific class, given the split styling contracts of data-size vs class-based across brn and hlm, rail and thumb.  I suppose a user could define their own size variants and pass them to brn to make the dx a bit smoother, but my initial inclination was to make size variants a bit easier since they philosophically feel like more of a `hlm` thing. My solution to add additional size variants doesn't really solve this underlying tension though haha, it just smooths out the dx by adding more shorthands for size (since their definitions are brm owned).

goetzrobin:
Hey there! I think our goal is to stay as close to shadcn as possible. Since people own hlm, they can add this functionality in their own code base. Not saying you are wrong about ergonomics, it just becomes really hard to figure out what discrepancies between shadcn and spartan are deliberate and which ones are not! Hope that makes sense!

Issue #1597, opened 2026-07-01
cannot run init when using pnpm catalog


FabienDehopre:
### Please provide the environment you discovered this bug in.

Create an empty angular application with PNPM and move all the package version into catalog.
My package.json looks like this:
```json
{
  "name": "my-angular-app",
  "version": "0.0.0",
  "private": true,
  "files": [
    ".",
    "dist"
  ],
  "scripts": {
    "ng": "ng",
    "start": "ng serve",
    "build": "ng build",
    "watch": "ng build --watch --configuration development",
    "test": "ng test"
  },
  "dependencies": {
    "@angular/cdk": "catalog:angular",
    "@angular/common": "catalog:angular",
    "@angular/compiler": "catalog:angular",
    "@angular/core": "catalog:angular",
    "@angular/forms": "catalog:angular",
    "@angular/platform-browser": "catalog:angular",
    "@angular/router": "catalog:angular",
    "@opentelemetry/auto-instrumentations-web": "catalog:opentelemetry",
    "@opentelemetry/context-zone": "catalog:opentelemetry",
    "@opentelemetry/exporter-trace-otlp-http": "catalog:opentelemetry",
    "@opentelemetry/instrumentation": "catalog:opentelemetry",
    "@opentelemetry/resources": "catalog:opentelemetry",
    "@opentelemetry/sdk-trace-base": "catalog:opentelemetry",
    "@opentelemetry/sdk-trace-web": "catalog:opentelemetry",
    "@opentelemetry/semantic-conventions": "catalog:opentelemetry",
    "rxjs": "catalog:",
    "tslib": "catalog:",
    "zod": "catalog:",
    "zone.js": "catalog:"
  },
  "devDependencies": {
    "@angular/build": "catalog:angular",
    "@angular/cli": "catalog:angular",
    "@angular/compiler-cli": "catalog:angular",
    "@spartan-ng/cli": "catalog:spartanui",
    "@tailwindcss/postcss": "catalog:tailwindcss",
    "@testing-library/angular": "catalog:testing-library",
    "@testing-library/jest-dom": "catalog:testing-library",
    "@testing-library/user-event": "catalog:testing-library",
    "@vitest/browser": "catalog:vitest",
    "@vitest/browser-playwright": "catalog:vitest",
    "@vitest/coverage-v8": "catalog:vitest",
    "@vitest/ui": "catalog:vitest",
    "jsdom": "catalog:",
    "postcss": "catalog:",
[...]

PR #1610, opened 2026-07-04
Handle pnpm catalog dependencies in the CLI generators

xonaib:
The CLI was still passing packages to addDependenciesToPackageJson even when they were already declared in the workspace package.json.

That breaks with pnpm catalogs because existing versions like catalog:angular are handed back to Nx/pnpm during the install task. This filters out dependencies that are already declared in dependencies, devDependencies, peerDependencies, or optionalDependencies before registering new packages.

Tests:
- pnpm exec vitest run --config libs/cli/vite.config.mts libs/cli/src/generators/base/lib/package-json-dependencies.spec.ts libs/cli/src/generators/base/lib/build-dependency-array.spec.ts libs/cli/src/utils/version-utils.spec.ts

## Summary by CodeRabbit

* **Bug Fixes**
  * Avoided writing duplicate dependencies and devDependencies when they’re already present in `package.json`, including catalog-managed references.
  * Skips updating `package.json` when there are no newly eligible dependency entries to apply.
* **New Features**
  * Added dependency filtering during init/generation so only unregistered, string-version candidates are added; catalog-managed entries are ignored.
* **Tests**
  * Added Jest coverage for catalog reference detection, existing concrete-version retention, peer/optional handling, and dropping null version entries.

### DIFF
--- a/libs/cli/src/generators/base/lib/package-json-dependencies.ts
+++ b/libs/cli/src/generators/base/lib/package-json-dependencies.ts
@@ -18,7 +18,8 @@ export function filterUnregisteredDependencies(tree: Tree, dependencies: Depende
 	const filter = (deps: Dependencies) =>
 		Object.fromEntries(
 			Object.entries(deps).filter(
-				([packageName, version]) => typeof version === 'string' && !isCatalogManagedDependency(packageJson, packageName),
+				([packageName, version]) =>
+					typeof version === 'string' && !isCatalogManagedDependency(packageJson, packageName),
 			),
 		) as Record<string, string>;
 
--- a/libs/cli/src/generators/base/lib/package-json-dependencies.ts
+++ b/libs/cli/src/generators/base/lib/package-json-dependencies.ts
@@ -8,13 +8,9 @@ type PackageJsonDependencies = {
 	optionalDependencies?: Record<string, string>;
 };
 
-function hasDeclaredDependency(packageJson: PackageJsonDependencies, packageName: string): boolean {
-	return (
-		packageName in (packageJson.dependencies ?? {}) ||
-		packageName in (packageJson.devDependencies ?? {}) ||
-		packageName in (packageJson.peerDependencies ?? {}) ||
-		packageName in (packageJson.optionalDependencies ?? {})
-	);
+function isCatalogManagedDependency(packageJson: PackageJsonDependencies, packageName: string): boolean {
+	const version = packageJson.dependencies?.[packageName] ?? packageJson.devDependencies?.[packageName];
+	return typeof version === 'string' && version.startsWith('catalog:');
 }
 
 export function filterUnregisteredDependencies(tree: Tree, dependencies: Dependencies, devDependencies: Dependencies) {
@@ -22,7 +18,7 @@ export function filterUnregisteredDependencies(tree: Tree, dependencies: Depende
 	const filter = (deps: Dependencies) =>
 		Object.fromEntries(
 			Object.entries(deps).filter(
-				([packageName, version]) => typeof version === 'string' && !hasDeclaredDependency(packageJson, packageName),
+				([packageName, version]) => typeof version === 'string' && !isCatalogManagedDependency(packageJson, packageName),
 			),
 		) as Record<string, string>;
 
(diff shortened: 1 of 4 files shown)


libs/cli/src/generators/base/lib/package-json-dependencies.ts
```diff
@@ -0,0 +1,33 @@
...
+
+type Dependencies = Record<string, string | null | undefined>;
+type PackageJsonDependencies = {
+	dependencies?: Record<string, string>;
+	devDependencies?: Record<string, string>;
+	peerDependencies?: Record<string, string>;
+	optionalDependencies?: Record<string, string>;
+};
+
+function hasDeclaredDependency(packageJson: PackageJsonDependencies, packageName: string): boolean {
```
goetzrobin: One thing I'd love your take on re: scope. The crash in #1597 is specific to `catalog:` **values** (nx can't parse them as semver), and those only ever live in `dependencies` / `devDependencies`. But `hasDeclaredDependency` skips a package if it's declared **anywhere** — including as a *concrete* range — which has a side effect beyond the bug: today spartan relies on `addDependenciesToPackageJson` to raise an already-declared concrete dep (e.g. `tailwind-merge`) to its supported minimum, and skipping all declared deps turns that bumping off for **everyone**, not just catalog users.

Could we narrow this to only skip catalog-managed installs? Something like:

```ts
function isCatalogManaged(packageJson: PackageJsonDependencies, name: string): boolean {
  const existing = packageJson.dependencies?.[name] ?? packageJson.devDependencies?.[name];
  return typeof existing === 'string' && existing.startsWith('catalog:');
}
```

and keep the dep unless it's catalog-managed. That fixes the crash while (a) still bumping concrete deps, and (b) sidestepping the peer/optional question — a package listed only under `peerDependencies` (common for Angular libs declaring `@angular/cdk`) would otherwise be skipped even though it may genuinely need to be added.

If the intent is instead *"never touch anything the user already declared"*, that's a legitimate call — but since it changes behavior for all consumers, it might be worth gating behind a `--skip-existing` flag / prompt so that bumping stays the default. Happy to help with either direction!

libs/cli/src/generators/base/lib/package-json-dependencies.ts
```diff
@@ -0,0 +1,33 @@
...
+		packageName in (packageJson.optionalDependencies ?? {})
+	);
+}
+
+export function filterUnregisteredDependencies(tree: Tree, dependencies: Dependencies, devDependencies: Dependencies) {
+	const packageJson = readJson<PackageJsonDependencies>(tree, 'package.json');
+	const filter = (deps: Dependencies) =>
+		Object.fromEntries(
+			Object.entries(deps).filter(
+				([packageName, version]) => typeof version === 'string' && !hasDeclaredDependency(packageJson, packageName),
```
goetzrobin: 👍 Keeping the `typeof version === 'string'` guard is the right move — `buildDependencyArray` really can hand back a `null` (`@spartan-ng/brain` resolves to `null` when `@spartan-ng/cli` isn't installed and no fallback is passed), so this correctly drops it. Worth keeping regardless of which filtering direction we land on.

goetzrobin (commented):
Thanks a lot for tackling #1597 — the helper is clean and well-tested, and the null-handling is a genuinely good catch. I have one question about the filtering *scope* that I think is worth settling before merge, plus a small note. Nothing blocking from my side beyond that discussion. 🙏

libs/cli/src/generators/base/lib/package-json-dependencies.ts
xonaib: Good point on `Object.entries()` too, changed that to `Object.keys()` since only the key count matters there.

libs/cli/src/generators/base/lib/package-json-dependencies.ts
xonaib: Thanks, that makes sense. I narrowed this to only skip dependencies that are already catalog-managed in `dependencies` / `devDependencies`, so the PR stays focused on fixing the `catalog:` semver crash.

I didn’t add a `--skip-existing` option here because that felt broader than the original issue, but I’d be happy to give that direction a try if you think it’s worth handling in this PR.

I also kept the `typeof version === 'string'` guard and added regression coverage for concrete existing versions plus peer/optional declarations.

goetzrobin:
@xonaib can you check why formatting/linting is failing

PR #1617, opened 2026-07-06
chore(repo): generate-ui-docs inputs/outputs

MerlinMoos:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our
      guidelines: https://github.com/spartan-ng/spartan/blob/main/CONTRIBUTING.md#-commit-message-guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [ ] Other... Please describe:

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] autocomplete
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] button
- [ ] button-group
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [ ] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] empty
- [ ] dropdown-menu
- [ ] field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-group
- [ ] input-otp
- [ ] item
- [ ] kbd
- [ ] label
- [ ] menubar
- [ ] native-select
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] resizable
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] sidebar
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [x] repo
- [ ] cli
- [ ] mcp

## What is the current behavior?

Closes #

## What is the new behavior?

[...]

PR #1619, opened 2026-07-08
chore(repo): resolve api-docs references via ts-morph instead of a name registry

goetzrobin:
> **Stacked on top of #1617** (base branch: `api-docs-generator-fix`). This is a
> friendly review-follow-up to your extractor work — it keeps all of your new
> behavior (inheritance + host directives + alias handling) and the same public
> output, but leans on `ts-morph`'s type checker to do the resolution your
> registry was doing by hand. Net **−57 lines** in `executor.ts`. Happy to adjust
> anything or split it up.

## TL;DR

PR added the right *behavior*. This keeps behavior but swaps the
**mechanism**: instead of building a name→class registry and matching packages by
string heuristics, we load the workspace `tsconfig.base.json` into the `Project`
and let `getBaseClass()` / symbol resolution follow the real imports (barrels
included). 

## Why — the problems this fixes

**1. The name registry could silently resolve to the wrong class.**
`resolveClassByName` disambiguated same-named classes by matching the import's
module-specifier *tail* against candidate file paths with `includes()`:

```ts
// '@spartan-ng/brain/dialog' -> look for a path containing '/brain/dialog'
candidate.getSourceFile().getFilePath().includes(`/${specifierTail}`)
```

`'/brain/dialog'` is a substring of `'/brain/dialog-content'`, so an import of
`@spartan-ng/brain/dialog` can match a **different package** and, when the tail
match fails, it falls back to `candidates[0]` — an arbitrary pick by file order.
The extractor then documents the wrong class's inputs with no error. There's a
regression test for exactly this (`does not resolve to a sibling package whose
path is a prefix substring`) — it fails on the registry approach and passes here.

**2. Default values were mangled by a regex heuristic.**

```ts
const match = argText.match(/\(([^()]*)\)$/);
if (match && match[1]) defaultValue = match[1].replace(/['"]/g, '');
else defaultValue = argText.replace(/['"]/g, '');
```

- `input(signal(5))` was documented as `5` (the regex reached *into* the
  expression and grabbed the paren contents).
- `replace(/['"]/g, '')` stripped quotes indiscriminately: `input('')` became
  `""` (indistinguishable from "no default") and `{ label: 'x' }` became
  `{ label: x }`.

Now the default is taken **verbatim** from the argument node, so strings keep
their quotes (`'bottom'`, `''`) and expressions stay intact (`signal(5)`).

**3. The new helpers were untyped.** `componentInfo`, `target`,
`directiveMembers` were implicit `any`, with an `as object` cast to make the
spread compile.

## What changed

- **Deleted** `classRegistry`, `resolveClassByName`, and the name-based half of
  `resolveBaseClass` (~80 lines).
- **Added** a 6-line `resolveClass(expr)` that asks the type system for the
  declaration behind an `extends` / `directive:` reference.
- `resolveBaseClass` → just `cls.getBaseClass()`.
- Default-value extraction → verbatim `getText()` (see #2).
- Introduced `Member` / `MemberCollection` / `ComponentInfo` / `NestedStructure`
  types across the extractor; removed all `any` and the cast.
- Tests: the in-memory harness now mirrors the real workspace (barrel
  `index.ts` + `paths`), plus new cases for the collision fix, default-value
  formatting, JSDoc descriptions, and own-vs-host dedup.

## How — leaning on `ts-morph` instead of hand-wiring

The one enabling change:

```ts
const project = new Project({
  tsConfigFilePath: path.join(context.root, 'tsconfig.base.json'),
  skipAddingFilesFromTsConfig: true, // still only load brain/helm sources
});
```

Loading the tsconfig gives the checker the `paths` map, so `@spartan-ng/*`
imports resolve natively through the barrel re-exports. Everything the registry
was reconstructing — "which class does this name refer to, in which package" —
is now answered by the compiler that already knows. `getBaseClass()` walks
inheritance across packages; symbol resolution finds host-directive classes; the
same-name collision problem disappears because we resolve *symbols*, not strings.

## Behavior notes for reviewers

- **Public JSON shape is unchanged**; the `executor.spec.ts` snapshot regenerates
  identically (it strips `defaultValue`, so #2 doesn't move it).
- The **emitted `defaultValue` strings do change** for string/expression defaults
  (now quoted / verbatim). The docs table renders the value as-is, so e.g.
  `side` now shows `'bottom'` instead of `bottom`. This is intentional — see #2.

## Testing

- `nx test tools` — 23 passing (17 of yours + 6 added).
- Ran the real `nx run tools:generate-ui-docs` against the workspace; snapshot
  matches byte-for-byte.
- `tsc` + `eslint` clean.

---

## PR Checklist

- [ ] Docs have been added / updated (docs *output* changes; no doc pages)

## PR Type

- [x] Bugfix
- [x] Refactoring (no api changes)

- [x] repo

The extractor (as of #1617) resolves base classes / host directives via a
name→class registry with module-specifier string matching, extracts default
values with a regex, and leaves the new helpers untyped.

## What is the new behavior?

Resolution is delegated to `ts-morph` via the workspace tsconfig, default values
are kept verbatim, the extractor is fully typed, and the name-collision +
default-value bugs are fixed. See details above.

- [x] No (public JSON shape unchanged; `defaultValue` string formatting differs)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UI docs generation to honor workspace module resolution via `tsconfig` path mappings.
  * Refined host directive extraction and inheritance traversal for more accurate member resolution, including safer collision handling.
  * Updated `input()`/`model()` default handling to preserve the first argument exactly as written.
  * Ensured member descriptions from inline documentation are captured, and omitted `exportAs` when not provided.

* **Tests**
  * Expanded extraction tests to validate path resolution, default preservation, host-directive de-duplication, inheritance lookup behavior, and description/exportAs output.

### DIFF
--- a/libs/tools/src/executors/docs/generate-ui-docs/executor.ts
+++ b/libs/tools/src/executors/docs/generate-ui-docs/executor.ts
@@ -1,8 +1,15 @@
 import type { ExecutorContext } from '@nx/devkit';
 import fs from 'fs';
 import path from 'path';
-import type { ClassDeclaration, Decorator, Node, SourceFile } from 'ts-morph';
-import { ArrayLiteralExpression, CallExpression, ObjectLiteralExpression, Project, PropertyAssignment } from 'ts-morph';
+import type { ClassDeclaration, Decorator } from 'ts-morph';
+import {
+	ArrayLiteralExpression,
+	CallExpression,
+	Node,
+	ObjectLiteralExpression,
+	Project,
+	PropertyAssignment,
+} from 'ts-morph';
 import { writePerComponentData } from '../../../write-per-component-data';
 import type { GenerateUiDocsExecutorSchema } from './schema';
 
@@ -16,7 +23,11 @@ export default async function runExecutor(options: GenerateUiDocsExecutorSchema,
 		fs.mkdirSync(outputDir, { recursive: true });
 	}
 
-	const project = new Project();
+	// Load the workspace tsconfig so `@spartan-ng/*` imports resolve via its `paths` map.
+	const project = new Project({
+		tsConfigFilePath: path.join(context.root, 'tsconfig.base.json'),
+		skipAddingFilesFromTsConfig: true,
+	});
 	project.addSourceFilesAtPaths([
 		`${brainDir}/**/*.ts`,
 		`${helmDir}/**/*.ts`,
@@ -36,25 +47,7 @@ export default async function runExecutor(options: GenerateUiDocsExecutorSchema,
 }
 
 export function extractInputsOutputs(project: Project, workspaceRoot: string) {
-	const inputsOutputs = {};
-
-	// Build a registry of every class by name so we can resolve base classes and
-	// host directives even when they are imported via TS path aliases (which the
-	// type checker can't resolve without a tsconfig). Multiple classes may share a
-	// name across packages, so we keep a list per name and disambiguate later.
-	const classRegistry = new Map<string, ClassDeclaration[]>();
-	project.getSourceFiles().forEach((sourceFile) => {
-		sourceFile.getClasses().forEach((cls) => {
-			const name = cls.getName();
-			if (!name) return;
-			const existing = classRegistry.get(name);
-			if (existing) {
-				existing.push(cls);
-			} else {
-				classRegistry.set(name, [cls]);
-			}
-		});
-	});
+	const inputsOutputs: NestedStructure = {};
 
 	project.getSourceFiles().forEach((sourceFile) => {
 		// if the source file is a .spec.ts file then skip
@@ -67,7 +60,7 @@ export function extractInputsOutputs(project: Project, workspaceRoot: string) {
 			// Get the full file path and make it relative to workspace root
 			const fullPath = sourceFile.getFilePath();
 			const relativeFilePath = path.relative(workspaceRoot, fullPath);
-			const componentInfo = {
+			const componentInfo: ComponentInfo = {
 				file: relativeFilePath,
 				inputs: [],
 				outputs: [],
@@ -95,11 +88,11 @@ export function extractInputsOutputs(project: Project, workspaceRoot: string) {
 			// Collect members from the class itself and every ancestor it extends,
 			// so inherited inputs/outputs/models show up in the generated docs.
 			const seenMembers = new Set<string>();
-			collectAllMembers(cls, componentInfo, classRegistry, seenMembers);
+			collectAllMembers(cls, componentInfo, seenMembers);
 
 			// Collect inputs and outputs exposed through `hostDirectives`, applying the
 			// renaming declared in the decorator (e.g. 'brnTabsContent: hlmTabsContent').
-			collectHostDirectiveMembers(cls, componentInfo, classRegistry, seenMembers);
+			collectHostDirectiveMembers(cls, componentInfo, seenMembers);
 
 			if (
 				componentInfo.inputs.length ||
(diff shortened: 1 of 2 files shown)


libs/tools/src/executors/docs/generate-ui-docs/executor.ts
```diff
@@ -16,7 +23,11 @@ export default async function runExecutor(options: GenerateUiDocsExecutorSchema,
 		fs.mkdirSync(outputDir, { recursive: true });
 	}
 
-	const project = new Project();
+	// Load the workspace tsconfig so `@spartan-ng/*` imports resolve via its `paths` map.
+	const project = new Project({
+		tsConfigFilePath: path.join(context.root, 'tsconfig.base.json'),
```
goetzrobin, line 28: **This one line is the whole idea.** Loading `tsconfig.base.json` gives the `Project` the `paths` map, so the type checker resolves `@spartan-ng/*` imports natively (through the barrel `index.ts` re-exports). `skipAddingFilesFromTsConfig: true` keeps us scoped to just the brain/helm globs below — we borrow the compiler options, not the whole workspace file set. Everything the name registry did by hand falls out of this.

libs/tools/src/executors/docs/generate-ui-docs/executor.ts
```diff
@@ -116,27 +109,39 @@ export function extractInputsOutputs(project: Project, workspaceRoot: string) {
...
+type NestedStructure = Record<string, Record<string, Record<string, ComponentInfo>>>;
+
+function collectAllMembers(cls: ClassDeclaration, componentInfo: MemberCollection, seen: Set<string>) {
 	const visitedClasses = new Set<ClassDeclaration>();
 	let currentClass: ClassDeclaration | undefined = cls;
 	while (currentClass && !visitedClasses.has(currentClass)) {
 		visitedClasses.add(currentClass);
 		collectMembers(currentClass, componentInfo, seen);
-		currentClass = resolveBaseClass(currentClass, classRegistry);
+		currentClass = currentClass.getBaseClass();
```
goetzrobin, line 140: Inheritance is now just `getBaseClass()`. Because the Project has the tsconfig paths, this walks the `extends` chain **across packages** (e.g. `HlmX extends BrnX` imported via `@spartan-ng/brain/...`) with no registry lookup.

libs/tools/src/executors/docs/generate-ui-docs/executor.ts
```diff
@@ -207,78 +207,21 @@ function mapHostDirectiveMembers(memberProp, directiveMembers, kind: 'input' | '
...
-		if (specifierTail) {
-			const match = candidates.find((candidate) =>
-				candidate.getSourceFile().getFilePath().includes(`/${specifierTail}`),
-			);
-			if (match) return match;
-		}
-	}
-
-	return candidates[0];
+function resolveClass(expr: Node | undefined): ClassDeclaration | undefined {
```
goetzrobin, line 218: `resolveClass` (6 lines) is what replaced `resolveClassByName` + the name-based half of `resolveBaseClass` (~80 lines). It asks the type system for the declaration behind an `extends` / `directive:` reference — resolving a **symbol**, not matching a string. That's why the same-name collision problem (and the `candidates[0]` arbitrary fallback) simply doesn't exist here.

libs/tools/src/executors/docs/generate-ui-docs/executor.ts
```diff
@@ -350,18 +293,11 @@ function collectMembers(cls: ClassDeclaration, componentInfo, seen: Set<string>)
...
-					if (match && match[1]) {
-						defaultValue = match[1].replace(/['"]/g, '');
-					} else {
-						defaultValue = argText.replace(/['"]/g, '');
-					}
-				}
+				// Only the plain `input`/`model` forms carry a default; the `*.required`
+				// forms take their options object as the first argument instead. Keep the
+				// default exactly as written so strings stay quoted and expressions intact.
+				const hasDefault = !isRequired && (exprText === 'input' || exprText === 'model') && callArgs.length > 0;
```
goetzrobin, line 299: Default value is now taken **verbatim** from the argument node. The previous `/(([^()]*))$/` regex + blanket `replace(/['"]/g, '')` had two issues: `input(signal(5))` was documented as `5` (it grabbed the paren contents), and `input('')` became `""` — indistinguishable from *no default*. Now strings keep their quotes and expressions stay intact. Note: this changes the emitted `defaultValue` strings (the snapshot strips `defaultValue`, so it stays green; the docs table renders whatever we emit).

libs/tools/src/executors/docs/generate-ui-docs/extract-inputs-outputs.spec.ts
```diff
@@ -426,5 +538,38 @@ describe('extractInputsOutputs', () => {
 			// Must inherit from the brain Base, not the helm one that shares the name.
 			expect(names(result['BrnDialog'].inputs)).toEqual(['own', 'fromBrain']);
 		});
+
+		it('does not resolve to a sibling package whose path is a prefix substring', () => {
```
goetzrobin, line 532: This is the regression test for the registry's silent-wrong-resolution bug: `'/brain/dialog'` is a substring of `'/brain/dialog-content'`, so the old `includes()` tail-match resolved `@spartan-ng/brain/dialog` to the wrong package. It **fails** on the registry approach and passes with symbol resolution.

libs/tools/src/executors/docs/generate-ui-docs/extract-inputs-outputs.spec.ts
```diff
@@ -116,6 +151,38 @@ describe('extractInputsOutputs', () => {
 				{ name: 'size', type: 'number', description: '', defaultValue: null, required: true },
 			]);
 		});
+
+		it('keeps default values exactly as written, including string quotes and expressions', () => {
```
goetzrobin, line 145: Characterization test for the default-value change: `'dialog'` stays quoted and `signal(5)` stays verbatim (previously `5`).

goetzrobin (commented):
A few inline pointers to make the diff faster to read — happy to discuss any of them. 🙏

goetzrobin:
Thanks @coderabbitai and Greptile — both notes were valid and are addressed in 9e69b0f:

1. **Fixture `ComponentInfo` type** (both bots): `selector`/`exportAs` are now optional (`?: string | null`) to match the extractor's real contract — they're left `undefined` when the decorator omits them, which is exactly what the `toBeUndefined()` assertion checks.
2. **Barrel/paths fixture regex** (CodeRabbit): broadened from `(?:brain|helm)` to `libs/<scope>/<pkg>` so it matches the helper's own doc comment and won't silently skip a fixture placed under another scope.

Both are test-harness-only; the 22 unit tests + the real-output snapshot stay green.

MerlinMoos (approved):
LGTM thanks for improving this feature

PR #1621, opened 2026-07-08
fix(cli): harden dependency resolution and scaffold integrity

### LINKED ISSUE #1613: Cannot generate ui

ng --version
22.0.5
node -v
v24.15.0
pnpm -v
11.10.0

node installed via fnm, and pnpm installed via npm i -g on latest macos

cli

### Description

generated new ng 22 app, then installed spartan and ran :init then tried to run :ui as acoording to docs, but got error "Cannot read properties of null (reading 'startsWith')".

```true
ng g @spartan-ng/cli:ui
? Choose which primitives you want to copy …
? Choose which primitives you want to copy …
✔ Choose which primitives you want to copy · all
Cannot read properties of null (reading 'startsWith')
```

### Other information

Fresh angular application and i updated the angular cli RIGHT before generating the app and adding spartan

- [x] Yes
- [ ] No

### LINKED ISSUE #1618: Command generator misses transitive input-group dependency, and healthcheck does not detect broken scaffold

- @spartan-ng/cli: 1.0.2
- Angular: 21 (workspace using Angular CLI)
- Package manager: pnpm
- OS: Linux

command

### Description

Running the Spartan UI generator for command creates code that imports input-group, but input-group is not auto-generated as a dependency.
Additionally, running healthcheck does not report this inconsistency.

### Reproduction

1. Run: `ng g @spartan-ng/cli:ui command`
2. Inspect generated command input component.
Observe the generated import of input-group: hlm-command-input.ts
3. Check generated path mappings: tsconfig.json
4. Run: `ng g @spartan-ng/cli:healthcheck`

### Actual Behavior

- command scaffold imports input-group: hlm-command-input.ts
- But input-group is not generated/added when generating command.
- In CLI internals, command dependency map only includes utils and button: primitive-deps.js
- Yet command template explicitly imports input-group: hlm-command-input.ts.template
- healthcheck does not catch this mismatch:
- Registered checks in generator.js
- No primitive dependency/path consistency check in healthcheck files: healthchecks

### Expected Behavior

- command generator should include input-group as a dependent primitive (directly or transitively), or
- command template should not import input-group unless it is guaranteed to exist.
- healthcheck should report/fix generated primitive dependency inconsistencies (for example: imports to helm primitives that were not scaffolded or mapped).

```true
`hlm-input-group` cannot be found in hlm-command-input.ts
```

### Other information

_No response_

- [ ] Yes
- [x] No

goetzrobin:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature

### Others

- [x] cli

CLI hardening covering two issues:

- **#1613** – On a workspace that does not pin `@spartan-ng/cli`, `ng g @spartan-ng/cli:ui`
  crashes with `Cannot read properties of null (reading 'startsWith')`. The derived
  `@spartan-ng/brain` version resolves to `null`, and on pnpm that `null` reaches nx's catalog
  check (`version.startsWith('catalog:')`).
- **#1618** – Generating `command` on its own scaffolds code importing `input-group`, which was
  never declared as a dependency (so it is never generated or path-mapped). The healthcheck did
  not detect the broken scaffold.

Closes #1613
Closes #1618

## What is the new behavior?

- **brain version fallback** – `@spartan-ng/brain` now falls back to the running cli's own version
  when the workspace does not pin `@spartan-ng/cli`, so it is always a valid, lockstep specifier
  instead of `null`. Composes with the catalog-dependency filter from #1610.
- **missing primitive deps** – the dependency map now declares every primitive its templates import.
  Auditing this surfaced four more undeclared deps beyond `command`:
  - `command` → `input-group`, `dialog`
  - `date-picker` → `input-group`
  - `dialog` → `button`
  - `tabs` → `button`
  - `tooltip` → `utils`
- **drift guard** – a new test fails if a template imports a primitive the dependency map does not
  declare, so this class of bug cannot regress.
- **scaffold-integrity healthcheck** – reports when a generated component imports a primitive via the
  import alias with no tsconfig path mapping (referenced but never generated) and points the user at
  the command to add it.

Verified with `nx test cli` (188 passed), `nx lint cli`, `nx build cli`, and a dist runtime check
that the self-version read resolves in the published-package layout.

- [ ] Yes
- [x] No

## Other information

The self-version fallback for #1613 only triggers when the workspace has no cli pinned, so it is
covered by a fast `buildDependencyArray` unit test rather than the Verdaccio smoke matrix (which
always installs the cli and so cannot reach that branch).

## Summary by CodeRabbit

* **New Features**
  * Added a “Scaffold integrity” workspace healthcheck that detects missing primitive `tsconfig` path mappings and offers remediation guidance.
  * Improved generated dependency resolution to fall back to the currently running CLI version when the workspace doesn’t pin it.
* **Bug Fixes**
  * Updated the primitive dependency graph to better match template import relationships.
* **Tests**
  * Expanded coverage for version fallback behavior (including semver validation) and for scaffold integrity + template “drift guard” checks.


MerlinMoos (approved):
LGTM thanks

PR #1622, opened 2026-07-08
fix(resizable): stop NG0103 feedback loop from non-idempotent layout sync

### LINKED ISSUE #1614: `resizable` — intermittent `NG0103` (infinite change detection) while dragging a handle, in zoneless apps

- `@spartan-ng/brain`: 1.0.4
- `@angular/core`: 22.0.5
- Zoneless (no `zone.js`, no `provideZonelessChangeDetection` call needed — Angular 22 default)
- Reproduces with the **official docs example** (nested horizontal/vertical `hlm-resizable-group`), no custom `minSize`/`maxSize` needed to trigger it.

resizable

### Description

1. Use the resizable example from the docs verbatim:
   ```html
   <hlm-resizable-group class="h-[200px] w-[500px] max-w-md rounded-lg border">
     <hlm-resizable-panel><div class="flex h-full items-center justify-center p-6">One</div></hlm-resizable-panel>
     <hlm-resizable-handle />
     <hlm-resizable-panel>
       <hlm-resizable-group direction="vertical">
         <hlm-resizable-panel><div class="flex h-full items-center justify-center p-6"><span class="font-semibold">Two</span></div></hlm-resizable-panel>
         <hlm-resizable-handle />
         <hlm-resizable-panel><div class="flex h-full items-center justify-center p-6"><span class="font-semibold">Three</span></div></hlm-resizable-panel>
       </hlm-resizable-group>
     </hlm-resizable-panel>
   </hlm-resizable-group>
   ```
2. Drag the horizontal handle with ordinary mouse movement (any slowdown/pause in the drag reproduces it more reliably; a Playwright-driven mousemove loop with occasional zero-delta frames reproduces it consistently).
3. Sometimes (not every drag) the panels snap to an extreme, and the console shows:
   ```
   ERROR RuntimeError: NG0103: Infinite change detection while refreshing application views. Ensure views are not calling `markForCheck` on every template execution or that afterRender hooks always mark views for check.
   ```

In `BrnResizableGroup._synchronizePanelSizes()`:

```js
const hasCompleteExternalLayout = !isInitialLayout &&
    currentLayout.length === totalPanels &&
    !this._layoutsEqual(currentLayout, this._appliedLayout);
```

When `_setLayout()` is called with a layout that is **already equal** to `_appliedLayout` (e.g. two consecutive drag frames compute the same percentages — completely normal, happens whenever mouse velocity momentarily drops to ~0), `_layoutsEqual(currentLayout, this._appliedLayout)` is `true`, so `hasCompleteExternalLayout` becomes `false`. The function then falls into the **"membership changed"** branch even though membership did not change:

```js
const rawSizes = panels.map((panel) => newPanelSizes.get(panel) ?? previousSizes.get(panel) ?? 0);
const totalSize = rawSizes.reduce((total, size) => total + size, 0);
sizes = totalSize > 0 ? rawSizes.map((size) => (size / totalSize) * 100) : rawSizes;
```

This recomputes each size via a `x / total * 100` round-trip, which is **not guaranteed to be bit-identical** to the input due to floating point rounding. Concretely, we logged (instrumenting the live component instance, no source patch):

```
setLayout sizes=[50.669642857142854, 49.330357142857146]
sync: currentLayout === appliedLayout  →  true
setLayout sizes=[50.66964285714285,  49.330357142857146]   // differs in the 15th significant digit
```

Since the recomputed `sizes` differs (however marginally) from `currentLayout`, `_layoutsEqual(currentLayout, sizes)` is `false`, so `_setLayout(sizes)` is called again — re-triggering the `afterRenderEffect` that calls `_synchronizePanelSizes`. Usually this converges after one extra iteration, but occasionally (depending on drag timing / how many other effects are competing for the same change-detection pass) it doesn't converge within Angular's zoneless retry budget for `afterRenderEffect`s, and Angular throws `NG0103`.

This matches the general "signal write in `afterRender`/`afterRenderEffect` must be a *stable, idempotent* value or it loops forever" footgun (see https://justangular.com/blog/protect-your-app-from-infinite-change-detection) — here the non-idempotence comes from floating-point arithmetic rather than an `update()` call, but the failure mode is the same.

```true
ERROR RuntimeError: NG0103: Infinite change detection while refreshing application views. Ensure views are not calling `markForCheck` on every template execution or that afterRender hooks always mark views for check.
```

### Other information

Clauds suggestion: 

`_synchronizePanelSizes` should short-circuit when `currentLayout` already equals `_appliedLayout` (or `panels.length === this._knownPanels.length` with matching identities) *before* doing any recompute — i.e. treat "no external layout change AND no membership change" as a no-op, rather than falling through to the proportional-rescale branch. That branch's math should also be guarded to skip when `previousPanelsTotal === totalSize` for the unchanged panels, to avoid re-deriving values that were already correct.

- [ ] Yes
- [x] No

goetzrobin:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [x] resizable

Closes #1614

Dragging a `hlm-resizable-handle` in a zoneless app (Angular's default) intermittently throws
`NG0103: Infinite change detection while refreshing application views`, sometimes snapping panels to
an extreme.

Root cause is in `BrnResizableGroup._synchronizePanelSizes()`. The `afterRenderEffect` re-runs
whenever `layout()` changes. During a drag, `_setLayout()` sets both `layout` and `_appliedLayout`
to the same value, so the effect re-fires with `currentLayout` already equal to `_appliedLayout`.
`hasCompleteExternalLayout` is therefore `false`, and — because it is not the initial render —
control falls into the "membership changed" branch even though membership did **not** change. That
branch recomputes every size via `size / total * 100`, which is **not bit-identical** for fractional
layouts:

```
50.669642857142854 / 100 * 100  →  50.66964285714285   // differs in the last digit
```

The drifted result differs from `currentLayout` under the exact (`===`) guard, so `_setLayout()`
runs again, re-triggering the effect. This normally converges in one extra pass, but occasionally
does not converge within Angular's zoneless retry budget → `NG0103`. Reproduces with the official
docs example (nested horizontal/vertical group), most reliably when the drag momentarily pauses
(mouse velocity ~0, so two consecutive frames compute identical percentages).

## What is the new behavior?

Two independent layers of protection:

1. **Membership guard** — `_synchronizePanelSizes` now short-circuits as a no-op when neither the
   external layout nor the panel membership changed, so it never enters the proportional-rescale
   branch spuriously. That branch now runs only on a genuine membership change, which is what its
   comment always described.
2. **Tolerance-based comparison** — `_layoutsEqual` is renamed to `_layoutsEquivalent` and compares
   with an epsilon (`1e-9`) instead of exact `===`. Percentages this close render identically, so
   sub-nanopercent floating-point drift from the rescale round-trip can no longer register as a
   change and feed back into the model — even inside the legitimate membership-change branch.

Added a regression test that re-applies an already-applied fractional layout (a zero-velocity drag
frame) and asserts the model does not drift. It fails on `main` (`50.66964285714285` leaks back) and
passes with this change. All 417 brain tests pass; lint, format, and build are green.

- [ ] Yes
- [x] No

## Other information

The membership guard alone fixes the reported loop; the epsilon comparison is defense-in-depth for
the same class of floating-point feedback in the rescale branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved layout stability when reapplying the same resize proportions, preventing tiny floating-point shifts.
  * Fixed cases where resizing panels could unexpectedly drift or trigger repeated update loops in complex rendering setups.
  * Kept panel layouts unchanged when no meaningful change was detected, making resizable areas more reliable.

### DIFF
--- a/libs/brain/resizable/src/lib/brn-resizable-group.spec.ts
+++ b/libs/brain/resizable/src/lib/brn-resizable-group.spec.ts
@@ -273,6 +273,28 @@ describe('BrnResizableGroup', () => {
 		expect(flexValues).toEqual(['25 1 0', '75 1 0']);
 	});
 
+	it('is idempotent when re-applying the already-applied fractional layout (no NG0103 feedback)', async () => {
+		// A drag frame that computes the same percentages as the last one re-sets the model to a
+		// value that already equals the applied layout. Re-synchronizing must be a no-op: the
+		// proportional-rescale branch round-trips through x/total*100, which is not bit-identical for
+		// these fractions, and any drift feeds back into the model and can loop until Angular throws
+		// NG0103 in zoneless apps.
+		const fractional = [50.669642857142854, 49.330357142857146];
+		const view = await render(DynamicResizableHost);
+		view.fixture.componentInstance.layout.set([...fractional]);
+		view.detectChanges();
+		await view.fixture.whenStable();
+		expect(view.fixture.componentInstance.layout()).toEqual(fractional);
+
+		// Re-apply the identical percentages via a fresh array reference (a zero-velocity drag frame).
+		view.fixture.componentInstance.layout.set([...fractional]);
+		view.detectChanges();
+		await view.fixture.whenStable();
+
+		// No floating-point drift leaked back into the model.
+		expect(view.fixture.componentInstance.layout()).toEqual(fractional);
+	});
+
 	it('applies an external layout update without a panel membership change', async () => {
 		const view = await render(DynamicResizableHost);
 		view.fixture.componentInstance.layout.set([35, 65]);
--- a/libs/brain/resizable/src/lib/brn-resizable-group.ts
+++ b/libs/brain/resizable/src/lib/brn-resizable-group.ts
@@ -96,7 +96,16 @@ export class BrnResizableGroup {
 		const hasCompleteExternalLayout =
 			!isInitialLayout &&
 			currentLayout.length === totalPanels &&
-			!this._layoutsEqual(currentLayout, this._appliedLayout);
+			!this._layoutsEquivalent(currentLayout, this._appliedLayout);
+		const membershipChanged = !isInitialLayout && !this._panelsEqual(panels, this._knownPanels);
+
+		// Nothing external changed and the same panels are still mounted: the applied layout is already
+		// correct. Falling through to the proportional-rescale branch would round-trip each size through
+		// x/total*100, which is not bit-identical for fractional layouts; the drift feeds back into the
+		// model and can loop until Angular throws NG0103 in zoneless apps.
+		if (!isInitialLayout && !hasCompleteExternalLayout && !membershipChanged) {
+			return;
+		}
 
 		let sizes: number[];
 		if (isInitialLayout) {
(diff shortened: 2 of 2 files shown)


marcjulian (approved):
The NG0103 error is now gone. 

Probably outside of the scope for this PR: 
I noticed in main and also this branch that content can collapse when the handle is dragged to one end. Only in mobile size. https://spartan.ng/components/resizable#examples__handle

Image

Image

goetzrobin:
@marcjulian do you want to create an issue for that

PR #1666, opened 2026-07-28
fix(combobox): skip scrollIntoView when activated by pointer event, c…

### LINKED ISSUE #1658: [combobox] List scrolls on its own when the pointer nears an edge

| | |
|---|---|
| `@spartan-ng/brain` | `1.1.1` and `1.1.2` (identical code in both) |
| `@angular/core` | `22` |
| Style | `default` |

combobox

### Description

`BrnComboboxItem` scrolls itself into view every time it becomes active, and the pointer makes an item active on `mouseenter`. The result is that hovering near the top or bottom edge of the list makes the list scroll on its own.

```ts
// spartan-ng-brain-combobox.mjs
host: { listeners: { 'mouseenter': 'activate()' } }

activate() {
  if (this._disabled()) return;
  this._combobox.keyManager.setActiveItem(this);   // -> setActiveStyles()
}

setActiveStyles() {
  this._highlighted.set(true);
  // ensure the item is in view
  if (isPlatformBrowser(this._platform)) {
    this._elementRef.nativeElement.scrollIntoView({ block: 'nearest' });
  }
}
```

`block: 'nearest'` is not a no-op for the row straddling an edge — it pulls that row fully into view. That slides a *different* half-row under a cursor that never moved, which activates and scrolls again. So a sliver a few pixels tall makes the list creep whenever the pointer approaches an edge.

The comment says "ensure the item is in view", which is an active-descendant concern: when the **keyboard** moves the highlight, the highlighted row has to stay visible. The pointer path inherits it only because `activate()` and keyboard navigation both funnel through the same `keyManager.setActiveItem`. When the pointer is what made the item active, the item is by definition already under the cursor and needs no scrolling.

**Steps to reproduce**

1. Open a combobox with enough items to scroll.
2. Scroll so the list rests off a row boundary (any normal wheel scroll does this).
3. Move the pointer onto the partially visible row at the bottom edge.

The list jumps. Measured on a 500-item list with 28px rows and a 221px viewport, at four scroll offsets:

| `scrollTop` | jump on hover |
|---|---|
| 137 | 13.5px |
| 201 | 5.5px |
| 263 | 27.5px |
| 319 | 27.5px |

**Why this reads differently in `select`**

`BrnSelectItem` has the identical code, but the select does not feel this way because `BrnSelectScrollUp`/`BrnSelectScrollDown` are sticky, opaque and pinned to the edges, so a partially visible option is never hoverable. Where edge scrolling *is* intended there, it is deliberate and rate-limited — `interval(15)` with `scrollBy({ top: ±8 })`, cancelled on `mouseleave`. The combobox has no such affordance, so the same line only produces an unrequested lurch.

**Suggested fix**

Skip the scroll when the activation came from the pointer, leaving keyboard navigation untouched:

```diff
+ _activatedByPointer = false;
  setActiveStyles() {
      this._highlighted.set(true);
-     if (isPlatformBrowser(this._platform)) {
+     if (!this._activatedByPointer && isPlatformBrowser(this._platform)) {
          this._elementRef.nativeElement.scrollIntoView({ block: 'nearest' });
      }
  }

  activate() {
      if (this._disabled()) return;
+     this._activatedByPointer = true;
      this._combobox.keyManager.setActiveItem(this);
+     this._activatedByPointer = false;
  }
```

Hover still highlights, the keyboard still keeps its row in view. We are running exactly this as a `pnpm patch` against 1.1.1 and it resolves the jitter with no other behaviour change.

### Other information

Overriding this from user code does not appear to be possible. `setActiveStyles()` is public and `_highlighted` is protected, so subclassing `BrnComboboxItem` looks like the intended seam — but Angular matches `contentChildren(BrnComboboxItem)` against directive definitions rather than element-injector providers, so a subclass aliased with `{ provide: BrnComboboxItem, useExisting: ... }` leaves `BrnCombobox` with an empty item list and the panel renders its empty state over a full list. That is why this is filed rather than worked around locally.

marcjulian:
…loses #1658

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [x] combobox

Closes #1658, which can be reproduced with the [Groups](https://spartan.ng/components/combobox#group) example  by moving the mouse to `London` and it moves into view.

## What is the new behavior?

Mouse pointer doesn't scroll items into view, only when activated by keyboard. Try again to move the mouse to `London` and the item doesn't move into view on mouse pointer, but when using the keyboard

- [ ] Yes
- [x] No

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved combobox pointer interactions by preventing unnecessary scrolling when selecting an item with the mouse or other pointer input.
  * Preserved automatic scrolling for keyboard-based navigation.

### DIFF
--- a/libs/brain/combobox/src/lib/brn-combobox-item.ts
+++ b/libs/brain/combobox/src/lib/brn-combobox-item.ts
@@ -54,6 +54,8 @@ export class BrnComboboxItem<T> implements Highlightable {
 
 	protected readonly _highlighted = signal(false);
 
+	private _activatedByPointer = false;
+
 	/** @internal Determine if this item is visible based on the current search query */
 	public readonly visible = computed(() => {
 		return this._combobox.filter()(
@@ -68,7 +70,7 @@ export class BrnComboboxItem<T> implements Highlightable {
 		this._highlighted.set(true);
 
 		// ensure the item is in view
-		if (isPlatformBrowser(this._platform)) {
+		if (!this._activatedByPointer && isPlatformBrowser(this._platform)) {
 			this._elementRef.nativeElement.scrollIntoView({ block: 'nearest' });
 		}
 	}
@@ -95,6 +97,8 @@ export class BrnComboboxItem<T> implements Highlightable {
 			return;
 		}
 
+		this._activatedByPointer = true;
 		this._combobox.keyManager.setActiveItem(this);
+		this._activatedByPointer = false;
 	}
 }


MerlinMoos (approved):
LGTM

PR #1703, opened 2026-08-15
fix(avatar): support ngSrc on avatar image

### LINKED ISSUE #1702: add support for `ngSrc` on avatar components

avatar

### Information

At the moment, when using the Angular directive `ngSrc` on an avatar component like so:
```html
<hlm-avatar class="rounded-lg">
  <img alt="${userName()}'s avatar" fill hlmAvatarImage [ngSrc]="avatarUrl()" />
  <span class="rounded-lg bg-[#FD005B] text-white" hlmAvatarFallback>FD</span>
</hlm-avatar>
````
The displayed avatar is always the fallback avatar ("FD" in this case) even if the `avatarUrl()` signal point to an existing image.

It would be great if the avatar component can support the usage of the `NgOptimizedImage` directive.

This would also avoid this kind of warning in code:

Image

Here is how I do it until now:
```html
<hlm-avatar class="rounded-lg">
  <!-- eslint-disable-next-line @angular-eslint/template/prefer-ngsrc -- hlmAvatar and hlmAvatarImage does not seem to work with ngSrc -->
  <img alt="${userName()}'s avatar" hlmAvatarImage [src]="avatarUrl()" />
  <span class="rounded-lg bg-[#FD005B] text-white" hlmAvatarFallback>FD</span>
</hlm-avatar>
```
I always need to disable the "prefer-ngsrc" eslint rule on my projects.

- [ ] Yes
- [ ] No
- [x] Not sure yet ;)

wajrock:
Fixes #1702

The avatar image was only rendered once the load state became true, but the load state could only become true once the image had already loaded. This circular dependency prevented ngSrc from working since NgOptimizedImage relies on lazy loading.

The image is now always rendered and hidden with CSS until it finishes loading, so the load event can fire normally regardless of whether src or ngSrc is used.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved avatar image loading so fallback content appears only when the image cannot be displayed.
  * Ensured avatar images are correctly positioned and hidden until ready to show.
  * Refined avatar styling for more consistent image sizing, positioning, and appearance.

* **Tests**
  * Expanded coverage for image loading, fallback rendering, visibility, and positioning.
  * Updated styling expectations to reflect the improved avatar layout.

### DIFF
--- a/libs/brain/avatar/src/lib/image/brn-avatar-image.ts
+++ b/libs/brain/avatar/src/lib/image/brn-avatar-image.ts
@@ -1,4 +1,4 @@
-import { Directive, ElementRef, afterNextRender, computed, inject, signal } from '@angular/core';
+import { Directive, computed, signal } from '@angular/core';
 
 @Directive({
 	selector: 'img[brnAvatarImage]',
--- a/libs/brain/avatar/src/lib/image/brn-avatar-image.ts
+++ b/libs/brain/avatar/src/lib/image/brn-avatar-image.ts
@@ -1,4 +1,4 @@
-import { Directive, computed, signal } from '@angular/core';
+import { Directive, ElementRef, afterNextRender, computed, inject, signal } from '@angular/core';
 
 @Directive({
 	selector: 'img[brnAvatarImage]',
@@ -6,6 +6,8 @@ import { Directive, computed, signal } from '@angular/core';
 	host: {
 		'(load)': '_onLoad()',
 		'(error)': '_onError()',
+		class: 'absolute inset-0',
+		'[class.invisible]': '!canShow()',
 	},
 })
 export class BrnAvatarImage {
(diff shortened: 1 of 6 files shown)


libs/brain/avatar/src/lib/image/brn-avatar-image.ts
```diff
@@ -6,6 +6,8 @@ import { Directive, computed, signal } from '@angular/core';
 	host: {
 		'(load)': '_onLoad()',
 		'(error)': '_onError()',
+		class: 'absolute inset-0',
```
marcjulian: Brain components are not allowed to use tailwind classes as they are hidden in node_modules.

wajrock: fixed

PR #1704, opened 2026-08-17
ci: replace wagoid action with inline commitlint, stagger format-and-lint

dowsprogress:
## Problem
CI was failing repeatedly on push with 429 (Too Many Requests) and 503 errors when GitHub Actions tried to download external action archives at job startup:
- `nrwl/nx-set-shas` — hit by all 4 jobs simultaneously at `t=0`
- `wagoid/commitlint-github-action` — 503 on codeload CDN

## Changes
1. **Replaced `wagoid/commitlint-github-action`** with an inline `pnpm commitlint` run. `@commitlint/cli` is already a dev dependency so no new packages needed. Eliminates one external archive download entirely.

2. **Added `needs: [commitlint]` to `format-and-lint`** to stagger the start of `nrwl/nx-set-shas` downloads. Previously all 4 jobs hit the codeload endpoint simultaneously; now they start in two waves (build/unit/e2e together, then format-and-lint ~1-2 min later), reducing the concurrent burst from 4 to 3.

## Summary by CodeRabbit

* **Chores**
  * Improved automated commit message validation in the CI workflow.
  * Added dependency installation and caching to support consistent checks.
  * Updated job sequencing so formatting and lint checks run after commit validation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,25 +26,35 @@ jobs:
     permissions:
       actions: read
       contents: read
-      # Required by wagoid/commitlint-github-action
       pull-requests: read
     steps:
       - uses: actions/checkout@v6
         with:
-          # Required by wagoid/commitlint-github-action
           fetch-depth: 0
+      - name: Install pnpm
+        uses: pnpm/action-setup@v5
       - name: Install Node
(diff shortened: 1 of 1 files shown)


dowsprogress:
pushed in error

PR #1709, opened 2026-08-21
fix: forward the scroll strategy input through the helm overlay wrappers

khocef:
`BrnOverlay` declares a `scrollStrategy` input, and `getScrollStrategy()` already accepts `'close'`, `'reposition'`, or a strategy object. None of that is reachable from an app. `hlm-select`, `hlm-select-multiple`, `hlm-combobox`, `hlm-autocomplete` and `hlm-popover` all leave it out of their host directive inputs, so the reposition default is the only behaviour available.

Where this bites is a trigger inside a scrollable container, under a sticky header. Once the container is registered with `cdkScrollable`, the overlay follows its trigger correctly, and then keeps following it under the header, drawn on top of it, because the CDK measures the whole viewport and knows nothing about app chrome. Measured against a 56px header, the overlay reached `y=-445` while still painting over it. Closing on scroll, or repositioning with `autoClose`, would both fix that, and neither was reachable.

This forwards the existing input on the five wrappers. Nothing changes by default: with no binding the input still falls back to the strategy the tokens provide, which is `reposition` for popover-based overlays.

The new story binds `scrollStrategy='close'`, and the spec scrolls the page while the overlay is open. Without the forwarding the binding is silently ignored (Storybook compiles templates JIT, so there is no build error), the overlay stays put, and the spec fails on the content still being in the DOM. Against `main` that is 6 passing and 1 failing; with the change, 7 passing.

One note on the spec: it dispatches the scroll event explicitly. Cypress scrolls by assigning `scrollTop`, which never reaches the listener `ScrollDispatcher` installs, so without that the strategy would never run and the spec would pass without measuring anything.

### DIFF
--- a/apps/ui-storybook-e2e/src/integration/popover/popover.cy.ts
+++ b/apps/ui-storybook-e2e/src/integration/popover/popover.cy.ts
@@ -166,3 +166,20 @@ describe('popover--near-viewport-edge', () => {
 		});
 	});
 });
+
+describe('popover--close-on-scroll', () => {
+	beforeEach(() => {
+		cy.visit('/iframe.html?id=popover--close-on-scroll');
+		cy.get('#closing-trigger').click();
+		cy.get('hlm-popover-content').should('be.visible');
+	});
+
+	it('honours the scroll strategy the wrapper now forwards', () => {
+		// Cypress scrolls by assigning `scrollTop`, which never reaches the listener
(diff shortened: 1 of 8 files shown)


marcjulian:
@khocef thanks for adding them as input. An alternative is to use `provideBrnPopoverDefaultOptions({ scrollStrategy: 'close' })`, its good to have both options though to opt out from the default.

marcjulian (approved):

PR #1711, opened 2026-08-24
fix(combobox): dont reopen popover when item is selected via enter

### LINKED ISSUE #1685: BrnComboboxInput: Enter commits, closing the panel, then the same handler re-opens it
### Package / version

`@spartan-ng/brain` **1.3.0**

Verified in the **shipped bundle**, not inferred from the repo source:
`node_modules/@spartan-ng/brain/fesm2022/spartan-ng-brain-combobox.mjs`.

### Summary

In `BrnComboboxInput.onKeyDown`, pressing **Enter** commits the selection — which closes the panel — and then the **same handler** re-reads whether the panel is open and re-opens it, a few lines later.

The user sees the list flash shut and spring back. The selected value is correct; only the panel state is wrong.

### The sequence, in the order the handler executes it

`BrnComboboxInput.onKeyDown` (shipped bundle, ~line 601):

```js
onKeyDown(event) {
    if (event.key === 'Enter') {
        event.preventDefault();
        this._combobox.selectActiveItem();   // (1) commits -> closes the panel
    }
    if (this._isExpanded()) {                // (2) re-reads state, now false
        ...
    }
    else {
        if (event.key === 'Enter' || event.key === 'ArrowDown' || event.key === 'ArrowUp') {
            this._combobox.open();           // (3) re-opens the panel just closed
        }
        ...
    }
    this._combobox.keyManager.onKeydown(event);
}
```

Step (1) closes via `BrnCombobox.selectActiveItem()` → `select(value)`:

```js
select(itemValue) {
    this.value.set(itemValue);
    this._onChange?.(itemValue);
    if (this.closeOnSelect()) {   // default is true (BRN_COMBOBOX config: closeOnSelect: true)
        this.close();
    }
}
```

So with the default `closeOnSelect: true`, the `else` in step (2) is entered precisely *because* step (1) succeeded, and `Enter` is in the re-open arm.

### Reproduction

1. Single-select `brnCombobox` with `brnComboboxInput`, default options.
2. Focus the input, arrow down to highlight an item.
3. Press **Enter**.

**Expected:** the item is selected and the panel stays closed.
**Actual:** the item is selected, the panel closes, and it immediately re-opens.

### Evidence that this is an oversight rather than intent

The sibling `BrnComboboxChipInput` in the same file (~line 364) has the same structure, and **its re-open arm deliberately excludes `Enter`**:

```js
if (!this._isExpanded()) {
    if (event.key === 'ArrowDown' || event.key === 'ArrowUp') {   // no 'Enter'
        this._combobox.open();
    }
    ...
}
```

That is the correct set for an input where Enter has just committed. `BrnComboboxInput` including `Enter` in the same arm looks like the divergence.

### Suggested fix

Either capture the expanded state **before** the commit and branch on that, or drop `Enter` from the re-open arm once a commit has occurred in this event — matching what `BrnComboboxChipInput` already does.

Note that `Enter` is presumably in that arm to let Enter *open* a closed panel. If that behaviour is wanted, it needs to distinguish "closed when the handler started" from "closed by this handler", which capturing the state up front gives you for free.

### Workaround in use

We compose around it rather than patching `node_modules`, and the composition is mutation-covered so it cannot later be deleted as ceremony by someone who mistakes it for a no-op. When the ordering is fixed upstream, the workaround goes.

Happy to open a PR if the `BrnComboboxChipInput`-style fix is the direction you'd want.

skolldev:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [ ] accordion
- [ ] alert
- [ ] alert-dialog
- [ ] aspect-ratio
- [ ] attachment
- [ ] autocomplete
- [ ] avatar
- [ ] badge
- [ ] breadcrumb
- [ ] bubble
- [ ] button
- [ ] button-group
- [ ] calendar
- [ ] card
- [ ] carousel
- [ ] checkbox
- [ ] collapsible
- [x] combobox
- [ ] command
- [ ] context-menu
- [ ] data-table
- [ ] date-picker
- [ ] dialog
- [ ] empty
- [ ] dropdown-menu
- [ ] field
- [ ] hover-card
- [ ] icon
- [ ] input
- [ ] input-group
- [ ] input-otp
- [ ] item
- [ ] kbd
- [ ] label
- [ ] marker
- [ ] menubar
- [ ] message
- [ ] native-select
- [ ] navigation-menu
- [ ] pagination
- [ ] popover
- [ ] progress
- [ ] radio-group
- [ ] resizable
- [ ] scroll-area
- [ ] select
- [ ] separator
- [ ] sheet
- [ ] sidebar
- [ ] skeleton
- [ ] slider
- [ ] sonner
- [ ] spinner
- [ ] switch
- [ ] table
- [ ] tabs
- [ ] textarea
- [ ] toggle
- [ ] toggle-group
- [ ] tooltip
- [ ] typography

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli
- [ ] mcp

With a Combobox that has `closeOnSelect` true (the default), clicking an item correctly closes the popover. However, pressing Enter on keyboard does not close the popover, or rather closes and immediately reopens.

Closes #1685 

## What is the new behavior?

A combobox with `closeOnSelect` true correctly closes the popover on Enter, while still opening via Enter if it was closed

- [ ] Yes
- [x] No

## Other information

The solution snapshots the expanded state at the moment Enter was pressed, and doesnt reopen the popover if it was open.

The original shadcn combobox correctly has the "new" behaviour: https://ui.shadcn.com/docs/components/base/combobox

## Bug Fixes

- Fixed keyboard behavior in comboboxes so pressing Enter on an expanded popup selects the active item and closes it without reopening.
- Pressing Enter on a collapsed combobox now opens it as expected.

## New Features

- Added support for configuring comboboxes to close automatically after selecting an option.

### DIFF
--- a/libs/brain/combobox/src/lib/brn-combobox-input.ts
+++ b/libs/brain/combobox/src/lib/brn-combobox-input.ts
@@ -104,6 +104,12 @@ export class BrnComboboxInput<T> {
 
 	/** Listen for keydown events */
 	protected onKeyDown(event: KeyboardEvent): void {
+		// Snapshot the expansion state BEFORE selecting, since `selectActiveItem()` may close the
+		// popover synchronously (closeOnSelect) and flip `_isExpanded()` to false mid-handler.
+		// Branching on this snapshot keeps "Enter opens a closed combobox" without re-opening the
+		// popover on the very Enter that just selected and closed it.
+		const wasExpanded = this._isExpanded();
+
 		if (event.key === 'Enter') {
 			// prevent form submission if inside a form
 			event.preventDefault();
@@ -111,7 +117,7 @@ export class BrnComboboxInput<T> {
 			this._combobox.selectActiveItem();
 		}
 
-		if (this._isExpanded()) {
+		if (wasExpanded) {
 			if (event.key === 'Tab') {
 				// In popup mode the input lives inside a CDK overlay which is appended to <body>.
 				// Without preventDefault the browser has nowhere to Tab to inside the overlay and
--- a/libs/brain/combobox/src/lib/brn-combobox-input.ts
+++ b/libs/brain/combobox/src/lib/brn-combobox-input.ts
@@ -104,10 +104,7 @@ export class BrnComboboxInput<T> {
 
 	/** Listen for keydown events */
 	protected onKeyDown(event: KeyboardEvent): void {
-		// Snapshot the expansion state BEFORE selecting, since `selectActiveItem()` may close the
-		// popover synchronously (closeOnSelect) and flip `_isExpanded()` to false mid-handler.
-		// Branching on this snapshot keeps "Enter opens a closed combobox" without re-opening the
-		// popover on the very Enter that just selected and closed it.
+		// Snapshot the expansion state before handling selection to avoid re-opening the popover if it is closed on select
 		const wasExpanded = this._isExpanded();
 
 		if (event.key === 'Enter') {
--- a/libs/brain/combobox/src/lib/brn-combobox-input.spec.ts
+++ b/libs/brain/combobox/src/lib/brn-combobox-input.spec.ts
@@ -1,5 +1,5 @@
 import { ChangeDetectionStrategy, Component, computed, signal } from '@angular/core';
-import { render, screen } from '@testing-library/angular';
+import { fireEvent, render, screen } from '@testing-library/angular';
 import { BrnComboboxContent } from './brn-combobox-content';
 import { BrnComboboxInput } from './brn-combobox-input';
 import { BrnComboboxBaseToken } from './brn-combobox.token';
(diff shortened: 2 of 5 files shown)


MerlinMoos (approved):

libs/brain/combobox/src/lib/brn-combobox-input.ts
```diff
@@ -104,14 +104,20 @@ export class BrnComboboxInput<T> {
 
 	/** Listen for keydown events */
 	protected onKeyDown(event: KeyboardEvent): void {
+		// Snapshot the expansion state BEFORE selecting, since `selectActiveItem()` may close the
+		// popover synchronously (closeOnSelect) and flip `_isExpanded()` to false mid-handler.
+		// Branching on this snapshot keeps "Enter opens a closed combobox" without re-opening the
+		// popover on the very Enter that just selected and closed it.
```
marcjulian: Keep the comment a bit shorter

marcjulian (changes requested):
Looks good to me! Thanks for fixing that. Can you add `closeOnSelect` to `HlmCombobox` and `HlmComboboxMultiple` for the host directive inputs?

libs/brain/combobox/src/lib/brn-combobox-input.ts
skolldev: Slimmed down

skolldev:
> Looks good to me! Thanks for fixing that. Can you add `closeOnSelect` to `HlmCombobox` and `HlmComboboxMultiple` for the host directive inputs?

Forwarded the inputs for both components and regenerated the snapshot!

marcjulian (approved):

PR #1757, opened 2026-09-18
fix: autocomplete and combobox accessibility

### LINKED ISSUE #1728: autocomplete: input never exposes aria-activedescendant, listbox has no accessible name, autocomplete="off" cannot be overridden

```html
<hlm-autocomplete [search]="search()" (searchChange)="search.set($event)">
  <hlm-autocomplete-input inputId="plz" />
  <hlm-autocomplete-content *hlmAutocompletePortal>
    <div hlmAutocompleteList>
      <hlm-autocomplete-item value="a">A</hlm-autocomplete-item>
      <hlm-autocomplete-item value="b">B</hlm-autocomplete-item>
    </div>
  </hlm-autocomplete-content>
</hlm-autocomplete>
```

Type a character, press ArrowDown. Resulting DOM:

```html
<input role="combobox" aria-expanded="true" aria-controls="brn-autocomplete-list-1" autocomplete="off" …>   <!-- no aria-activedescendant -->
<div role="listbox" id="brn-autocomplete-list-1">                                                          <!-- no aria-label(ledby) -->
  <hlm-autocomplete-item role="option" id="brn-autocomplete-item-1" data-highlighted aria-selected="false">A</hlm-autocomplete-item>
```

Expected behavior

```html
<input role="combobox" aria-expanded="true" aria-controls="brn-autocomplete-list-1" aria-activedescendant="brn-autocomplete-item-1" autocomplete="postal-code" …>
<div role="listbox" id="brn-autocomplete-list-1" aria-labelledby="plz-label">
```

autocomplete

### Description

`BrnAutocomplete` drives the option list with an `ActiveDescendantKeyManager`, but `BrnAutocompleteInput` never renders `aria-activedescendant`. The highlighted option is only marked with `data-highlighted` on the `role="option"` element. Screen readers therefore announce nothing while the user moves through the list with the arrow keys; NVDA stays silent until Enter selects an option. WCAG 2.1 SC 4.1.2 (Name, Role, Value).

Two related gaps in the same component:

1. `BrnAutocompleteList` (`role="listbox"`) only gets an accessible name when the consumer passes `ariaLabel`. There is no default wiring to the field label (`aria-labelledby`), so the popup is announced as an unnamed list.
2. `BrnAutocompleteInput` sets `autocomplete="off"` as a static host attribute. Consumers cannot set a purpose token (`postal-code`, `address-level2`, …) declaratively, which WCAG 2.1 SC 1.3.5 requires for fields collecting user data.

```true

```

### Other information

Suggested fix

- `BrnAutocompleteInput`: host binding `'[attr.aria-activedescendant]': '_isExpanded() ? _autocomplete.keyManager.activeItem?.id() ?? null : null'` (both `BrnAutocomplete` and `BrnAutocompleteList` own a key manager; expose the active item id as a signal).
- `BrnAutocompleteList`: default `aria-labelledby` to the id of the `BrnAutocompleteLabel` when present.
- `BrnAutocompleteInput`: turn `autocomplete` into an input with default `'off'` instead of a static host attribute.

- [x] Yes
- [ ] No

### LINKED ISSUE #1759: Combobox doesn't have disabled styles if it has a trigger instead of input

"@spartan-ng/cli": "^1.4.1",
"@angular/core": "^22.0.7",

1. Open the StackBlitz below
https://stackblitz.com/edit/5ddgxfdq?file=src%2Fapp%2Fexample.ts
2. Compare the variant with only `hlm-combobox-input` vs. the one with `hlm-combobox-trigger`

combobox

### Description

`hlm-combobox-trigger` doesn't apply any disabled styles when the combobox is disabled. The combobox correctly doesn't open (behavior is fine), but there's no visual feedback that it's disabled (no reduced opacity, no `cursor-not-allowed`, etc.).

This only happens when using `hlm-combobox-trigger`:

- `hlm-combobox-input` **without** a trigger → disabled styles are applied correctly
- `hlm-combobox-input` **with** `hlm-combobox-trigger` → disabled styles are **not** applied

I found this while wrapping `hlm-combobox` in a custom select component, but I was able to reproduce it using the code from the docs (see the reproduction link).

### Expected behavior

Disabled styles should be applied consistently, regardless of whether a `hlm-combobox-trigger` is used.

```true

```

### Other information

_No response_

- [ ] Yes
- [x] No

Nicoss54:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### Primitives

- [X] autocomplete
- [x] combobox

### Others

- [ ] trpc
- [ ] nx
- [ ] repo
- [ ] cli
- [ ] mcp

Closes #1728

Adds missing `data-disabled` for brn-combobox-trigger, closes #1759

- [ ] Yes
- [X] No

## Summary by CodeRabbit

* **New Features**
  * Added configurable native `autocomplete` support to autocomplete inputs, defaulting to `off`.
  * Improved screen-reader navigation by associating autocomplete and combobox inputs with the active option while panels are open.
  * Added accessible listbox labeling through `aria-label` and `aria-labelledby`.
  * Generated autocomplete components now include these accessibility options by default.

* **Tests**
  * Added accessibility coverage for autocomplete and autocomplete search components, including active-option announcements and list labeling.


Nicoss54:
@marcjulian sometimes the e2e test failed. Here it's on menu context that have no deal with the content i changed. Is there a possibility that there are some flacky test in the e2e part ?

Nicoss54:
@marcjulian about the minor comment of CodeRabbit review. Here is my thoughts:
 
 If contentChild(BrnAutocompleteLabel) defaults to descendants: true. Labels inside hlmAutocompleteGroup would have been picked up too. In a grouped list, the listbox would then get aria-labelledby pointing at the first group's label (for example "Fruits"), which is wrong. The label now has to be a direct child of the list:

marcjulian:
> @marcjulian sometimes the e2e test failed. Here it's on menu context that have no deal with the content i changed. Is there a possibility that there are some flacky test in the e2e part ?

Hey, yess sometimes e2e are flaky, you can ignore them!

libs/brain/autocomplete/src/lib/brn-autocomplete-list.ts
```diff
@@ -10,6 +11,7 @@ import { injectBrnAutocompleteBase } from './brn-autocomplete.token';
 		'[id]': 'id()',
 		'[attr.data-empty]': '!_visibleItems() ? "" : null',
 		'[attr.aria-label]': 'ariaLabel()',
+		'[attr.aria-labelledby]': '_ariaLabelledBy()',
```
marcjulian, line 14: The `BrnAutocompleteGroup` already gets aria-labelledby based on the BrnAutocompleteLabel.id

Does it need to be added here to the list as well? What if there are multiple groups w/ labels?

I checked against the base-ui combobox and autocomplete examples. Only the `<div role="group" aria-labelledby="...">` receives `aria-labelledby` instead of the `<div role="listbox">` 

https://ui.shadcn.com/docs/components/base/combobox#groups and https://base-ui.com/react/components/autocomplete#grouped

marcjulian (changes requested):
`aria-labelledby` is already wired up for the `BrnAutocompleteGroup` following base-ui. Should we remove it from brn-autocomplete-list again?

libs/brain/autocomplete/src/lib/brn-autocomplete-list.ts
marcjulian: Ignore my message.

This only applies aria-labelledby when a hlmAutocompleteLabel is a direct child of the hlmAutocompleteList, otherwise the aria-labelledby is managed by the hlmAutocompleteGroup.

marcjulian (approved):

Issue #1754, opened 2026-09-18
select: BrnSelectTrigger never emits aria-controls, and the helm listbox has no id to point at


MPolokIQ:
### Please provide the environment you discovered this bug in.

```
@spartan-ng/brain 1.4.1 / @spartan-ng/helm 1.4.1 (also 1.3.1 and main@94587d1)
Angular 21
```

An hlm-select as documented. DOM while open:

```
<button role="combobox" aria-haspopup="listbox" aria-expanded="true">   <!-- no aria-controls -->
<div class="cdk-overlay-pane" id="brn-overlay-1">
  <div role="listbox">                                                  <!-- no id -->
    <hlm-select-item role="option" id="brn-select-item-1">
```

### Which area/package is the issue in?

select

### Description

APG, combobox pattern:

`The combobox element has aria-controls set to a value that refers to the element that serves as the popup.`

BrnSelectTrigger declares role="combobox" and aria-haspopup="listbox" but never binds aria-controls ([brn-select-trigger.ts#L9-L23](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/brain/select/src/lib/brn-select-trigger.ts#L9-L23)), so nothing connects the combobox to its listbox. On the other side, neither BrnSelectList ([brn-select-list.ts](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/brain/select/src/lib/brn-select-list.ts)) nor helm's <div role="listbox"> ([hlm-select-content.ts](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/helm/select/src/lib/hlm-select-content.ts)) carries an id, so there is nothing for the attribute to point at even if a consumer wanted to add it by hand.

[...]

Issue #1753, opened 2026-09-18
select: BrnSelectTrigger host-binds aria-invalid from the resolved NgControl with no way to override it, unlike BrnAutocompleteInput


MPolokIQ:
### Please provide the environment you discovered this bug in.

```
@spartan-ng/brain 1.4.1 (also 1.3.1 and main@94587d1)
Angular 21
```

A wrapper component that owns its own error timing, i.e. shows an error only once the control is touched or the form was submitted, and binds the form control on its own host:

`<my-select formControlName="status" />   <!-- required -->`

1. Render the form. The control is required and therefore invalid, but untouched.
2. aria-invalid="true" is already on the trigger, and the wrapper cannot take it off.

### Which area/package is the issue in?

select

### Description

BrnSelectTrigger host-binds the raw validity of whatever control BrnFieldControl resolved:

`'[attr.aria-invalid]': '_invalid?.() ? "true" : null',`
([brn-select-trigger.ts#L17](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/brain/select/src/lib/brn-select-trigger.ts#L17), computed at [#L47](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/brain/select/src/lib/brn-select-trigger.ts#L47))

Two things make this hard to live with:

1. BrnFieldControl resolves the control from the node injector, this._injector.get(NgControl, null) ([brn-field-control.ts#L43-L49](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/brain/field/src/lib/brn-field-control.ts#L43-L49)), which walks up. A wrapper that binds formControlName on its own host therefore has that control adopted by brain, and brain publishes its raw validity on an element the wrapper does not own.
2. There is no reliable element to correct it on. hlm-select-trigger renders the <button> inside its own template and exposes only forceInvalid to the outside ([hlm-select-trigger.ts#L16-L24](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/helm/select/src/lib/hlm-select-trigger.ts#L16-L24)), so a helm consumer never touches that element at all. A consumer on brain does own the <button>, but then the directive's host binding and a template [attr.aria-invalid] write the same attribute, and which one lands is a question of binding order rather than of intent.
[...]

Issue #1751, opened 2026-09-18
select: PageUp/PageDown do nothing because the key manager is built without withPageUpDown()


MPolokIQ:
### Please provide the environment you discovered this bug in.

```
@spartan-ng/brain 1.4.1 (also 1.3.1 and main@94587d1)
@angular/cdk 21.2.14, Angular 21
```

A select with several dozen options, for instance a country list.

1. Open the listbox.
2. Press PageDown.

Expected (APG): the caret jumps ten options down. Actual: the caret does not move. The keydown falls through to the page, so the document behind the open panel scrolls instead.

### Which area/package is the issue in?

select

### Description

The key manager is configured without withPageUpDown(), in both variants:

```
this.keyManager
  .withVerticalOrientation()
  .withHomeAndEnd()
  .withTypeAhead()
  .withWrap()
  .skipPredicate((item) => item.disabled);
```

([brn-select.ts#L107-L112](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/brain/select/src/lib/brn-select.ts#L107-L112), [brn-select-multiple.ts#L107-L112](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/brain/select/src/lib/brn-select-multiple.ts#L107-L112))

In the CDK, both keys early-return unless the feature was switched on:

```
case PAGE_UP:
  if (this._pageUpAndDown.enabled && isModifierAllowed) { ... break; }
  else { return; }
```

Because the manager returns before reaching its trailing event.preventDefault(), the browser gets the key and scrolls the page behind the overlay.

APG, select-only combobox, listbox displayed:

`PageUp - Jumps visual focus up 10 options (or to first option). PageDown - Jumps visual focus down 10 options (or to last option).`

The CDK's default delta is exactly 10 (withPageUpDown(enabled = true, delta = 10)), so the fix is one line per variant and produces the documented behaviour with no further code:

```
this.keyManager
  .withVerticalOrientation()
  .withHomeAndEnd()
  .withPageUpDown()
  .withTypeAhead()
  ...
```

[...]

Issue #1747, opened 2026-09-18
select: selectActiveItem() commits disabled options, unlike the pointer path and unlike the key manager's own skipPredicate


MPolokIQ:
### Please provide the environment you discovered this bug in.

```
@spartan-ng/brain 1.4.1 (also 1.3.1 and main@94587d1)
Angular 21
```

```
<brn-select-multiple [(ngModel)]="assignees">
  <button brnSelectTrigger>...</button>
  <ng-template brnSelectContent>
    <div brnSelectList>
      <div brnSelectItem [value]="'anna'" disabled>Anna Weber (gesperrt)</div>
      <div brnSelectItem [value]="'ben'">Ben Krause</div>
    </div>
  </ng-template>
</brn-select-multiple>
```

The model starts as ['anna'], and that option is disabled: a locked assignment the user may see but not change.

1. Open the panel. The disabled row is highlighted, because the on-open effect points the key manager at the option matching the current value.
2. Press Enter.

Expected: nothing happens, the option is disabled. Actual: 'anna' is removed from the model. A value the user is not allowed to change is deleted from the keyboard.

### Which area/package is the issue in?

select

### Description

Brain is consistent everywhere except this one path:

- the pointer path guards it: BrnSelectItem.select() returns early when the item is disabled ([brn-select-item.ts#L72-L79](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/brain/select/src/lib/brn-select-item.ts#L72-L79));
- the key manager is told to skip them: .skipPredicate((item) => item.disabled) ([brn-select.ts#L112](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/brain/select/src/lib/brn-select.ts#L112), [brn-select-multiple.ts#L112](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/brain/select/src/lib/brn-select-multiple.ts#L112));
[...]

PR #1762, opened 2026-09-21
fix: a11y and focus handling for select, autocomplete, combobox

### LINKED ISSUE #1748: select(multiple): Tab toggles the highlighted option instead of leaving the control, and the panel stays open

```
@spartan-ng/brain 1.4.1 (also 1.3.1 and main@94587d1)
Angular 21
```

```
<brn-select-multiple [(ngModel)]="tags">
  <button brnSelectTrigger>...</button>
  <ng-template brnSelectContent>
    <div brnSelectList>
      <div brnSelectItem [value]="'a'">Alpha</div>
      <div brnSelectItem [value]="'b'">Beta</div>
    </div>
  </ng-template>
</brn-select-multiple>
```

1. Select Beta with the keyboard. The panel stays open, as intended for a multi-select.
2. Press Tab to move on to the next field.

Expected: focus moves to the next control, the panel closes, the selection is unchanged. Actual: Beta is de-selected, the panel stays open with nothing focused inside it, and focus has moved on.

select

### Description

BrnSelectTrigger.onKeyDown maps Enter and Tab onto selectActiveItem() while expanded ([brn-select-trigger.ts#L83-L91](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/brain/select/src/lib/brn-select-trigger.ts#L83-L91)).

For the single select this is correct and matches the APG:

`Tab - Sets the value to the content of the focused option in the listbox. Closes the listbox.`

BrnSelect.selectActiveItem() commits and then calls close() ([brn-select.ts#L159-L176](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/brain/select/src/lib/brn-select.ts#L159-L176)).

For the multiple select the same path means something else. BrnSelectMultiple.select() toggles membership and never closes ([brn-select-multiple.ts#L160-L171](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/brain/select/src/lib/brn-select-multiple.ts#L160-L171)), so Tab flips the highlighted option's selection state. Tab is not prevented, so focus leaves at the same moment and the panel is left open with nothing focused inside it.

The on-open effect makes this land on the worst possible option: it highlights the one matching the last value in the model ([#L126-L134](https://github.com/spartan-ng/spartan/blob/94587d1f4e08e584eccfcc2da34384e26c5d9028/libs/brain/select/src/lib/brn-select-multiple.ts#L126-L134)), so a user who picks a value and tabs on removes exactly the value they just picked.

The APG listbox pattern keeps Tab in the page tab sequence and gives it no selection meaning in a multi-select listbox.

Suggested fix: in BrnSelectMultiple, Tab should close the panel without touching the selection. Either gate the Tab branch in the trigger on single mode, or give the base a separate commitAndClose() that the multiple variant implements as plain close().

```true
No exception. Silent data change.
```

### Other information

The single-select behaviour is correct as it stands; only the multiple variant needs to change.

- [x] Yes
- [ ] No

marcjulian:
## PR Type

- [x] Bugfix

### Primitives

- [x] autocomplete
- [x] combobox
- [x] select

Closes #1754 
Closes #1746
Closes #1687
Closes #1747
Closes #1751
Closes #1753

- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
  - Added Page Up/Page Down navigation to autocomplete, combobox, and select menus.
  - Improved keyboard accessibility with active-item tracking and ARIA relationships.
  - Preserved trigger focus when clicking options.
- **Bug Fixes**
  - Tab now closes open menus without selecting highlighted options.
  - Disabled options can no longer be selected accidentally.
  - Improved focus movement and selection behavior across menu variants.
- **Breaking Changes**
  - Renamed the public `ariaInvalidOverride` input to `ariaInvalidInput`; the `aria-invalid` attribute remains supported.


libs/brain/select/src/lib/brn-select-trigger.ts
```diff
@@ -37,14 +41,28 @@ export class BrnSelectTrigger {
 	/** Whether to force the trigger into an invalid state. */
 	public readonly forceInvalid = input<boolean, BooleanInput>(false, { transform: booleanAttribute });
 
+	/** Manual override for aria-invalid. When not set, auto-detects from the parent autocomplete error state. */
+	public readonly ariaInvalidOverride = input<boolean | undefined, BooleanInput>(undefined, {
```
MerlinMoos:
```suggestion
	public readonly ariaInvalidInput = input<boolean | undefined, BooleanInput>(undefined, {
```

libs/helm/select/src/lib/hlm-select-trigger.ts
```diff
@@ -44,4 +45,10 @@ export class HlmSelectTrigger {
 
 	/** Whether to force the trigger into an invalid state. */
 	public readonly forceInvalid = input<boolean, BooleanInput>(false, { transform: booleanAttribute });
+
+	/** Manual override for aria-invalid. When not set, auto-detects from the parent autocomplete error state. */
+	public readonly ariaInvalidOverride = input<boolean | undefined, BooleanInput>(undefined, {
```
MerlinMoos:
```suggestion
	public readonly ariaInvalidInput = input<boolean | undefined, BooleanInput>(undefined, {
```

MerlinMoos (approved):

